PR #247, opened 2023-05-05
Added support for multi-epoch training

francescodeaglio:
Right now there is a static mapping 1 trigger 1 epoch and this might not be sufficient if we reset the model every trigger. A new pipeline parameter is added (epochs_per_trigger). If not present, the default behaviour is kept (1 epoch per trigger)

### DIFF
--- a/modyn/supervisor/internal/grpc_handler.py
+++ b/modyn/supervisor/internal/grpc_handler.py
@@ -59,6 +59,7 @@ class GRPCHandler:
         self.connected_to_selector = False
         self.progress_mgr = progress_mgr
         self.status_bar = status_bar
+        self.epochs_per_trigger = 1
 
         self.init_storage()
         self.init_selector()
@@ -294,6 +295,9 @@ class GRPCHandler:
         else:
             criterion_config = "{}"
 
+        if "epochs_per_trigger" in pipeline_config["training"]:
+            self.epochs_per_trigger = pipeline_config["training"]["epochs_per_trigger"]
+
         if "transformations" in pipeline_config["data"]:
             transform_list = pipeline_config["data"]["transformations"]
         else:
@@ -349,6 +353,7 @@ class GRPCHandler:
             label_transformer=PythonString(value=label_transformer),
             lr_scheduler=TrainerServerJsonString(value=json.dumps(lr_scheduler_configs)),
             grad_scaler_configuration=TrainerServerJsonString(value=json.dumps(grad_scaler_config)),
+            epochs_per_trigger=self.epochs_per_trigger,
         )
 
         response: StartTrainingResponse = self.trainer_server.start_training(req)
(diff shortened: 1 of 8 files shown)


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         status_query_queue: mp.Queue,
         status_response_queue: mp.Queue,
         logger: logging.Logger,
+        epochs_per_trigger: int,
```
MaxiBoether, line 40: this should go to the training_info

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -353,6 +356,7 @@ def train(
     exception_queue: mp.Queue,
     status_query_queue: mp.Queue,
     status_response_queue: mp.Queue,
+    epochs_per_trigger: int,
```
MaxiBoether, line 359: see other comment, this is part of the training info

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -243,93 +245,94 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
...
-            # we try to fetch an element within 10ms. If there is no
-            # element within that timeframe returned, we continue.
-            try:
-                req = self._status_query_queue.get(timeout=0.01)
-                if req == TrainerMessages.STATUS_QUERY_MESSAGE:
-                    self.send_status_to_server(batch_number)
-                elif req == TrainerMessages.MODEL_STATE_QUERY_MESSAGE:
-                    self.send_model_state_to_server()
+        for _ in range(self.epochs_per_trigger):
+            for batch_number, batch in enumerate(self._train_dataloader):
```
MaxiBoether, line 249: please add a test that this setting actually works. I don't think it would right now. the OnlineDataset has some internal state that always fetches the next training set, which means this would not always fetch the same data

MaxiBoether (changes requested):
here is a first review :) I think this is not complete yet - it needs tests which I think would fail right now since the onlinedataset is not written to revisit the sample training set multiple times

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -294,6 +295,9 @@ def start_training(
         else:
             criterion_config = "{}"
 
+        if "epochs_per_trigger" in pipeline_config["training"]:
+            self.epochs_per_trigger = pipeline_config["training"]["epochs_per_trigger"]
```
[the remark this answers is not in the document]
francescodeaglio: Makes sense, moved

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -411,10 +419,19 @@ def wait_for_training_completion(
                     ), f"Inconsistent server response:\n{res}"
 
                     new_samples = res.samples_seen - last_samples
+
                     if new_samples > 0:
                         sample_pbar.update(new_samples)
                         last_samples = res.samples_seen
 
+                    if res.samples_seen == total_samples * (current_epoch + 1):
```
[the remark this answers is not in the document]
francescodeaglio: I have tested it in 4 different trainings (even reaching 8 epochs) and it has never given any problems. I have now added logic to handle when answers are lost.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
francescodeaglio: I tried doing simple tests (adding a for loop to existing tests) and trying to run by analysing the chosen keys run by run and it seems to work. Any other clever tests I could add?

PR #248, opened 2023-05-05
Option to train just on the data of the last X triggers

francescodeaglio:
By experimenting on the yearbook dataset, we observed that during a distribution shift, recent points are preferred when downsampling techniques are used. With this PR we want to give the possibility to only take the latest triggers into account. 

For this first PR, this parameter is only implemented with NewDataStrategy but the goal is to have it for every strategy


modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -46,6 +46,20 @@ def __init__(
...
+            ), "Reset after trigger is equivalent to setting tail triggers to 0."
+            assert not (
+                not self.reset_after_trigger and self.tail_triggers == 0
+            ), "Reset after trigger is equivalent to setting tail triggers to 0."
+        else:
+            if self.reset_after_trigger:
+                self.tail_triggers = 0  # consider only the current trigger
+            else:
+                self.tail_triggers = -1  # consider every datapoint
+
```
MaxiBoether, line 62: Please add in all existing strategies checks that self.tail_triggers == -1 or not given at all, so we don't ignore it silently.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -46,6 +46,20 @@ def __init__(
...
         self.has_limit = self.training_set_size_limit > 0
         self.reset_after_trigger: bool = config["reset_after_trigger"]
+        if "tail_triggers" in config:
+            self.tail_triggers = config["tail_triggers"]
+            assert not (
+                self.reset_after_trigger and self.tail_triggers > 0
+            ), "Reset after trigger is equivalent to setting tail triggers to 0."
+            assert not (
+                not self.reset_after_trigger and self.tail_triggers == 0
+            ), "Reset after trigger is equivalent to setting tail triggers to 0."
```
MaxiBoether, line 56: Please use ValueErrors for invalid user input. assertions are for invalid internal state

modyn/selector/internal/selector_strategies/new_data_strategy.py
```diff
@@ -86,6 +88,16 @@ def _get_data_reset(self) -> Iterable[list[int]]:
 
             yield samples
 
+    def _get_data_tail(self) -> Iterable[list[int]]:
+        assert not self.reset_after_trigger and self.tail_triggers > 0
+
+        for samples in self._get_tail_triggers_data():
```
MaxiBoether, line 94: Can you please add tests for the NewDataStrategy testing for correct behavior :)? Thanks!

modyn/selector/internal/selector_strategies/new_data_strategy.py
```diff
@@ -131,7 +143,31 @@ def _get_current_trigger_data(self) -> Iterable[list[int]]:
...
+            list[int]: Keys of used samples
+        """
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            stmt = (
+                select(SelectorStateMetadata.sample_key)
+                # Enables batching of results in chunks. See https://docs.sqlalchemy.org/en/20/orm/queryguide/api.html#orm-queryguide-yield-per
+                .execution_options(yield_per=self._maximum_keys_in_memory)
+                .filter(
+                    SelectorStateMetadata.pipeline_id == self._pipeline_id,
+                    self._next_trigger_id - SelectorStateMetadata.seen_in_trigger_id <= self.tail_triggers,
```
MaxiBoether, line 170: Could you rephrase this as conditions on SelectorStateMetadata? Like SelectorStateMetadata. seen_in_trigger_id >= - self.tail_triggers - self._next_trigger_id or something that is maybe a bit more readable? It's a bit more easy to read to condition on SelectorStateMetadata I think.

modyn/selector/internal/selector_strategies/new_data_strategy.py
```diff
@@ -131,7 +143,31 @@ def _get_current_trigger_data(self) -> Iterable[list[int]]:
                 .execution_options(yield_per=self._maximum_keys_in_memory)
                 .filter(
                     SelectorStateMetadata.pipeline_id == self._pipeline_id,
-                    SelectorStateMetadata.seen_in_trigger_id == self._next_trigger_id,
+                    self._next_trigger_id == SelectorStateMetadata.seen_in_trigger_id,
```
MaxiBoether, line 146: Could you reverse the swap here?

MaxiBoether (changes requested):
Thanks for the PR! I have some suggestions, see below :) Thank you!!

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
francescodeaglio: I implemented it on all methods except freshness sampling, on which I put the check

modyn/selector/internal/selector_strategies/new_data_strategy.py
francescodeaglio: Sure!

MaxiBoether (approved):
I have one minor suggestion of not using -1 as a magic number - everything else looks good!

Issue #175, opened 2023-03-02
Accelerate online dataset prefetching if needed


fotstrt:
Currently, prefetching the next partition of keys and samples is done synchronously by each data worker. It is not clear yet if this will block/delay training or not. If it turns out to do so, we would probably need some kind of multithreading or `asyncio` support.

MaxiBoether:
This also affects the EvaluationDataset introdocued in #269

MaxiBoether:
Solved (for now) by #301.

PR #253, opened 2023-05-10
Asynchronous prefetching of data

MaxiBoether:
We add an additional thread to load data from the selector in order to hide the latency of data transfer from selector during training.

Solves #175.


fotstrt (approved):
Looks great!

francescodeaglio (approved):
Looks good to me!

MaxiBoether:
Note: This PR is on hold until we can measure the impact of this. We cannot be sure the current implementation here helps since we use multithreading. It _should_ help, since after  `yield` in the generator, we should be able to switch in the prefetch thread in the worker process, but who in the world understands the intricacies of Python multithreading

MaxiBoether:
Closing due to #301

PR #254, opened 2023-05-11
Seed Everything

francescodeaglio:
To fairly compare methods that involve random operations we need to support seeding. This is done by adding a parameter seed to the pipeline.


modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -391,3 +391,12 @@ def _persist_samples(self, keys: list[int], timestamps: list[int], labels: list[
 
         for proc in processes:
             proc.join()
+
+    def seed_selector(self, seed: float) -> bool:
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            if database.drivername == "postgresql":
+                assert -1 <= seed <= 1
+                database.session.execute(func.setseed(seed))
```
MaxiBoether, line 399: I see one potential problem here: You execute this function for this particular session that only exists in the seed_selector function. Is this setting _global_ for the entire database or local? Can you try whether creating a session, executing that function, creating a new session and getting the seed / at least some random number works correctly? I fear it might be per-session, in which case we need to move the seed to the modyn config and use it in the `MetadataDatabaseConnection` instead

MaxiBoether (changes requested):
Thanks! I have some comments regarding postgres seeding and running the seeding in the trainer server

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -391,3 +391,12 @@ def _persist_samples(self, keys: list[int], timestamps: list[int], labels: list[
 
         for proc in processes:
             proc.join()
+
+    def seed_selector(self, seed: float) -> bool:
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            if database.drivername == "postgresql":
+                assert -1 <= seed <= 1
```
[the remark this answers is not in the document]
francescodeaglio: The problem is that all libraries want an integer while postgres wants a float between [-1,+1]. Since having two different seeds doesn't seem necessary, only the _int_ seed is requested in the pipeline and the float is taken from there (rescaling the value between [-1,1]). Not the most elegant solution but it ensures reproducibility

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -224,6 +224,10 @@ properties:
         type: number
         description: |
           The modulus to use for the hash partitioning of the metadata.
+      seed:
+        type: number
+        description: |
+          The number used to seed the database. Must be in [-1,1].
```
MaxiBoether, line 230:
```suggestion
          If provided, this number is used to seed the database. Must be in [-1,1].
```

modyn/metadata_database/metadata_database_connection.py
```diff
@@ -34,6 +35,24 @@ def __init__(self, modyn_config: dict) -> None:
             if "hash_partition_modulus" in self.modyn_config["metadata_database"]
             else 16
         )
+        self.seed: int = (
+            self.modyn_config["metadata_database"]["seed"] if "seed" in self.modyn_config["metadata_database"] else None
+        )
+        if self.seed is not None:
+            assert -1 <= self.seed <= 1
```
MaxiBoether, line 42: use ValueError instead of assert (user input)

modyn/utils/utils.py
```diff
@@ -123,3 +127,13 @@ def package_available_and_can_be_imported(package: str) -> bool:
...
+
+
+def seed_everything(seed: int) -> None:
+    random.seed(seed)
+    os.environ["PYTHONHASHSEED"] = str(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    torch.backends.cudnn.deterministic = True
+    torch.backends.cudnn.benchmark = False
```
MaxiBoether, line 139: I think we might also want to call this function on the selector when we instantiate the AbstractSelectionStrategy (in __init__) since some strategies use `random`

MaxiBoether (changes requested):
thanks again - let me know regarding gRPC optional, other than that I had some minor comments

modyn/utils/utils.py
francescodeaglio: Yep, there's a random shuffling of samples. I added a GRPC call from the supervisor to the selector just after the connection.

modyn/protos/trainer_server.proto
```diff
@@ -53,6 +53,8 @@ message StartTrainingRequest {
   PythonString label_transformer = 19;
   JsonString grad_scaler_configuration = 20;
   int32 epochs_per_trigger = 21;
+  bool seed_available= 22;
+  optional int32 seed = 23;
```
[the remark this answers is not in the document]
MaxiBoether: based on your investigations, the `optional` keyword is basically useless, isn't it? If so, I would be happy if you could remove it. Thx!

MaxiBoether (approved):
almost LGTM - there still is a leftover (?) optional. Please remove that, update the gRPC and then feel free to merge

MaxiBoether:
let's get this merged soon if you have 5 minutes :)

MaxiBoether:
feel free to merge @francescodeaglio

francescodeaglio:
I found a nice way to handle optionals in Foteini's code. I changed the structure accordingly, removing seed_available

francescodeaglio:
Possible problem: we use the same seed for each connection to the database so this might interfere with multi epoch training (since each epoch would see the same data). It's not a problem across triggers since the query is different and thus the result changes.

My suggestion is to increment each time the seed, so it's still deterministic and reproducible but each time we get different data.

For example, we can use a seed with up to 3 decimal digits and then every time we call the set_seed function we can do something like
```
...seed postgres...
self.seed += 0.001
if self.seed > 1:
  self.seed = 0.000
```

MaxiBoether:
As discussed today, the database is not used when querying the finalized trigger training set. Hence, I think this should be safe for now.

PR #255, opened 2023-05-11
Random Downsampling

francescodeaglio:
Randomly sampling datapoints is frequently used as a baseline for downsampling methods. This PR implements it as an instance of AbstractDownsampler without the remote part (since we just sample from the database)


modyn/selector/internal/selector_strategies/random_downsampling_strategy.py
```diff
@@ -0,0 +1,7 @@
+from modyn.selector.internal.selector_strategies.abstract_downsample_strategy import AbstractDownsampleStrategy
+
+
+class RandomDownsamplingStrategy(AbstractDownsampleStrategy):
```
MaxiBoether, line 4: Can you please add a few tests that this class has its expected behavior (probably just copy/paste some tests from AbstractDownsampleStrategy)?

modyn/tests/selector/internal/selector_strategies/test_actual_samplers.py
```diff
@@ -5,6 +5,7 @@
```
MaxiBoether, line 5: One small thing: Could you update the file structure in the tests such that we have one test_blabla.py per blabla.py in the main source? So instead of test_actual_samplers, we have the tests in the respective test_loss_downsampling_strategy.py etc. I know the tests are short but otherwise it is hard to navigate what is tested where

MaxiBoether (approved):
One comment about file structure, afterwards feel free to resolve + merge

Issue #167, opened 2023-02-27
Implement Model Storage Component


MaxiBoether:
Right now, we download a trained model to the Supervisor and then re-upload it to the trainer server for training.

We want to switch to a disaggregated model storage. This model storage shall behave similar to our storage component, i.e., it should support different underlying file systems (local/network file systems, S3 buckets, ...). 

After finishing a training, the trainer server should be able to "push" a model to the model storage. For this, we need to offer a class that wraps knowledge how to push data to storage. Basically, I want to avoid that the trainer server still runs an FTP server. Instead, the trainer server uses something like a `ModelStorageHandler` (find better name without handler). This handler first sends a gRPC request to the model storage component to find out the type of storage being used and where to persist it with what identifier. Then, using that information, the handler knows how to send data either to local filesystem (e.g., via FTP/mounting a NFS/...) or to S3/GCloud/... and pushes the model there. So we want to avoid the intermediate step of somehow pushing the model to the central model storage component via gRPC (which won't work anyway) and instead just querying where to push the model. We then have hold of a unique identifier we can return to the supervisor and the supervisor can then in the next step say "please continue training on model X" to the trainer server or "please evaluate model X" to the evaluator.

PR #257, opened 2023-05-15
Connect Model Storage

robin-oester:
Connect model storage to the other components (namely supervisor and trainer server) according to https://github.com/eth-easl/modyn/issues/167. The models should be stored in the model storage instead of downloading to the supervisor. The supervisor takes care of the previous model id in order to continue training on the stored model.


modyn/protos/trainer_server.proto
```diff
@@ -40,7 +40,7 @@ message StartTrainingRequest {
   JsonString model_configuration = 6;
   bool use_pretrained_model = 7;
   bool load_optimizer_state = 8;
-  string pretrained_model_path = 9;
+  optional int32 pretrained_model_id = 9;
```
MaxiBoether, line 43: This is actually something that we also discuss with @francescodeaglio in the seed everything PR:

Apparently, at least for non-optional fields, if they are None, they default to 0. This is why we have the use_pretrained_model boolean.

However, I didn't know about optional fields. The question for me is what the `optional` keyword changes. If the optional keyword prevents fields from defaulting to 0 such that they are None instead of 0 on the receiving end, we can actually remove the `use_pretrained_model` flag. However, if they still default to 0, we still require the bool, but then can remove `optional`. So either bool+no optional or optional + no bool, but not optional + bool. If you (or @francescodeaglio, whoever gets to it first :D) could figure out what `optional` does that'd be great.

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
```diff
@@ -26,15 +34,19 @@
 from modyn.trainer_server.internal.utils.trainer_messages import TrainerMessages
 from modyn.trainer_server.internal.utils.training_info import TrainingInfo
 from modyn.trainer_server.internal.utils.training_process_info import TrainingProcessInfo
-from modyn.utils import current_time_millis, dynamic_module_import
+from modyn.utils import current_time_millis, dynamic_module_import, grpc_connection_established
 
 logger = logging.getLogger(__name__)
 
+EMIT_MESSAGE_PERCENTAGES = [0.25, 0.5, 0.75]
```
MaxiBoether, line 41: Let's keep this in modyn.utils and export it there. I think we can have this global across the package

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
```diff
@@ -75,12 +100,39 @@ def start_training(
...
+                    f"Training cannot be started."
+                )
                 return StartTrainingResponse(training_started=False)
 
+            last_progress = 0.0
+
+            def download_callback(current_progress: float) -> None:
+                nonlocal last_progress
+                for emit_perc in EMIT_MESSAGE_PERCENTAGES:
+                    if last_progress <= emit_perc < current_progress:
```
MaxiBoether, line 119: sorry, but I actually don't know how to read this comparison statement haha. Can you add brackets to make this a bit more readable? That would be great :D I don't know in which order this is evaluated

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
```diff
@@ -75,12 +100,39 @@ def start_training(
...
+                hostname=self._config["model_storage"]["hostname"],
+                port=int(self._config["model_storage"]["ftp_port"]),
+                user="modyn",
+                password="modyn",
+                remote_file_path=pathlib.Path(fetch_resp.model_path),
+                local_file_path=pretrained_model_path,
+                callback=download_callback,
+            )
+
+            logger.info("Completed pretrained model download.")
```
MaxiBoether, line 134: let's put this in a separate function, like _download_pretrained_model

MaxiBoether (changes requested):
Looks really solid. I did't have many comments. My only question would be whether you have also tested running a pipeline that continuous training, like have you actually run the end2end workflow? That would be good I think, because we have no training e2e test. Otherwise, again, really solid code!

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
robin-oester: This feature of python is called comparison operator chaining and syntactic sugar for (in this case):
`last_progress <= emit_perc AND emit_perc < current_progress`
I agree, not the most readable solution if one is not used to it.
Should I still change this?

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
robin-oester: Done

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
robin-oester: I agree, done in latest commit.

modyn/protos/trainer_server.proto
robin-oester: Okay, yes I agree with this. Removed the optional in latest commit.

robin-oester:
Thank you Maxi for the feedback! I run a pipeline with continuous training on my local machine. Seems to work, I got no errors and the models were stored in the model storage as expected. The trainer also downloaded and resumed training on the stored models.

Something I thought about: The models are always downloaded and stored after training on the trainer server. A (little) enhancement would be to delete them after finishing training. What do you think?

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
MaxiBoether: ahh it's actually read like the mathematical statement. Thank you - I actually did not know this! No we can keep it that way.

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
```diff
@@ -160,29 +218,47 @@ def get_training_status(
...
+                port=int(self._config["trainer_server"]["ftp_port"]),
+                model_path=prefix_path,
+            )
+
+            register_response: RegisterModelResponse = self.model_storage_stub.RegisterModel(register_request)
+
+            if not register_response.success:
+                logger.error(f"Could not store final model from training id {training_id}.")
+                return StoreFinalModelResponse(valid_state=False)
+
```
MaxiBoether, line 257: As suggested by you, I would delete the snapshop from local disk here. This is a good idea. You are right that we clean up the temp directory anyways at the end, but if the trainer server runs for a long time, we might overflow the disk. Hence, we should clean up here

MaxiBoether (approved):
Thank you so much! My last comment is just that your suggestion of cleaning up after registering the model makes a

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
```diff
@@ -255,11 +257,22 @@ def store_final_model(
...
 
+    def _cleanup_stored_models(self, training_id: int, final_checkpoint_path: pathlib.Path) -> None:
+        pretrained_model_path = self._training_dict[training_id].pretrained_model_path
+        if pretrained_model_path and pretrained_model_path.is_file():
+            os.remove(pretrained_model_path)
+            logger.info(f"Successfully deleted pretrained model at {pretrained_model_path}")
+        if final_checkpoint_path.is_file():
+            os.remove(final_checkpoint_path)
+            logger.info(f"Successfully deleted final model at {final_checkpoint_path}")
+
```
robin-oester, line 275: I deleted the pretrained model as well as the final model (if they exist) as they are both persistent on model storage at this point. The checkpoint files are not touched.
Can we go with this?

MaxiBoether: +1

modyn/trainer_server/internal/grpc/trainer_server_grpc_servicer.py
robin-oester: Done in next commit.

Issue #104, opened 2023-01-20
Seperate component-specific dependencies


fotstrt:
Currently, all dependencies for the various components are included in the `environment.yml` file. However, the different modules have different requirements, e.g. the Trainer depends on PyTorch and Torchvision, which are not needed by the Supervisor.

We should be able to clearly separate component-specific dependencies, from dependencies required for the whole project.

Furthermore, since training will most commonly run on GPUs, we need to specify the [pytorch-cuda version](https://pytorch.org/get-started/locally/) we are using, and install cudatoolkit. Furthermore, on macOS, the pytorch channel PyTorch version breaks, so we somehow want conditional dependencies.

Last, we might want conditional dependencies in a sense that we specify "use_apex" somewhere and then install apex if that is set on the trainer server.

MaxiBoether:
PR #258 addresses the conditional dependencies for mac/cuda.

However, we still have to think about whether we want to have component-specific dependencies. Maybe it also does not make sense because I see many Docker issues coming up, need to think about this.

PR #258, opened 2023-05-17
Multi-stage build to cache Apex, conditional dependency management for ARM-systems, ask for CUDA/Apex

MaxiBoether:
- Prevents Apex from being rebuilt everytime the Modyn sourcecode changes
- Finally at least a partial solution for adjusting the environment.yml on mac
- Helps in installing Apex, CUDA, etc. by providing an interactive initial setup script

Partially solves #104.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -130,7 +130,7 @@ jobs:
       uses: ./.github/actions/base
 
     - name: Setup dev-requirements and run pytest within container
-      run: docker run modyn conda run -n modyn bash -c "pip install -r dev-requirements.txt && echo Running pytest && pytest"
+      run: docker run modynbase conda run -n modyn bash -c "pip install -r dev-requirements.txt && echo Running pytest && pytest"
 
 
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -80,6 +80,7 @@ services:
 #    volumes:
 #      - .:/modyn_host
 #    shm_size: 4gb
+# CUDASTART (ignore/delete if enabling cuda manually)
```
vGsteiger, line 83: Just from an understanding point, this signs for cuda to include the following code?

docker/Dependencies/Dockerfile
vGsteiger, line 1: Much cleaner!

vGsteiger (approved):
Much cleaner, thanks!

docker-compose.yml
MaxiBoether: This is for the `initial_setup.sh` script :D To find the lines we are searching for reliably.

docker/Dependencies/Dockerfile
MaxiBoether: :)

PR #264, opened 2023-05-25
Check whether trigger sample directory exists and is writable on Selector startup

MaxiBoether:
The selector validates whether the directory exists and is writable on startup. This prevents wasting time waiting for experiments that fail as soon as we store a trigger training set to disk.


vGsteiger (approved):
Lgtm thanks!

PR #267, opened 2023-05-26
Add `Stopwatch` for performance benchmarks

### LINKED ISSUE #239: Benchmarking Stopwatch
As we introduce some benchmarking code, we should probably not do this:

```
            init_start = current_time_millis()

            file_wrapper = get_file_wrapper(
                file_wrapper_type, file_path, dataset.file_wrapper_config, filesystem_wrapper
            )
            number_of_samples = file_wrapper.get_number_of_samples()
            logger.debug(
                f"[Process {process_id}] Found new, unknown file: {file_path} with {number_of_samples} samples."
            )

            time_spent["init"] += current_time_millis() - init_start
```

but have something better in the long term.

For example, we could write a BenchmarkStopwatch class. This could do .start_measurement("name") and .stop_measurement("name") and create a dict with measurements at the end. I will address this soon.

MaxiBoether:
This adds a stopwatch for performance benchmarks. Only to be used for Modyn development (internal tooling).

Next step would be actually creating a big json containing timing for all steps in the pipeline and output this measurement json (if requested from the user) for a pipeline.

Closes #239.

### DIFF
--- a/modyn/common/benchmark/__init__.py
+++ b/modyn/common/benchmark/__init__.py
@@ -1,0 +1,7 @@
+import os
+
+from .stopwatch import Stopwatch  # noqa: F401
+
+files = os.listdir(os.path.dirname(__file__))
+files.remove("__init__.py")
+__all__ = [f[:-3] for f in files if f.endswith(".py")]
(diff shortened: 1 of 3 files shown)


vGsteiger (approved):
Lgtm

Issue #26, opened 2022-12-10
Evaluator documentation and testing


vGsteiger:
Document the Evaluator interface. Additionally, implement unit tests and integration tests on the Evaluator node level.

MaxiBoether:
@robin-oester is working on this in context of his bachelor thesis :)

MaxiBoether:
Closed by #269

Issue #15, opened 2022-12-10
Implement Evaluator


vGsteiger:
The following tasks need to be done:

- [ ] Define the interface to the trainer
- [ ] Define the interface to the Training Supervisor
- [ ] Implement a new class and put it on the backend node

MaxiBoether:
@robin-oester is working on this in context of his bachelor thesis :)

MaxiBoether:
Closed by #269

PR #256, opened 2023-05-15
Introduce Sample then Batch

francescodeaglio:
In this PR, the second downsampling mode, sample-then-batch, is introduced.
In this mode, first the dataset is downsampled and then the batches are created.

PR #269, opened 2023-05-26
Implement Evaluator

robin-oester:
Implement evaluator component in order to evaluate our trained models. First, define GRPC interface to supervisor. Then, implement adjustments to pipeline schema in order to make the evaluation process configurable. The models should be downloaded from model storage and it should be possible to evaluate them on several datasets with different metrics. Implement simple metrics like accuracy or AUC-ROC to test. Corresponding issues: #26, #15.


modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+            hostname=self._config["model_storage"]["hostname"],
+            port=int(self._config["model_storage"]["ftp_port"]),
+            user="modyn",
+            password="modyn",
+            remote_file_path=pathlib.Path(fetch_resp.model_path),
+            local_file_path=local_model_path,
+            callback=get_pretrained_model_callback(logger),
+        )
+
+        logger.info(f"Successfully downloaded trained model {request.trained_model_id} to {local_model_path}.")
```
MaxiBoether, line 110: maybe we can move lines 84-110 in another function to make this one shorter?

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+        process.start()
+        self._evaluation_process_dict[evaluation_id] = EvaluationProcessInfo(
+            process, exception_queue, status_query_queue, status_response_queue
+        )
+
+        logger.info(f"Started evaluation {evaluation_id}.")
+        return EvaluateModelResponse(evaluation_started=True, evaluation_id=evaluation_id)
+
+    def _setup_metrics(self, metrics: list[str]) -> list[AbstractEvaluationMetric]:
+        return list(map(self._metric_manager.get_evaluation_metric, set(metrics)))
```
MaxiBoether, line 139: could you write a comment what this is doing? sorry, just from looking at it it's not clear to me why we need it

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+        # As qsize() is unreliable and not implemented on macOS,
+        # we try to fetch an element within 100ms. If there is no
+        # element within that timeframe returned, we return None.
+        try:
+            exception = exception_queue.get(timeout=0.1)
+            return exception
+        except queue.Empty:
+            return None
+
+    def get_final_evaluation(
```
MaxiBoether, line 203: maybe rename this to get_evaluation_results?

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+        self, request: FinalEvaluationRequest, context: grpc.ServicerContext
+    ) -> FinalEvaluationResponse:
+        evaluation_id = request.evaluation_id
+        logger.info(f"Received get final evaluation request for evaluation {evaluation_id}.")
+
+        if evaluation_id not in self._evaluation_dict:
+            logger.error(f"Evaluation with id {evaluation_id} has not been registered.")
+            return FinalEvaluationResponse(valid=False)
```
MaxiBoether, line 211: similarily, we could name this EvaluationResultReponse etc.

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+        if self._evaluation_process_dict[evaluation_id].process_handler.is_alive():
+            logger.error(f"Evaluation with id {evaluation_id} is still running.")
+            return FinalEvaluationResponse(valid=False)
+
+        logger.error("Returning results of all metrics.")
+        evaluation_data: list[EvaluationData] = []
+        for metric in self._evaluation_dict[evaluation_id].metrics:
+            evaluation_data.append(
+                EvaluationData(metric=metric.__class__.__name__, result=metric.get_evaluation_result())
+            )
```
MaxiBoether, line 222: do we want to clean up (delete the downloaded model) here, after the results have been fetched? i think that is safe to do

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
MaxiBoether: (and then the next lines could be sth like run evaluation etc, so we factor this method here in short method calls)

modyn/evaluator/internal/metric_manager.py
```diff
@@ -0,0 +1,13 @@
+from modyn.evaluator.internal.metrics import AbstractEvaluationMetric
+from modyn.utils import dynamic_module_import
+
+
+class MetricManager:
```
MaxiBoether, line 5: sorry, the purpose of this class not yet clear to me by its name. So does this allow us to instantiate a metric by its name? Maybe it is a `MetricFactory` then with a function `from_name` that returns an AbstractEvaluationMetric?

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC, abstractmethod
+
+import torch
+
+
+class AbstractEvaluationMetric(ABC):
```
MaxiBoether, line 6: just a note: here we can keep the Metric in the name. Accuracy then is an instance of a AbstractEvaluationMetric

modyn/evaluator/internal/metrics/accuracy_metric.py
```diff
@@ -0,0 +1,34 @@
...
+            y_pred = y_pred.unsqueeze(1)
+
+        correct_labeled = torch.sum(torch.all(torch.eq(y_pred, y_true), dim=-1)).item()
+        assert correct_labeled <= batch_size, "Total correct amount is greater than batch size."
+
+        self.total_correct += correct_labeled
+        self.seen_samples += batch_size
+
+    def get_evaluation_result(self) -> float:
+        if self.seen_samples == 0:
```
MaxiBoether, line 31: maybe warn in this case

modyn/evaluator/internal/metrics/accuracy_metric.py
```diff
@@ -0,0 +1,34 @@
...
+        self.total_correct = 0
+
+    def evaluate_batch(self, y_true: torch.tensor, y_pred: torch.tensor, batch_size: int) -> None:
+        if y_true.shape != y_pred.shape:
+            raise TypeError(f"Shape of y_true and y_pred must match. Got {y_true.shape} and {y_pred.shape}.")
+        assert y_true.shape[0] == batch_size, "Batch size and target label amount is not equal."
+
+        if y_true.dim() < 2:
+            y_true = y_true.unsqueeze(1)
+            y_pred = y_pred.unsqueeze(1)
```
MaxiBoether, line 22: could you maybe add a comment what is happening here :)?

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, progressive: bool):
+        """
+        Abstract class for all evaluation metrics.
+        Currently we only allow progressive metrics i.e metrics where the final evaluation value can be calculated
+        on-the-fly without having the model output from all batches.
+
+        Args:
+            progressive: whether the metrics is progressive.
+        """
+        self.progressive = progressive
```
MaxiBoether, line 20: usually, you say a metric is _decomposable_ for what you call progressive. non-decomposable metrics (like median) are also called holistic. we could either flip the meaning and call this holistic or rename progressive to decomposable (I'd probably do that)

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if not self.progressive:
+            raise NotImplementedError("Only progressive metrics are allowed.")
+
+    @abstractmethod
+    def evaluate_batch(self, y_true: torch.tensor, y_pred: torch.tensor, batch_size: int) -> None:
```
MaxiBoether, line 26: maybe rename batch_evaluated_callback? not sure WDYT?

modyn/evaluator/internal/metric_manager.py
MaxiBoether: (in that case I think from_name can also be static)

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -42,7 +42,6 @@ def __init__(
         self._trigger_id = trigger_id
         self._training_id = training_id
         self._dataset_id = dataset_id
-        self._dataset_len = 0
```
MaxiBoether, line 45: i assume this was just unused?

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -673,6 +669,7 @@ def test_train(
         if elapsed >= timeout:
             raise AssertionError("Did not reach desired queue state after 5 seconds.")
 
+        # TODO(@robin-oester & @MaxiBoether): this does not seem to be right...
```
MaxiBoether, line 672: Could you elaborate here? What is the problem?

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -654,10 +654,6 @@ def test_train(
     assert test_send_metadata.call_count == len(trainer._callbacks)
     test_cleanup.assert_called_once()
 
-    if not platform.system() == "Darwin":
-        assert status_queue.qsize() == 1
-    else:
-        assert not status_queue.empty()
```
MaxiBoether, line 660: why did you remove this?

modyn/tests/trainer_server/internal/grpc/test_trainer_server_grpc_servicer.py
```diff
@@ -166,6 +166,12 @@ def test_start_training_invalid(test_hasattr, test_connect_to_model_storage):
         assert not trainer_server._training_dict
         assert trainer_server._next_training_id == 0
 
+
+@patch.object(TrainerServerGRPCServicer, "connect_to_model_storage", return_value=DummyModelStorageStub())
+@patch("modyn.trainer_server.internal.grpc.trainer_server_grpc_servicer.hasattr", return_value=True)
+def test_start_training_invalid_id(test_hasattr, test_connect_to_model_storage):
+    with tempfile.TemporaryDirectory() as modyn_temp:
+        trainer_server = TrainerServerGRPCServicer(modyn_config, modyn_temp)
```
MaxiBoether, line 174: is that just me or does the diff look weird? it looks like you added a function header here, but where is the actual test coming from?

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,177 @@
...
+import traceback
+from inspect import isfunction
+from typing import Any, Callable, Optional, Union
+
+import torch
+from modyn.evaluator.internal.dataset.evaluation_dataset import EvaluationDataset
+from modyn.evaluator.internal.utils import EvaluationInfo, EvaluatorMessages
+
+EVALUATION_TRANSFORMER_FUNC_NAME = "evaluation_transformer"
+LABEL_TRANSFORMER_FUNC_NAME = "label_transformer_function"
```
MaxiBoether, line 16: this is the same for the pytorch trainer, right? let's move these constants to common

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,177 @@
...
+        self._debug("Creating DataLoader.")
+        dataloader = torch.utils.data.DataLoader(
+            dataset, batch_size=evaluation_info.batch_size, num_workers=evaluation_info.num_dataloaders
+        )
+
+        return dataloader
+
+    def _setup_transformer_function(self, serialized_code: str, func_name: str) -> Optional[Callable]:
+        if serialized_code == "":
+            return None
```
MaxiBoether, line 79: why is that required? we did not have that in the trainer server, right?

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,177 @@
...
+        self._mod_dict: dict[str, Any] = {}
+        self._evaluation_transformer = self._setup_transformer_function(
+            evaluation_info.evaluation_transformer, EVALUATION_TRANSFORMER_FUNC_NAME
+        )
+
+        self._metrics = evaluation_info.metrics
+
+        self._label_tranformer_function = self._setup_transformer_function(
+            evaluation_info.label_transformer, LABEL_TRANSFORMER_FUNC_NAME
+        )
```
MaxiBoether, line 48: in general I really like this approach, but as said in another comment, if we introduce it, let's unify it in all places where we expect code in the pipeline dsecription

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,177 @@
...
+                    output = self._model.model(data)
+
+                    if self._evaluation_transformer:
+                        output = self._evaluation_transformer(output)
+
+                    for metric in self._metrics:
+                        metric.evaluate_batch(target, output, batch_size)
+
+                self._num_samples += batch_size
+
```
MaxiBoether, line 151: can you add a todo here (and create an issue) to support holistic metrics? I think it should be the goal of your thesis that we support at least F1, Accuracy, and AUC :)

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,177 @@
...
+    def send_status_to_server(self, batch_number: int) -> None:
+        self._status_response_queue.put({"num_batches": batch_number, "num_samples": self._num_samples})
+
+    def evaluate(self) -> None:
+        self._info(f"Process {os.getpid()} starts evaluation.")
+
+        self._model.model.eval()
+        with torch.no_grad():
+            batch_number = -1
```
MaxiBoether, line 112: why is this required?

modyn/config/examples/example-pipeline.yaml
```diff
@@ -57,8 +57,19 @@ data:
...
+    - dataset_id: mnist
+      batch_size: 64
+      dataloader_workers: 2
+      metrics: ["AccuracyMetric"]
+  device: "cpu"
+  amp: False
+  evaluation_transformer: |
+    import torch
+    def evaluation_transformer(model_output: torch.Tensor) -> torch.Tensor:
+      return torch.argmax(model_output, dim=-1)
```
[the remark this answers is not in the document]
MaxiBoether: that should also be documented in the AbstractEvaluationMetric and the pipeline schema I think (so it should be documented what the expected input is)

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+        self._evaluation_dict[evaluation_id] = evaluation_info
+
+        exception_queue: mp.Queue[str] = mp.Queue()  # pylint: disable=unsubscriptable-object
+        status_query_queue: mp.Queue[str] = mp.Queue()  # pylint: disable=unsubscriptable-object
+        status_response_queue: mp.Queue[dict[str, Any]] = mp.Queue()  # pylint: disable=unsubscriptable-object
+
+        process = mp.Process(
+            target=evaluate,
+            args=(
+                self._evaluation_dict[evaluation_id],
```
MaxiBoether, line 123: I am not sure this approach works. it might work on linux devices with fork but e.g. on mac where we use spawn this is a problem. this dict is not shared between the processes. So i think we need to use a queue into which the evaluator pushes the results upon termination, and then pop from that queue when obtaining the final evaluation results

MaxiBoether (changes requested):
Thank you! This looks really solid already.

My main two points are the input to the metrics (one hot vs softmax) and the evaluation dict in the multiprocessing process where I think we need to use a queue instead to obtain the results.

Let's talk more tomorrow (today for you :D)!

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
MaxiBoether: note: there should not be identation after the waiting

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -0,0 +1,223 @@
...
+            logger.error(f"Evaluation with id {evaluation_id} has not been registered.")
+            return FinalEvaluationResponse(valid=False)
+
+        if self._evaluation_process_dict[evaluation_id].process_handler.is_alive():
+            logger.error(f"Evaluation with id {evaluation_id} is still running.")
+            return FinalEvaluationResponse(valid=False)
+
+        logger.error("Returning results of all metrics.")
+        evaluation_data: list[EvaluationData] = []
+        for metric in self._evaluation_dict[evaluation_id].metrics:
```
MaxiBoether, line 219: as discussed during our call, the dict is not shared, so we cannot access the results like that

modyn/config/examples/example-pipeline.yaml
```diff
@@ -57,8 +57,19 @@ data:
...
 trigger:
   id: DataAmountTrigger
   trigger_config:
     data_points_for_trigger: 100
+evaluation:
+  datasets:
+    - dataset_id: mnist
+      batch_size: 64
+      dataloader_workers: 2
+      metrics: ["AccuracyMetric"]
```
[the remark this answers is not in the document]
robin-oester: I agree, fixed.

modyn/config/examples/example-pipeline.yaml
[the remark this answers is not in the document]
robin-oester: As we have discussed, I changed the name of this function to evaluation_transform_function. It is a per metric function that is applied before evaluation in order to transform the labels and model outputs. I documented it in the pipeline schema.

modyn/evaluator/evaluator.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Evaluator:
+    def __init__(self, config: dict) -> None:
+        self.config = config
+
+        valid, errors = self._validate_config()
+        if not valid:
+            raise ValueError(f"Invalid configuration: {errors}")
+
+        self.working_directory = pathlib.Path(tempfile.gettempdir()) / "modyn"
```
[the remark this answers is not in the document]
robin-oester: Done

modyn/storage/internal/grpc/storage_grpc_servicer.py
```diff
@@ -333,3 +374,36 @@ def remove_empty_files(self, session: Session, dataset: Dataset) -> None:
...
+
+        threshold = total_num_keys % total_workers
+        if threshold > 0:
+            if worker_id < threshold:
+                worker_subset_size += 1
+                start_index = worker_id * (subset_size + 1)
+            else:
+                start_index = threshold * (subset_size + 1) + (worker_id - threshold) * subset_size
+        else:
+            start_index = worker_id * subset_size
```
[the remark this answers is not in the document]
robin-oester: Yes, I saw this function as well. They are not identical however they both try to achieve the same. Therefore, I kept my version and moved it to `modyn.utils`.

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robin-oester: Done

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robin-oester: Done

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robin-oester: This is done in `pytorch_evaluator._load_state`. The downloaded model is removed after the state is loaded.

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
robin-oester: Okay, I agree.

modyn/evaluator/internal/metrics/accuracy_metric.py
robin-oester: Done

modyn/evaluator/internal/metrics/accuracy_metric.py
robin-oester: Sure :)

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
robin-oester: I see, thank you for clarification! In the next commit I did some changes to reflect this.

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
robin-oester: Naming is hard^^
Yes, I think this makes sense. It's a function that is called whenever we can make progress in our metric calculation (meaning that model outputs and labels are ready to be processed for one batch).

modyn/trainer_server/internal/dataset/online_dataset.py
robin-oester: Yes, unused attribute.

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
robin-oester: Fixed in next commit. The indentation is removed and some minor edits were done.

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
robin-oester: The status_queue might not contain the result yet. We have to wait (which is done in lines 662-674).

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Yes, exactly. I moved them to `modyn.utils`.

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: It is required as we want to differentiate between empty string = no function and a malformed serialized function where we throw an error. This was also done similarly in the trainer server. But maybe I'm missing something here.

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: I agree, done.

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Yes, sure :)
I didn't add a TODO here as I want to implement the functionality in this PR.

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Are you referring to `batch_number = -1`? I used it to log the total number of evaluated batches.

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robin-oester: You are totally correct! As discussed, I changed this in the next commit to a queue. However, I do not know if other solutions might be more suitable (like shared memory?).

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robin-oester: Agree, see comment above :)

modyn/evaluator/internal/pytorch_evaluator.py
MaxiBoether: Ah, ok, I'm not aware that this was a problem in the trainer server, but ok makes sense

modyn/config/schema/pipeline-schema.yaml
```diff
@@ -309,6 +309,76 @@ properties:
...
+                    description: |
+                      The name of the evaluation metric. Currently we support F1-score, Accuracy and ROC-AUC.
+                  config:
+                    type: object
+                    description: |
+                      Optional configuration for the evaluation metric.
+                  evaluation_transformer_function:
+                    type: string
+                    description: |
+                      The (optional) function used to transform the labels and the model output before evaluation.
```
MaxiBoether, line 360: Was there a reason we pass the label through this? Or should we just always pass the label hard coded to the metric? I remember that we need to pass the model output through this, but do not remember the reason for the label

modyn/evaluator/internal/dataset/evaluation_dataset.py
```diff
@@ -0,0 +1,128 @@
...
+    deserialize_function,
+    grpc_connection_established,
+)
+from torch.utils.data import IterableDataset, get_worker_info
+from torchvision import transforms
+
+logger = logging.getLogger(__name__)
+
+
+class EvaluationDataset(IterableDataset):
```
MaxiBoether, line 24: As discussed, we should open an issue + add a TODO to unify this with the OnlineDataset (or build common abstractions) after #256 is merged (since that PR introduces many structural changes to the OD)

modyn/evaluator/internal/metrics/accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __init__(self, evaluation_transform_func: str, config: dict[str, Any]) -> None:
+        super().__init__(evaluation_transform_func, config)
+        self.seen_samples = 0
+        self.total_correct = 0
+
+    def batch_evaluated_callback(self, y_true: torch.tensor, y_pred: torch.tensor, batch_size: int) -> None:
+        if self.evaluation_transformer_function:
+            y_true, y_pred = self.evaluation_transformer_function(y_true, y_pred)
```
MaxiBoether, line 19: If this is something all metrics do, maybe we should have a function in the abstract class that does this, and then calls the AbstractImpl? So evaluator calls function XYZ on abstractclass, that does this, and then calls the specific impl

modyn/evaluator/internal/metrics/accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Accuracy(AbstractDecomposableMetric):
+    """
+    Accuracy metric implementation.
+    """
+
+    def __init__(self, evaluation_transform_func: str, config: dict[str, Any]) -> None:
+        super().__init__(evaluation_transform_func, config)
+        self.seen_samples = 0
```
MaxiBoether, line 14: minor, but rename to samples_seen :D

modyn/utils/utils.py
```diff
@@ -143,3 +148,58 @@ def is_directory_writable(path: pathlib.Path) -> bool:
...
+def deserialize_function(serialized_function: str, func_name: str) -> Optional[Callable]:
+    """
+    Use this method to deserialize a particular python function given its string representation.
+
+    Args:
+        serialized_function: the function in plaintext.
+        func_name: the function name to be returned
+
+    Returns:
+        Optional[Callable]: None if function name could not
```
MaxiBoether, line 162: incomplete text

modyn/storage/internal/grpc/storage_grpc_servicer.py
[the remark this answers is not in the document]
MaxiBoether: Sorry, here I am not sure: Did you replace the existing version with your version in the TriggerSampleStorage? Where are the two functions different? Can we rely on one version maybe? What was the problem with the existing func?

MaxiBoether (changes requested):
Looking great! I did not review F1 yet (as it's not implemented). But everything else looks solid, also thanks for the thorough testing of everything. I had some comments, I think my biggest concern is the difference of the old partitioning method and your new one as I don't understand the differences (also when looking at the test changes) just yet. It's completely fine if we decide to use yours (no need to rollback!) but I just would like to understand what the differences are.

I suppose you were not able to e2e test the evaluator yet, that will come when we connect it to the supervisor

modyn/storage/internal/grpc/storage_grpc_servicer.py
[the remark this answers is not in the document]
robin-oester: I see, sorry for my bad wording. Yes, I replaced the existing version (`get_training_set_partition`) with my version.
In order to see the subtle difference, here is an example with 16 total elements and 3 workers:
The existing version assigns 6 elements to workers [0, 1] whereas worker [2] only gets 4 elements. It hence assigns less elements to the last worker in many cases.
My version assigns 6 elements to worker [0] and 5 elements to workers [1, 2]. I guess it's a matter of taste which version to prefer so I can rollback if you want.

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Yes, you are right. I initialized it by -1 as then we get consistent results if the dataloader does not return any batches.

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -654,14 +655,10 @@ def test_train(
...
     elapsed = 0
     while True:
         if not platform.system() == "Darwin":
-            if status_queue.qsize() == 1:
+            if status_queue.qsize() == 2:
```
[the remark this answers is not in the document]
robin-oester: Sure, no problem. This test had some minor errors (apart from the indentation issue, which resulted in the test always running through). Note that the model state is not part of the response to the `TrainerMessages.STATUS_QUERY_MESSAGE`. We have to fetch the model state explicitly via `TrainerMessages.MODEL_STATE_QUERY_MESSAGE`. The trainer server will hence create two responses (one for the status and one for the model state) resulting in 2 objects in the status_queue.

modyn/selector/internal/trigger_sample/trigger_sample_storage.py
```diff
@@ -184,44 +184,6 @@ def _get_all_samples(self, pipeline_id: int, trigger_id: int, partition_id: int)
             )
         ]
 
-    @staticmethod
-    def get_training_set_partition(
-        worker_id: int, total_workers: int, num_samples_trigger_partition: int
```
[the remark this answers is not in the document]
robin-oester: Yes, see the comment above.

modyn/tests/selector/internal/trigger_sample/test_trigger_sample_storage.py
```diff
@@ -154,42 +154,35 @@ def test_get_trigger_samples():
...
 
     result = TriggerSampleStorage(TMP_DIR).get_trigger_samples(1, 2, 3, 0, 1, 16)
     assert len(result) == 16
     assert result == expected_order
 
     result = TriggerSampleStorage(TMP_DIR).get_trigger_samples(1, 2, 3, 9, 10, 16)
-    assert len(result) == 0
-    assert not result
+    assert len(result) == 1
+    assert result == [expected_order[15]]
```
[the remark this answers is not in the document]
robin-oester: To give some context, we are checking the result of the last worker [9] if we have a total of 16 elements and 10 workers. The existing version gives 0 elements to the last worker (and 2 to workers [0, ..., 7], 0 to worker [8]). My version assigns 2 elements to workers [0, ..., 5] and 1 element to workers [6, 7, 8, 9].

modyn/config/schema/pipeline-schema.yaml
robin-oester: Yes, you are right. I initially thought that we will also need to transform the label for some metrics. However, this should not be necessary.

modyn/evaluator/internal/dataset/evaluation_dataset.py
robin-oester: Done, see #275 and TODOs.

modyn/evaluator/internal/metrics/accuracy.py
robin-oester: Done

modyn/tests/evaluator/internal/test_pytorch_evaluator.py
```diff
@@ -0,0 +1,307 @@
...
+        sleep(1)
+        elapsed += 1
+
+        if elapsed >= timeout:
+            raise TimeoutError("Did not reach desired queue state within time limit.")
+
+
+@patch("modyn.evaluator.internal.pytorch_evaluator.EvaluationDataset", MockEvaluationDataset)
+@patch.object(PytorchEvaluator, "_load_state")
+def test_train(load_state_mock: MagicMock):
```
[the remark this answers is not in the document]
robin-oester: Done

modyn/tests/evaluator/internal/test_pytorch_evaluator.py
```diff
@@ -0,0 +1,307 @@
...
+
+        if elapsed >= timeout:
+            raise TimeoutError("Did not reach desired queue state within time limit.")
+
+    with pytest.raises(ValueError):
+        evaluator.evaluate()
+
+    elapsed = 0
+    while not (query_status_queue.empty() and response_queue.empty()):
+        sleep(1)
```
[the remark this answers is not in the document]
robin-oester: Good catch! Done by setting to 0.1.

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,184 @@
...
+
+                with torch.autocast(self._device_type, enabled=self._amp):
+                    output = self._model.model(data)
+
+                    for metric in self._metrics:
+                        if isinstance(metric, AbstractDecomposableMetric):
+                            metric.batch_evaluated_callback(target, output, batch_size)
+
+                    if self._contains_holistic_metric:
+                        y_true.append(target.cpu())
```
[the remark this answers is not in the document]
robin-oester: Done

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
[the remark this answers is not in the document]
MaxiBoether: Alright, thank you, I remember now :D

modyn/storage/internal/grpc/storage_grpc_servicer.py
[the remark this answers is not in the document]
MaxiBoether: Hmm, I always had in mind that we are exactly trying to balance it that way already. But if your function is achieving that, I am fine with it :D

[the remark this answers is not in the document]
MaxiBoether: Thank you for the explanation.

modyn/evaluator/internal/pytorch_evaluator.py
MaxiBoether: I guess off by one depends on what you deem off by one, right? If you think index 0 is the first batch (which I guess I do), then it is not off by one in the trainer, is it?

MaxiBoether:
Ahh sorry I somehow was in flow of reviewing but you didn't request a review yet, did you? I will un-resolve the comments to not interfere with your process. Sorry about that

modyn/tests/selector/internal/trigger_sample/test_trigger_sample_storage.py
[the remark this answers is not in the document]
MaxiBoether: Alright, thanks for explaining that again. I am not actually sure what is the better strategy here but let's just keep yours, since it is more balanced overall (could also be valid to say that we use more workers to their maximum in the other version, but again not sure)

modyn/tests/selector/internal/trigger_sample/test_trigger_sample_storage.py
```diff
@@ -154,42 +154,35 @@ def test_get_trigger_samples():
...
-    assert result == [expected_order[8], expected_order[9], expected_order[10], expected_order[11]]
+    assert result == expected_order[8:12]
 
     result = TriggerSampleStorage(TMP_DIR).get_trigger_samples(1, 2, 3, 3, 4, 16)
     assert len(result) == 4
-    assert result == [expected_order[12], expected_order[13], expected_order[14], expected_order[15]]
+    assert result == expected_order[12:16]
 
-    result = TriggerSampleStorage(TMP_DIR).get_trigger_samples(1, 2, 3, 1, 3, 16)
+    result = TriggerSampleStorage(TMP_DIR).get_trigger_samples(1, 2, 3, 0, 3, 16)
```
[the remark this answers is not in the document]
MaxiBoether: kk

MaxiBoether:
Alright, unresolved everything that is not commited yet (left some stuff that was just discussion for me resolved)

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Yes but in both cases (evaluator, trainer server) we are logging the total number of processed batches. In the former version we logged the last `batch_number` seen. Since this is 0-based numbering, we are then off by 1 batch e.g we log 4 as this was the last `batch_number` however we have seen batches 0, 1, 2, 3, 4 - so 5 in total. It is very subtle and I can revert to the old version if you want.

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -763,7 +763,7 @@ def test_create_trainer_with_exception(
     timeout = 5
     elapsed = 0
     while query_status_queue.empty():
-        sleep(1)
+        sleep(0.1)
         elapsed += 1
```
robin-oester, line 767: Must be 0.1

modyn/evaluator/internal/metrics/f1_score.py
```diff
@@ -0,0 +1,104 @@
...
+                self.average = F1ScoreTypes(average_type_name)
+            else:
+                raise ValueError(
+                    f"Provided invalid average strategy {average_type_name}. "
+                    f"Must be one of: {', '.join(VALID_SCORE_TYPES)}"
+                )
+
+        self.pos_label = 1
+        if "pos_label" in config:
+            self.pos_label = config["pos_label"]
```
MaxiBoether, line 44: I am probably just blind, but is there - pipeline schema, comment, ... - an explanation of what `pos_label` is referring to? If not, maybe we could add a small comment here

modyn/evaluator/internal/pytorch_evaluator.py
MaxiBoether: No, no need to rollback (this is a minor detail anyways, as you said), I am just trying to understand (but this is not blocking for this PR - feel free to answer me on element and then resolve this). I still don't see how initializing this to -1 helps here and where exactly you are referring to with logging. During the first batch, `batch_number` will immediately be 0. So I don't see how it makes a difference, unless we do +1 somewhere to get the number of processed batches, but even then, I don't see how off by one is happening and how this would help, as we will always put batch_number = 0 in the queue, as we only write to the queue within the loop. But I think I am just missing something

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -0,0 +1,184 @@
...
+                except queue.Empty:
+                    pass
+
+                data: Union[torch.Tensor, dict]
+                if isinstance(batch[1], torch.Tensor):
+                    data = batch[1].to(self._device)
+                elif isinstance(batch[1], dict):
+                    data: dict[str, torch.Tensor] = {}  # type: ignore[no-redef]
+                    for name, tensor in batch[1].items():
+                        data[name] = tensor.to(self._device)
```
MaxiBoether, line 125: Could you open an issue that after merging #256 and this PR, we should also try to extract common code of the trainer and the evaluator into some helper class? to reduce code duplication, basically

MaxiBoether (approved):
Thank you! This is looking great. Regarding the minor -1/0 thing, let's discuss on Element. Let's merge #256 first though, I think it will be easier to merge #256 first and then this one, otherwise there will be many more conflicts

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Done, see #284

modyn/evaluator/internal/metrics/f1_score.py
MaxiBoether: this might be still open

modyn/evaluator/internal/pytorch_evaluator.py
MaxiBoether: thanks!

MaxiBoether (approved):
thanks for the (probably painful) merge! there was one small comment left open, feel free to address that accordingly + then just click on merge without me approving again :) don't want to block you while I sleep hehe

modyn/evaluator/internal/pytorch_evaluator.py
robin-oester: Ah, I forgot to answer. Yes you are right, this was previously not documented.
Now, it is done via configuration option comment in F1-score :)

modyn/evaluator/internal/metrics/f1_score.py
MaxiBoether: (feel free to resolve this comment and merge by yourself)

PR #272, opened 2023-06-19
Fix initial_setup script name

vGsteiger:


MaxiBoether (approved):
Thanks for finding this!

PR #273, opened 2023-06-22
Add CSVFileWrapper

francescodeaglio:
In this PR, the CSV file wrapper is introduced.

Bytes are decoded to strings, the required rows are extracted and labels are separated from samples. The labels are returned as integers, the samples are reassembled in csv-like format and converted to bytes. Pandas is not needed.

The bytes_parser_function takes care of decoding the bytes, splitting the string to separators and converting/casting the data.

Two open questions:
- is it necessary to validate the content of the file (that it is a csv, that all rows are the same size, that the 'label' column exists...)? Because this requires a preliminary reading of the file. The method is already written, it is just a question of whether to call it or not.
- can we assume that when an index list is requested (get_samples_from_indices), it is ordered and without duplicates? If not, it would complicate the function a bit to select the rows (nothing terrible, but with this assumption it is much more elegant)


vGsteiger:
I am a bit worried about the order of PR that we will go through with regards to #270 as this is a change in a rewritten part of the storage component. Most likely your PR will win the race but it would be great if you could already think about the port to C++ as well :) Thanks!

modyn/tests/storage/internal/file_wrapper/test_csv_file_wrapper.py
```diff
@@ -0,0 +1,197 @@
+import os
```
MaxiBoether, line 1: With Francesco's comment on the storage C++ rewrite in mind, can you please also add an integration test with csv? That would be amazing so we can test whether the C++ implementation will also work correctly :)

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -37,7 +37,8 @@ properties:
       sample_batch_size:
         type: number
         description: |
-          The size of a batch when requesting new samples from storage. All new samples are returned, however, to reduce the size of a single answer the keys are batched in sizes of `sample_batch_size`.
+          The size of a batch when requesting new samples from storage. All new samples are returned, however, to reduce 
+          the size of a single answer the keys are batched in sizes of `sample_batch_size`.
```
MaxiBoether, line 41: i think this might not be intended

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -49,7 +50,9 @@ properties:
       sample_table_unlogged:
         type: boolean
         description: |
-          This configures whether the table storing all samples is UNLOGGED (= high performance) or crash resilient. Defaults to True. For datasets with many samples (such as Criteo), this is recommended for highest insertion performance. In other scenarios, this might not be necessary.
+          This configures whether the table storing all samples is UNLOGGED (= high performance) or crash resilient. 
+          Defaults to True. For datasets with many samples (such as Criteo), this is recommended for highest insertion performance. 
+          In other scenarios, this might not be necessary.
```
MaxiBoether, line 55: is this intended?

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -99,15 +102,34 @@ properties:
...
                 byteorder:
                   type: string
                   description: |
-                    The byteorder when reading an integer from multibyte data in a binary file. Should either be "big" or "little".
+                    [BinaryFileWrapper] The byteorder when reading an integer from multibyte data in a binary file. 
+                    Should either be "big" or "little".
+                separator:
+                  type: string
+                  description: |
+                    [CsvFileWrapper] The separator used in the CSV file. The default is ";"
```
MaxiBoether, line 118: I think usually the default for csv is `,`, isn't it (after all, csv means comma separated values :D)

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -99,15 +102,34 @@ properties:
...
+                    For example, if the columns are "width", "height", "age", "label" you should set label_index to 3.
+                ignore_first_line:
+                  type: boolean
+                  description: |
+                    [CsvFileWrapper] If the first line is the table header, you can skip it setting this parameter to True. 
+                    Default is False.
+                encoding:
+                  type: string
+                  description: |
+                    [CsvFileWrapper] Encoding of your file. Default is utf-8
```
MaxiBoether, line 132:
```suggestion
                    [CsvFileWrapper] Encoding of the csv files. Default is utf-8.
```

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if "encoding" in file_wrapper_config:
+            self.encoding = file_wrapper_config["encoding"]
+        else:
+            self.encoding = "utf-8"
+
+        # check that the file is actually a CSV
+        self._validate_file_extension()
+
+        self._validate_file_content()
```
MaxiBoether, line 41: To answer your question: It would be great to make this configurable, whether these checks should be performed, and default to `True`.

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -99,15 +102,34 @@ properties:
...
                   type: string
                   description: |
-                    The byteorder when reading an integer from multibyte data in a binary file. Should either be "big" or "little".
+                    [BinaryFileWrapper] The byteorder when reading an integer from multibyte data in a binary file. 
+                    Should either be "big" or "little".
+                separator:
+                  type: string
+                  description: |
+                    [CsvFileWrapper] The separator used in the CSV file. The default is ";"
+                label_index:
```
MaxiBoether, line 119: This should be made a required parameter in schema. Right now, Modyn always assumes labeled data

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if "separator" in file_wrapper_config:
+            self.separator = file_wrapper_config["separator"]
+        else:
+            self.separator = ";"
+
+        if "label_index" not in file_wrapper_config:
+            raise ValueError(
+                "Please specify the index of the column that contains the label. "
+                "Use None if no column contains the label"
```
MaxiBoether, line 23: Right now we assume we always have labels so this should be schema-enforced

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+
+        labels: list[Optional[int]] = []
+
+        if self.label_index is None:
+            return [None] * self.get_number_of_samples()
+
+        for row in reader:
+            # labels are integer in modyn
+            int_label = int(row[self.label_index])
+            labels.append(int_label)
```
MaxiBoether, line 121: maybe use list comprehension here to define labels? might avoid resizing several times

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+        Returns:
+            list of byte-encoded rows
+
+        """
+        reader = self._get_csv_reader()
+
+        # Iterate over the rows and keep the selected ones
+        filtered_rows = []
+        for i, row in enumerate(reader):
+            if i in indices:
```
MaxiBoether, line 167: i think converting indices to a set is helpful to speed this up (also everywhere where you do these checks)

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+                # Remove the label, convert the row to bytes and append to the list
+                row_without_label = [col for j, col in enumerate(row) if j != self.label_index]
+                # the row is transformed in a similar csv using the same separator and then transformed to bytes
+                filtered_rows.append(bytes(self.separator.join(row_without_label), self.encoding))
+
```
MaxiBoether, line 172: i think this is what your question refers to. right now if we request [3,2,1] we still would get the results ordered as [1,2,3]. I think we cannot make this assumption (other classes do not), hence I would say we just need to reorder filtered_rows at the end according to the order in indices and return that

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
```diff
@@ -0,0 +1,198 @@
...
+        # Iterate over the rows and keep the selected ones
+        filtered_rows = []
+        for i, row in enumerate(reader):
+            if i in indices:
+                # labels are integer in modyn
+                int_label = int(row[self.label_index])
+                filtered_rows.append(int_label)
+
+        return filtered_rows
```
MaxiBoether, line 195: same comments as above apply

MaxiBoether (changes requested):
Thank you for the PR - added my comments and hopefully answered your questions :)

modyn/tests/storage/internal/file_wrapper/test_csv_file_wrapper.py
francescodeaglio: OK, I can certainly do that. Regarding the implementation in C++: I have never written a line of code and it would be quite a big overhead to write this class in C++. Are there any updates on Olga? Maybe she can take care of it since it was her PR

modyn/config/schema/modyn_config_schema.yaml
francescodeaglio: In my opinion, having these kilometre-long descriptions on a single line makes the yaml difficult to read. These two lines were very long and I broke them up with an enter while editing the schema for something else. However, if you prefer the descriptions as they were before I can rollback

modyn/config/schema/modyn_config_schema.yaml
francescodeaglio: OK, for me CSV has always been with the semicolon. I investigated a bit and apparently it's my Italian origin (as decimals are separated by the comma, like `1.32` is `1,32` so typically ; is used as separator. And this is handled automatically by the operating system depending on the OS language).

Anyway, now the comma is the default.

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
francescodeaglio: Ok, now it's a parameter `validate_file_content`

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
francescodeaglio: OK, but it is only a CSV file wrapper parameter. We have already discussed something similar in STB, we should either use conditional dependencies in the schema or raise the ValueError as here

modyn/config/schema/modyn_config_schema.yaml
francescodeaglio: Ok, I believed it was not an assumption since the typing of the `get_label` functions is Optional[int]. I can make it required and never return None

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
francescodeaglio: In this version yes, but now if I want to keep the same order I think it is more convenient to have a list

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
francescodeaglio: Ok, now the order is mantained

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
MaxiBoether: Ahh true, sorry, yeah then let's enforce it in init (that it is there and not None)

modyn/config/schema/modyn_config_schema.yaml
MaxiBoether: hmm I will take a look later, it looked weird on Github in the review system, but might actually look ok in the browser

modyn/tests/storage/internal/file_wrapper/test_csv_file_wrapper.py
MaxiBoether: No need for you to do it in C++ now. Before merging C++, all integration tests will need to pass through and before that we won't merge. No need for you to write that, that will happen inside the C++ PR

modyn/storage/internal/file_wrapper/csv_file_wrapper.py
francescodeaglio: Yep, now it checks that the value is available and that is a positive integer

modyn/config/schema/modyn_config_schema.yaml
MaxiBoether: yeah this is the reviewing system. it looked (on my macbook) like several random newlines all over the place :D

modyn/config/schema/modyn_config_schema.yaml
MaxiBoether: Good point, not sure why this is typed optional

MaxiBoether (approved):
LGTM! Thank you.

One note: Right now, we always read the entire file into memory, which we should generally avoid (`get_csv_reader` decodes the entire file into memory). This is due to our usage of `csv`. For now, this is fine, since we are rewriting storage in C++ anyways. In C++, at one point, we should try to not load everything into memory if possible, but instead only request the part of the file we need (if that is even possible with variable length data, but maybe using new lines or so there is some call). For now, this shall not be an issue.

PR #277, opened 2023-06-26
Refactor CoresetStrategy

francescodeaglio:
To remove duplicate code and allow for more flexibility, the existing downsampling strategies have been refactored. 

Prior to this PR, the way to use Loss as downsampling strategy was
```
-name: LossDownsamplingStrategy
-config: .....
```

Now you can use
```
-name: CoresetStrategy
-config:
           - downsampling_config:
                   -  strategy: "Loss"
                   -  ratio: 50
                   - sample_then_batch: True
           - presampling_config: 
                   -  strategy: "Empty"
                   -  ratio: 40
           - .... 
```

Where downsampling_strategy inherits from `AbstractDownsamplingStrategy` and presampling_strategy from `AbstractPresamplingStrategy`. In this way, you can combine as you wish presampling and downsampling while before presampling was always Random. 

New presampling strategies can be introduced with this refactor such as `RandomNoReplacement`, `RandomClassBalanced`, `RandomTriggerBalanced` and so forth. A presampling strategy is just a SQL query that returns the sampled ids. Also you can combine different presampling strategies simply using the SQL union statement. For example, you can train on all the points belonging to the last trigger plus a randomly presampled version of the whole dataset.

Also, you can combine different downsampling strategy using the `SchedulerDownsamplingStrategy`, where you use the `first_strategy` during the first `threshold` triggers and then the `second_strategy`. In the future, the switch might be performance based (if you are below a certain accuracy you can use a strategy and if above another one).
For example, the scheduler can be useful in the following two scenari:
 - start with all Data, then do GradNorm: in the original paper they mention that, at the beginning of training, the gradient norm is almost the same for all datapoints (all high loss since the weights are randomly initiated) so it's like sampling from a uniform distribution (no gain in computing the score, just use a stricter presampling)
 - start with STB, then switch to BTS: sample then batch might be very memory consuming (since you need to keep the score of all datapoints in memory) when the dataset grows. Hence, we can stop using it after some triggers and switch to the cheaper BTS

For example, the following pipeline produces the training pattern shown below
```
downsampling_config:
  strategy: "Scheduler"
  threshold: 3
  first_strategy:
    strategy: "Empty"
  second_strategy:
    ratio: 50
    sample_then_batch: True
    strategy: "Loss"
```
image

A similar approach can be used in the `SchedulerPresamplingStrategy`


modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -136,6 +135,7 @@ def inform_data(self, keys: list[int], timestamps: list[int], labels: list[int])
         Args:
             keys (list[str]): A list of keys of the data
             timestamps (list[int]): A list of timestamps of the data.
+            labels list[int]: A list of labels
```
francescodeaglio, line 138: Was missing in the docstring

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -352,7 +352,6 @@ def _persist_samples(self, keys: list[int], timestamps: list[int], labels: list[
             keys (list[str]): A list of keys of the data
             timestamps (list[int]): A list of timestamps of the data.
             labels (list[int]): A list of labels of the data.
-            database (MetadataDatabaseConnection): The database connection.
         """
```
francescodeaglio, line 355: No longer a function parameter

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+            yield [(sample, 1.0) for sample in samples]
+
+    def _get_data(self) -> Iterable[list[int]]:
+        """Returns all sample
+
+        Returns:
+            list[str]: Keys of used samples
+        """
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            if self.presampling_strategy.requires_trigger_dataset_size():
```
francescodeaglio, line 59: Many presampling strategies require to know the full dataset size. To avoid code repetition, this bool function is added to the abstract class.

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
```diff
@@ -18,36 +14,44 @@ class AbstractDownsampleStrategy(AbstractPresampleStrategy):
...
-        if "downsampling_ratio" not in config:
-            raise ValueError("Please specify downsampling_ratio to use downsampling methods")
+        if "ratio" not in downsampling_config:
+            raise ValueError("Please specify downsampling ratio to use downsampling methods")
 
-        self.downsampling_ratio = config["downsampling_ratio"]
+        self.downsampling_ratio = downsampling_config["ratio"]
 
-        if not (0 < self.downsampling_ratio < 100) or not isinstance(self.downsampling_ratio, int):
+        if not (0 < self.downsampling_ratio <= 100) or not isinstance(self.downsampling_ratio, int):
```
francescodeaglio, line 43: Now it can be 100 to support presampling-only strategies

modyn/selector/internal/selector_strategies/downsampling_strategies/empty_downsampling_strategy.py
```diff
@@ -0,0 +1,32 @@
+from modyn.selector.internal.selector_strategies.downsampling_strategies import AbstractDownsamplingStrategy
+
+
+class EmptyDownsamplingStrategy(AbstractDownsamplingStrategy):
```
francescodeaglio, line 4: This class is a little bit hacky but I think that it'ss better to abstract as much as possible and "fix" in this way

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
```diff
@@ -0,0 +1,97 @@
+from modyn.selector.internal.selector_strategies.downsampling_strategies import AbstractDownsamplingStrategy
+from modyn.selector.internal.selector_strategies.downsampling_strategies.utils import instantiate_downsampler
+
+
+class SchedulerDownsamplingStrategy(AbstractDownsamplingStrategy):
+    def __init__(self, downsampling_config: dict, maximum_keys_in_memory: int):
+        if "first_strategy" not in downsampling_config:
```
francescodeaglio, line 7: Each strategy is a dict (with exactly the same format of a "main" strategy)

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+        first_strategy = downsampling_config["first_strategy"]
+
+        if "second_strategy" not in downsampling_config:
+            raise ValueError(
+                "If you want to use a Downsampling Scheduler, please specify the first and the second strategy."
+            )
+        second_strategy = downsampling_config["second_strategy"]
+
+        if "threshold" not in downsampling_config:
+            raise ValueError("Please specify after how many epochs you want to switch strategy.")
```
francescodeaglio, line 20: to switch from a strategy to another one

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
```diff
@@ -0,0 +1,141 @@
+from typing import Optional
+
+from modyn.metadata_database.metadata_database_connection import MetadataDatabaseConnection
+from modyn.metadata_database.models import SelectorStateMetadata
+from modyn.selector.internal.selector_strategies.presampling_strategies import AbstractPresamplingStrategy
+from sqlalchemy import Select, asc, func, select
+from sqlalchemy.sql.expression import func as sql_func
+
+
+def get_fair_share(capacity: int, requests: list[int]) -> int:
```
francescodeaglio, line 10: I think it's a cool way to compute how many samples should each class get. Thanks CCA :)

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
```diff
@@ -0,0 +1,141 @@
...
+        samples_count = self._get_samples_count_per_label(next_trigger_id, tail_triggers)
+        target_size = self.get_target_size(sum(samples_count), limit)
+        fair_share = get_fair_share(target_size, samples_count)
+
+        # randomly permute samples class by class. Then use row_number (in the next query) to get N random samples for
+        # each class
+        subquery = select(
+            SelectorStateMetadata,
+            sql_func.row_number()
+            .over(partition_by=SelectorStateMetadata.label, order_by=func.random())  # pylint: disable=not-callable
```
francescodeaglio, line 52: Here the magic happens

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
```diff
@@ -0,0 +1,141 @@
...
+
+def get_fair_share_predicted_total(fair_share: int, requests: list[int]) -> int:
+    return sum(min(fair_share, req) for req in requests)
+
+
+class LabelBalancedPresamplingStrategy(AbstractPresamplingStrategy):
+    def __init__(self, presampling_config: dict, modyn_config: dict, pipeline_id: int, maximum_keys_in_memory: int):
+        super().__init__(presampling_config, modyn_config, pipeline_id, maximum_keys_in_memory)
+
+        self.force_required_target_size = presampling_config.get("force_required_target_size", False)
```
francescodeaglio, line 33: Target size is enforced -- if we want 500 points, then 500 points are returned (even if not perfectly balanced across classes)

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
```diff
@@ -0,0 +1,141 @@
...
+        self.force_label_balancing = presampling_config.get("force_label_balancing", False)
```
francescodeaglio, line 34: Each class gets exactly the same number of datapoints (particularly interesting for classes that have few samples)

francescodeaglio (commented):
I've added some comments to make the first review smoother.

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+        self.downsampling_strategy: AbstractDownsamplingStrategy = instantiate_downsampler(
+            config, maximum_keys_in_memory
+        )
+
+        if isinstance(self.presampling_strategy, EmptyPresamplingStrategy) and isinstance(
+            self.downsampling_strategy, EmptyDownsamplingStrategy
+        ):
+            raise ValueError(
+                "Using a Coreset Method without presampling and downsampling is useless. "
+                "You can use NewDataStrategy instead. "
```
MaxiBoether, line 36:
```suggestion
                "You did not specify any presampling and downsampling strategy for the CoresetStrategy. "
                "You can use NewDataStrategy instead. "
```

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+        """Returns all sample
+
+        Returns:
+            list[str]: Keys of used samples
+        """
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            if self.presampling_strategy.requires_trigger_dataset_size():
+                target_size = self._get_dataset_size()
+            else:
+                target_size = None
```
MaxiBoether, line 62:
```suggestion
            target_size = None
            
            if self.presampling_strategy.requires_trigger_dataset_size():
                target_size = self._get_dataset_size()               
```

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+                    SelectorStateMetadata.pipeline_id == self._pipeline_id,
+                    SelectorStateMetadata.seen_in_trigger_id >= self._next_trigger_id - self.tail_triggers
+                    if self.tail_triggers is not None
+                    else True,
+                )
+                .count()
+            )
+
+    def get_downsampling_strategy(self) -> str:
+        return self.downsampling_strategy.get_downsampling_strategy()
```
MaxiBoether, line 93: Maybe name this something like `downsampling_query()`? I think in general we should avoid `get` as a prefix in Python if not necessary, it's not very Pythonic (more like java and sometimes C++). Furthermore, I thought this would return a class or something :D so make it clear we return a query here

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get_downsampling_params(self) -> dict:
+        return self.downsampling_strategy.get_downsampling_params()
+
```
MaxiBoether, line 97: similar comment for these with `get`

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,102 @@
...
+        self._persist_samples(keys, timestamps, labels)
+
+    def _on_trigger(self) -> Iterable[list[tuple[int, float]]]:
+        for samples in self._get_data():
+            random.shuffle(samples)
+            yield [(sample, 1.0) for sample in samples]
+
+    def _get_data(self) -> Iterable[list[int]]:
+        """Returns all sample
```
MaxiBoether, line 53:
```suggestion
        """Returns all samples
```

MaxiBoether: (but either delete this docstring or make it consistent in the entire class I suppose)

modyn/config/schema/pipeline-schema.yaml
```diff
@@ -117,22 +117,40 @@ properties:
...
+                    description: |
+                       Strategy used to downsample the data. Available strategies: Loss, Gradnorm, Empty (no downsampling)
+                  sample_then_batch:
+                    type: boolean
+                    description: |
+                      If True, the samples are first sampled and then batched and supplied to the training loop. If False, the datapoints are first divided into batches and then sampled.
+                  ratio:
+                    type: number
+                    description: |
+                      Ratio post_sampling_size/pre_sampling_size. For example, with 160 datapoints and a ratio of 50 we keep 80. Must be in (0,100).
```
MaxiBoether, line 149: it can be (0,100], right?

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
```diff
@@ -18,36 +14,44 @@ class AbstractDownsampleStrategy(AbstractPresampleStrategy):
...
             raise ValueError("The downsampling ratio must be an integer in (0,100)")
 
-        self._requires_remote_computation = True
+        self.requires_remote_computation = True
+        self.maximum_keys_in_memory = maximum_keys_in_memory
+
+    def get_requires_remote_computation(self) -> bool:
+        return self.requires_remote_computation
```
MaxiBoether, line 50: no getter required, just access the public attribute :)

modyn/selector/internal/selector_strategies/coreset_strategy.py
MaxiBoether: oh this is the class name, right? i think thi should be more like self.downsampling_strategy.identifier (so no getter, I think in general here you should re-check all classes for getters) so it's clear what it means. Furthermore, maybe check if this wrapper is 100% necessary or not

modyn/selector/internal/selector_strategies/downsampling_strategies/gradnorm_downsampling_strategy.py
```diff
@@ -0,0 +1,6 @@
+from modyn.selector.internal.selector_strategies.downsampling_strategies import AbstractDownsamplingStrategy
+
+
+class GradNormDownsamplingStrategy(AbstractDownsamplingStrategy):
+    def get_downsampling_strategy(self) -> str:
+        return "RemoteGradNormDownsampling"
```
MaxiBoether, line 6: maybe we should add a comment somewhere (parent class) that this is the name of the identifier of the class that will be instantiated on the trainer server. goes together with my comment this should be named identifier or sth

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
MaxiBoether: so right now this is hardcoded to have a first and second strategy, correct? we could also make this an array, maybe? like
```
strategies: [
{ "name": GradNorm, ..}, {"name:" Loss}, ... 
],
scheduling_interval: 3 # advance every 3 epochs until reaching the last strategy
```

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if "ratio" in downsampling_config:
+            raise ValueError(
+                "SchedulerDownsamplingStrategy has no downsampling ratio. Specify the downsampling ratio"
+                "inside first_strategy and second_strategy."
+            )
+        downsampling_config["ratio"] = 100  # just to deal with super
+
+        if "sample_then_batch" in downsampling_config:
+            raise ValueError("SchedulerDownsamplingStrategy has no downsampling mode (sample_then_batch parameter).")
```
MaxiBoether, line 31: maybe be more explicit (you can make it like "if ratio in downsampling_config or if stb in config or ...: ValueError(please specify the strategy configs within the strategy objects, not on scheduler level)" or sth like that

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+        downsampling_config["sample_then_batch"] = True  # useless, just to avoid ValueErrors
```
MaxiBoether, line 32: mhm, I am not a big fan of this, it shows that somewhere the design could be improved, because it shows the inheritance hierarchy is not optimal, but I thought about it a bit and did not think of a better solution just yet

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+        self.counter = 0
+
+        # To count the elapsed epochs we can rely on how many calls were made to the functions below.
+        # Calls can be generated from two components: Supervisor (get_training_status_bar_scale)
+        # TrainerServer (everything else). SInce we have no guarantees on the order, and we want to avoid double
+        # counting, these two variables are introduced.
+        # Note that, from the trainer server, we just need to consider get_requires_remote_computation since the other
+        # calls are made sequentially
```
MaxiBoether, line 50: This seems a bit hacky. Could we instead have a RemoteScheduledStrategy or something, that is instantiated on the trainer server, instantiates both remote downsamplers, and then switches accordingly on the trainer server? Then we would not need to keep track of this at the trainer server. 

Furthermore, does the trainer server re request which strategy to use? I thought it would only do this during setup at the beginning

MaxiBoether: this would also eliminate the need for a getter. I think we should definitely do this a bit differently instead of hacking into these function calls to implement the scheduler. currently i think the trainer server side Remote thing would be the best solution but maybe you hav another idea

modyn/selector/internal/selector_strategies/downsampling_strategies/utils.py
```diff
@@ -0,0 +1,24 @@
+from modyn.selector.internal.selector_strategies.downsampling_strategies import AbstractDownsamplingStrategy
+from modyn.utils import dynamic_module_import
+
+
+def instantiate_downsampler(config: dict, maximum_keys_in_memory: int) -> AbstractDownsamplingStrategy:
```
MaxiBoether, line 5: please add some tests for this :)

modyn/selector/internal/selector_strategies/presampling_strategies/empty_presampling_strategy.py
```diff
@@ -0,0 +1,44 @@
+from typing import Optional
+
+from modyn.metadata_database.models import SelectorStateMetadata
+from modyn.selector.internal.selector_strategies.presampling_strategies import AbstractPresamplingStrategy
+from sqlalchemy import Select, asc, select
+
+
+class EmptyPresamplingStrategy(AbstractPresamplingStrategy):
```
MaxiBoether, line 8: maybe name NoPresampling instead?

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
MaxiBoether: hehe. I saw you wrote tests for this but did not fully check whether all potential cases are tested. if you are sure we covered everything great :)

MaxiBoether: (I actually need to learn more about this algorithm. I would have thought if you say it should be balanced, then you'd just check how many datapoints you have of each class, and then return as many from each class as you can to sustain the fair share. Like you have 10k points, 10% presampling, classes A, B, C with 400, 400, and 300 samples, you would return 300, 300, 300 and go even below the presampling ratio, but maybe this algorithm is a better solution to this problem

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
MaxiBoether: maybe you can add a class comment what the default behavior is with some short examples? like this uses the fair share algorithm or sth, which means XYZ. But optionally, you can enforce ABC. That would be super helpful for this class.

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
MaxiBoether: actually, we should use the number of classes for this pipeline. PR #269 (which should be merged today) will introduce a `num_classes` parameter. This way, if you actually have 3 classes, but only have received data of 2, we will not assume there are only 2 classes

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
```diff
@@ -0,0 +1,141 @@
...
+        new_requests = [requests[i] for i in range(len(requests)) if i not in below_fair_share]
+        return get_fair_share(new_capacity, new_requests)
+    return fair_share
+
+
+def get_fair_share_predicted_total(fair_share: int, requests: list[int]) -> int:
+    return sum(min(fair_share, req) for req in requests)
+
+
+class LabelBalancedPresamplingStrategy(AbstractPresamplingStrategy):
```
MaxiBoether, line 29: this (and the other strategies) are cool, but we should also test they work in postgres. could you add some meaningful integrationtests (there are already examples for selector integration tests) to all new presampling strategies, showing they work somewhat as inteded? (i know its hard to test randomness but we can do assertions for the stuff we can test)

MaxiBoether: please also leave this comment open for me as a reminder to go through this class again. I am very tired right now and cannot really check the queries too detailed

modyn/selector/internal/selector_strategies/presampling_strategies/random_presampling_strategy.py
```diff
@@ -0,0 +1,51 @@
+from typing import Optional
+
+from modyn.metadata_database.models import SelectorStateMetadata
+from modyn.selector.internal.selector_strategies.presampling_strategies.abstract_presampling_strategy import (
+    AbstractPresamplingStrategy,
+)
+from sqlalchemy import Select, asc, func, select
+
+
+class RandomPresamplingStrategy(AbstractPresamplingStrategy):
```
MaxiBoether, line 10: more like `UniformPresampling`, right?

modyn/selector/internal/selector_strategies/presampling_strategies/utils.py
```diff
@@ -0,0 +1,31 @@
+from modyn.selector.internal.selector_strategies.presampling_strategies import AbstractPresamplingStrategy
+from modyn.utils import dynamic_module_import
+
+
+def instantiate_presampler(
```
MaxiBoether, line 5: i think we can write some helper function `instantiate_class` like Robin wrote for functions in PR #269: https://github.com/eth-easl/modyn/blob/131ee591930b1b87b83e5bc57e508211d108737a/modyn/utils/utils.py#L170C2-L170C2

MaxiBoether: (because right now we do this in several places, and esp. here with downsamplers it is quite duplicated)

modyn/selector/selector.py
```diff
@@ -112,14 +112,11 @@ def uses_weights(self) -> bool:
         return self._strategy.uses_weights
 
     def get_selection_strategy_remote(self) -> tuple[bool, str, dict]:
-        assert not (
-            self._strategy._requires_remote_computation and not isinstance(self._strategy, AbstractDownsampleStrategy)
-        )
-
-        if self._strategy._requires_remote_computation and isinstance(self._strategy, AbstractDownsampleStrategy):
+        if isinstance(self._strategy, CoresetStrategy):
+            requires_remote_computations = self._strategy.get_requires_remote_computation()
```
MaxiBoether, line 116: small typo (no s to match the function call) - but we might want to change this again to not use this "hack"

MaxiBoether (changes requested):
Thank you for the code and trying to make the review as easy as possible! I have two main comments:

a) In some places, in this PR you used getters where i think we should be more Python-ic and  use members/at least remove the get

b) My main comment I suppose is the "hack" for implementing the schduler. I addressed that in the appropriate comment. 

Everything else is more like smaller individually addressable comments, I suppose :) Thank you again!

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
francescodeaglio: Yep, we can make also a PeriodicScheduler as you suggested. But I think that for now a threshold one is enough (I don't see any situation in which I would use a PeriodicScheduler). Or maybe I can turn this one in a list of strategies and a list of thresholds to make it more flexible.

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
francescodeaglio: I copied the implementation from an ETH course and my exercise session. I hope it's correct ahah

francescodeaglio: The solution you presented is the simple one, where you set the per_class size to the smallest class size (there's a query for that). 
My solution is used if we want exactly to match the target size (so returning 1000 samples in your example) and it's used to re-allocate the remaining samples. So, in your allocation, A and B would get 350 samples while C just 300 samples. The cool part of this solution is that you can just use LIMIT fair_share in the SQL query (since class C is below so it's unaffected by the LIMIT). There can be other solutions to reallocate the remaining samples but probably more complex on the SLQ side (like a limit different for each class)

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
francescodeaglio: Why do you think it's needed? I mean, if a class has 0 samples (so it's ignored by the query above) you can either get 0 samples for each class (useless) or get fair_share samples (so the empty class gets zero data points)

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
francescodeaglio: Yep, I plan to add docstrings to the "exotic" presamplers.

modyn/selector/internal/selector_strategies/coreset_strategy.py
francescodeaglio: I'd keep the wrapper to make clear who (presampler/downsampler) it's responsible of what. But, yes I can replace the inner call with a parameter

modyn/selector/internal/selector_strategies/coreset_strategy.py
francescodeaglio: I'll add docstrings in a later commit

modyn/config/schema/pipeline-schema.yaml
francescodeaglio: Oh yes, but just with the EmptyPresampling (otherwise doing downsampling it's just a waste of time). I think it's better to leave it like that

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
francescodeaglio: I too have thought about it for a while but this seems the best solution. The problem is that this class (and also Empty) are not real downsamplers. The correct thing would probably be to have a `TrueAbstractDownsamplingStrategy(AbstractDownsamplingStrategy)` class which is responsible of sample_then_batch, ratios and so forth. And then leave the AbstractDownsamplingStrategy with almost no methods. But idk if it's worth, what do you think?

modyn/selector/internal/selector_strategies/presampling_strategies/empty_presampling_strategy.py
francescodeaglio: Yep, maybe `NoPresampling` or `NonePresampling` are better names than just `EmptyPresampling`

modyn/selector/internal/selector_strategies/presampling_strategies/random_presampling_strategy.py
francescodeaglio: Yep, all the presampling is uniform since there's no way to so something more advanced directly on the DB. I think that Uniform is implicit, but I can change if you prefer

modyn/selector/internal/selector_strategies/presampling_strategies/utils.py
francescodeaglio: I'm not a big fan of exec since it's one the unsafest operation you can do. Technically, you can do something like
`deserialize_function("import shutil; shutil.rmtree("/");")` and everything is gone. I think that instantiate_{presampler/downsampler} requires some knowledge on the presampler/downsampler class so it's difficult to abstract and rely on a single function. I think I'll be able to abstract no more than 50% of the function

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
MaxiBoether: I think it is just a more general implementation that basically is identical and a bit less hardcoded

modyn/selector/internal/selector_strategies/presampling_strategies/label_balanced_presampling_strategy.py
MaxiBoether: ok, true, agreed.

modyn/config/schema/pipeline-schema.yaml
MaxiBoether: Sorry, I don't get your point here. 100% presampling and just using downsampling is valid, isn't it? And same would be < 100% presamling and 100% downsampling (which is identical to EmptyPresampling). So if we allow 100% for the EmptyPresampling strategy, then it should be like that in the description? Or would you never explicitly define an EmptPresamplingStrategy?

modyn/selector/internal/selector_strategies/presampling_strategies/utils.py
MaxiBoether: Sorry, I don't understand your point here. `exec` would not be required. What I meant is implementing a helper function `instantiate_class` (similar to Robin's `instantiate_function` which requires exec, but _class wouldn't). So basically having the entire `hasattr`/`getattr` stuff in one class (since this pattern is used in many parts in Modyn)

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
MaxiBoether: Could you please add something like this:

```python
# In the current design, the SchedulingStrategy implements a DownsamplingStrategy, which by definition expects a concrete strategy as its subclasses, i.e., each downsampling strategy should have certain properties, such as sample_then_batch. The SchedulingStrategy does not fit this design, since itself it is not a strategy, but an object choosing a strategy. If this causes further issues, the canonical refactoring would be that _every_ downsampling strategy uses a scheduler, but some just never change the behavior
```

MaxiBoether: so basically a comment explaining the problem and the solution, which, in my opinion, would be to have everything as a schedulingstrategy with potentially just one downsampler

modyn/selector/internal/selector_strategies/downsampling_strategies/scheduler_downsampling_strategy.py
francescodeaglio: Ok, makes sense. Now it's a list

modyn/selector/internal/selector_strategies/presampling_strategies/utils.py
francescodeaglio: ok done

modyn/config/schema/pipeline-schema.yaml
francescodeaglio: Ok, now it is (0,100]. Previously, it did not make sense to give the option of explicitly defining Empty, now it does (as it can be part of a scheduling with something else)

modyn/selector/internal/selector_strategies/coreset_strategy.py
francescodeaglio: I tried to reduce the gets as much as possible but here I can't remove them

modyn/selector/internal/selector_strategies/downsampling_strategies/utils.py
francescodeaglio: Added :)

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_gradnorm_downsampling_strategy.py
```diff
@@ -0,0 +1,36 @@
...
+        },
+        1000,
+    )
+
+    assert strat.downsampling_ratio == 10
+    assert isinstance(strat.remote_downsampling_strategy_name, str)
+    assert strat.requires_remote_computation
+
+
+def test_command_gradnorm():
```
MaxiBoether, line 19: sorry, what does that test do in comparison to the init test? it initializes it the same way, right?

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_loss_downsampling_strategy.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    assert strat.downsampling_ratio == 10
+    assert isinstance(strat.remote_downsampling_strategy_name, str)
+    assert strat.requires_remote_computation
+    assert strat.maximum_keys_in_memory == 1000
+
+
+def test_command_loss():
+    # Test init works
```
MaxiBoether, line 28: (same question for the other tests as above, won't repeat everywhere)

modyn/utils/utils.py
```diff
@@ -136,6 +136,25 @@ def package_available_and_can_be_imported(package: str) -> bool:
         return False
 
 
+def instantiate_class(class_module_name: str, class_name: str, *class_params: Any) -> Any:
```
MaxiBoether, line 139: thanks for creating this! could you also add one test to the utils test file? that would be amazing. thank you!

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def inform_data(self, keys: list[int], timestamps: list[int], labels: list[int]) -> None:
+        assert len(keys) == len(timestamps)
+        assert len(timestamps) == len(labels)
+
+        self._persist_samples(keys, timestamps, labels)
+
+    def _on_trigger(self) -> Iterable[list[tuple[int, float]]]:
+        # we don't want to shuffle if the downsampler needs the samples ordered by label. THe downsampler will take care
+        # of shuffling the samples.
```
MaxiBoether, line 35: Outdated comment?

modyn/selector/internal/selector_strategies/presampling_strategies/random_no_replacement_presampling_strategy.py
```diff
@@ -0,0 +1,94 @@
...
+        assert trigger_dataset_size is not None
+        assert trigger_dataset_size >= 0
+
+        target_size = self.get_target_size(trigger_dataset_size, limit)
+
+        # get the query to sample up to target size points that have not been used since last_complete_trigger
+        subq = self._get_query_sample_random(next_trigger_id, tail_triggers, target_size)
+
+        # update last_used_in_trigger of the samples selected in the subquery
+        self._update_last_used_in_trigger(next_trigger_id, subq)
```
MaxiBoether, line 35: Should we move this to the CoresetStrategy level instead of doing this in every presampling strategy? Or is there a reason that needs to be per persampling strategy?

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def _validate_threshold(self) -> None:
+        if not len(self.dowsampling_configs) == len(self.dowsampling_thresholds) + 1:
+            raise ValueError("You must specify a threshold for each transition")
+        if not sorted(self.dowsampling_thresholds) == self.dowsampling_thresholds:
+            raise ValueError("Thresholds must be monotonically increasing")
+        if not len(self.dowsampling_thresholds) == len(set(self.dowsampling_thresholds)):
+            raise ValueError("Thresholds must be unique")
+
+    def _get_new_downsampler_and_threshold(self) -> Tuple[AbstractDownsamplingStrategy, Optional[int]]:
```
MaxiBoether, line 30: new => next?

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        return next_downsampler, next_threshold
+
+    def _update_downsampler_if_needed(self, next_trigger_id: int) -> None:
+        if self.next_threshold is not None and next_trigger_id >= self.next_threshold:
+            self.downsampler_index += 1
+            self.current_downsampler, self.next_threshold = self._get_new_downsampler_and_threshold()
+
+    def get_requires_remote_computation(self, next_trigger_id: int) -> bool:
+        self._update_downsampler_if_needed(next_trigger_id)
```
MaxiBoether, line 48: This should not be here. A getter like that should not have a side effect of updating something because this implies we rely on implicit behavior that is not obvious to the caller. I am not sure what the reason for this is, but we cannot have an update in the getter. Rather, we need to do this somewhere explicitly

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_downsampling_strategy(self, next_trigger_id: int) -> str:
+        self._update_downsampler_if_needed(next_trigger_id)
+        assert hasattr(
+            self.current_downsampler, "remote_downsampling_strategy_name"
+        ), "Your downsampler must specify the remote_downsampling_strategy_name"
+        return self.current_downsampler.remote_downsampling_strategy_name
+
+    def get_downsampling_params(self, next_trigger_id: int) -> dict:
+        self._update_downsampler_if_needed(next_trigger_id)
```
MaxiBoether, line 59: similar comments here

MaxiBoether (changes requested):
Thank you so much! I guess architecture wise we have converged and I just had some smaller comments on details. After this round, I suppose we will be able to merge. :) Let me know if you have any questions

modyn/selector/internal/selector_strategies/downsampling_strategies/gradnorm_downsampling_strategy.py
MaxiBoether: is that addressed?

modyn/selector/internal/selector_strategies/presampling_strategies/random_no_replacement_presampling_strategy.py
francescodeaglio: Right now, only randomNoReplacement writes to the database to select samples. The other strategies do not use it, I think it would be much more inefficient otherwise

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
francescodeaglio: Yep, next sounds better :)

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_gradnorm_downsampling_strategy.py
francescodeaglio: The second test made sense the way the classes were structured before. Now I can combine them

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_loss_downsampling_strategy.py
francescodeaglio: Same as above, fixed

modyn/utils/utils.py
francescodeaglio: Added some tests

modyn/selector/internal/selector_strategies/coreset_strategy.py
francescodeaglio: yep :(

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_scheduler.py
```diff
@@ -0,0 +1,200 @@
...
+
+def test_instantiate_scheduler_just_one():
+    config = {"downsampling_config": {"strategy": "Loss", "sample_then_batch": True, "ratio": 50}}
+    scheduler = instantiate_scheduler(config, 100)
+    assert isinstance(scheduler.current_downsampler, LossDownsamplingStrategy)
+    assert scheduler.next_threshold is None
+
+
+def test_instantiate_scheduler_list():
+    config = {"downsampling_config": {"downsampling_list": get_configs(), "downsampling_thresholds": 7}}
```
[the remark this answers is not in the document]
francescodeaglio: See the comment above

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
francescodeaglio: OK I created a function (`inform_next_trigger`) and removed the call from the getter. Also, as it no longer receives any parameter I removed all the `get_XY` and transformed them into python-style using the @property decorator. So now every property looks like this 
```
@property
def downsampling_params(self) -> dict:
        return self.current_downsampler.downsampling_params
```

modyn/selector/internal/selector_strategies/presampling_strategies/abstract_balanced_strategy.py
```diff
@@ -0,0 +1,136 @@
...
+from typing import Optional
+
+from modyn.metadata_database.metadata_database_connection import MetadataDatabaseConnection
+from modyn.metadata_database.models import SelectorStateMetadata
+from modyn.selector.internal.selector_strategies.presampling_strategies.abstract_presampling_strategy import (
+    AbstractPresamplingStrategy,
+)
+from sqlalchemy import Select, asc, func, select
+from sqlalchemy.sql.expression import func as sql_func
+
```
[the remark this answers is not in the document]
francescodeaglio: Now there is a file for testing the `fair_share`

modyn/tests/selector/internal/selector_strategies/downsampling_strategies/test_gradnorm_downsampling_strategy.py
MaxiBoether: Ok top :)

modyn/utils/utils.py
MaxiBoether: thx

modyn/selector/internal/selector_strategies/presampling_strategies/random_no_replacement_presampling_strategy.py
MaxiBoether: ahh, yes, I somehow thought all strategies use it. then it makes sense to just have that here.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
MaxiBoether: Oh that sounds cool. I will check it out when I do the final review round

modyn/selector/selector.py
```diff
@@ -96,10 +96,11 @@ def get_number_of_samples(self, trigger_id: int) -> int:
 
     def get_status_bar_scale(self) -> int:
         # the status bar scale is only meaningful if we are using a Downsampling strategy. Otherwise, it's always 100%
-        if not isinstance(self._strategy, AbstractDownsampleStrategy):
+        if not isinstance(self._strategy, CoresetStrategy):
             return 100
 
-        return self._strategy.get_training_status_bar_scale()
+        self._strategy.inform_next_trigger()
```
MaxiBoether, line 102: Hm, so this is still not really solving the core problem that I mentioned: A getter should not have any side effects. 

The get_status_bar_scale function should do exactly that: Return the status bar scale. It should not matter if I ask it once, twice, thrice, etc. However, right now, every time we call this function, the CoresetScheduler makes one step forward. This should not happen. A getter should not change some properties. It might do calculations, but advancing something is not ok. Especially here, since we introduce the assumption that get_status_bar_scale is called with a certain pattern, and that assumption is hardcoded into the selector currently.

modyn/selector/selector.py
```diff
@@ -111,14 +112,11 @@ def uses_weights(self) -> bool:
         return self._strategy.uses_weights
 
     def get_selection_strategy_remote(self) -> tuple[bool, str, dict]:
-        assert not (
-            self._strategy._requires_remote_computation and not isinstance(self._strategy, AbstractDownsampleStrategy)
-        )
-
-        if self._strategy._requires_remote_computation and isinstance(self._strategy, AbstractDownsampleStrategy):
+        if isinstance(self._strategy, CoresetStrategy):
+            self._strategy.inform_next_trigger()
```
MaxiBoether, line 116: similarily, here. asking the selector what strategy to use should not inform somebody about something. this is a getter and not an "hey selector something happened" function. We have a on_trigger function that gets called per trigger. Why don't we use that?

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -0,0 +1,93 @@
...
+                "You should specify the thresholds to switch from a downsampler to another. "
+                "Use downsampling_thresholds"
+            )
+        list_of_thresholds = config["downsampling_config"]["downsampling_thresholds"]
+
+        # Allow users to specify just the threshold (and not [threshold]) when there are just 2 downsamplers.
+        if isinstance(list_of_thresholds, int):
+            assert (
+                len(list_of_downsamplers) == 2
+            ), "You must specify a list of thresholds if you want to use more than two downsamplers."
```
MaxiBoether, line 90: ValueError instead of assertion, since this is user input

MaxiBoether (changes requested):
Alright, thanks for the changes. The getter problem is unfortunately still not fully fixed (since now the thing lives in another getter), but see my comment for more details

modyn/selector/selector.py
francescodeaglio: I partially disagree with your comment. The function is idempotent, you can call it 1000 times in the same trigger and nothing changes. But I agree that at the code level it is not exceptional. I removed it from there and put it in `trigger()`

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
francescodeaglio: OK, it seemed strange to me too, but you told me assertion in one of the previous comments. Anyway, I changed

modyn/selector/selector.py
MaxiBoether: Hmm, as far as I was able to trace it, `inform_next_trigger` got forwarded up to the scheduler, which increased it sampler by one. However, putting it in `on_trigger` is cleaner in any case.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
MaxiBoether: ah sorry, I sometimes use the term assertion as "check if something is correct", in this sense it was meant in an abstract way and not as a literal assertion. thanks for changing.

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def inform_data(self, keys: list[int], timestamps: list[int], labels: list[int]) -> None:
+        assert len(keys) == len(timestamps)
+        assert len(timestamps) == len(labels)
+
+        self._persist_samples(keys, timestamps, labels)
+
+    def trigger(self) -> tuple[int, int, int]:
+        trigger_id, total_keys_in_trigger, num_partitions = super().trigger()
+        self.downsampling_scheduler.inform_next_trigger(self._next_trigger_id)
```
MaxiBoether, line 35: Could we also do that in `_on_trigger`? Then we would not need to overwrite trigger. `on_trigger` is called once per trigger and you could potentially call it before the for loop, couldn't you?

francescodeaglio: Actually no, because `on_trigger` is called before incrementing `next_trigger_id` but as it is written now the scheduler needs `next_trigger_id` to be already updated

MaxiBoether: Ok, then let's leave it like that. If you added a test that `trigger` on `CoresetStrategy` ensures that inform_next_trigger is called, then feel free to merge!

francescodeaglio: Now there is a test

MaxiBoether (approved):
Thank you, feel free to merge!

PR #279, opened 2023-06-27
Revert changes on environment and docker-compose made by #254

MaxiBoether:
Not sure how this happened, this basically reverted changes I made in #264.


MaxiBoether:
late night merging fun funfun. sorry for not getting back to the other folks if you see this, will write you tmrw

PR #285, opened 2023-06-29
Wild Time Benchmarks

francescodeaglio:
Introduce benchmarks [WildTIme](https://github.com/huaxiuyao/Wild-Time/tree/main). Some helper methods are taken from [Wild-Time-data](https://github.com/wistuba/Wild-Time-Data)

In this PR, we introduce:
- scripts for downloading data and formatting them so that they can be used by Modyn
- the models used for the 4 benchmarks (the same as those used in the original paper) 
- the `tokenizer` option in the pipeline to support NLP tasks. Using `bytes_parser_function`, a new `tokenizer` is instantiated each time (with GET to huggingface for each sample) and the training is obviously much slower (x15 or so based on a few runs). Using `tokenizer`, the `tokenizer` is instantiated only once and called sample by sample


benchmark/wildtime_benchmarks/example_pipelines/fmow.yaml
```diff
@@ -0,0 +1,52 @@
...
+  bytes_parser_function: |
+    from PIL import Image
+    import io
+    def bytes_parser_function(data: bytes) -> Image:
+      return Image.open(io.BytesIO(data)).convert("RGB")
+
+trigger:
+  id: DataAmountTrigger
+  trigger_config:
+    data_points_for_trigger: 2000
```
MaxiBoether, line 52: change to time trigger?

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -0,0 +1,69 @@
...
+            year_rows = []
+            for i in range(len(self._dataset[year][0]["title"])):
+                text = self._dataset[year][0]["title"][i].replace("\n", " ")
+                label = self._dataset[year][0]["category"][i]
+                csv_row = f"{text}\t{label}"
+                year_rows.append(csv_row)
+
+            # store the year file
+            text_file = os.path.join(self.path, f"{year}.csv")
+            with open(text_file, "w") as f:
```
MaxiBoether, line 59: specify utf 8 encoding when storing the files? since we explicitly load with utf 8 i suppose we should store with utf8 everywhere

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
```diff
@@ -0,0 +1,50 @@
...
+      limit: -1
+      reset_after_trigger: True
+data:
+  dataset_id: yearbook
+  transformations: []
+  bytes_parser_function: |
+    import torch
+    import numpy as np
+    def bytes_parser_function(data: bytes) -> torch.Tensor:
+      return torch.from_numpy(np.frombuffer(data, dtype=np.float32)).reshape(1, 32, 32)
```
MaxiBoether, line 45: any reason that we do it like that and not like fmow?

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
```diff
@@ -0,0 +1,50 @@
...
+  description: Example pipeline
+  version: 1.0.0
+model:
+  id: YearbookNet
+  config:
+    num_input_channels: 1
+    num_classes: 2
+training:
+  gpus: 1
+  device: "cpu"
```
MaxiBoether, line 12: I guess that should be consistent over all pipelines

environment.yml
```diff
@@ -29,6 +29,7 @@ dependencies:
   - types-protobuf
   - types-psycopg2
   - types-PyYAML
+  - transformers
```
MaxiBoether, line 32: please also add the huggingface channel to the channels list?

modyn/config/examples/modyn_config.yaml
```diff
@@ -45,6 +45,90 @@ storage:
...
         file_watcher_interval: 5,
         selector_batch_size: 2000000,
+      },
+      {
+        name: "yearbook",
+        description: "Yearbook Dataset from Wild-Time",
+        version: "0.0.1",
+        base_path: "/datasets/yearbook",
+        filesystem_wrapper_type: "LocalFilesystemWrapper",
+        file_wrapper_type: "BinaryFileWrapper",
```
MaxiBoether, line 55: Why do we use the binary file wrapper here? That one was built for criteo and I wonder why it maps to image data. Shouldn't we use the single sample file wrapper with regular image functionality?

modyn/models/articlenet/articlenet.py
```diff
@@ -0,0 +1,50 @@
...
+        attention_mask = data[:, :, 1]
+        hidden_state = super().__call__(
+            input_ids=input_ids,
+            attention_mask=attention_mask,
+        )[0]
+        pooled_output = hidden_state[:, 0]
+        return pooled_output
+
+
+class ArticleNetwork(nn.Module):
```
MaxiBoether, line 42: Could you add some documentation to the models, i.e., where are they coming from? Is the term ´ArticleNetwork´ something I/a user should know? Where is that coming from, why are we using a distilbeter featurizer before some linear layer? Usually, when building a very simple NLP DNN, huggingface is not necesasry, right? Why do we need/use it here?

modyn/models/articlenet/articlenet.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ArticleNet:
+    # pylint: disable-next=unused-argument
+    def __init__(self, model_configuration: dict[str, Any], device: str, amp: bool) -> None:
+        self.model = ArticleNetwork(**model_configuration)
+        self.model.to(device)
+
+
+class DistilBertClassifier(DistilBertForSequenceClassification):
```
MaxiBoether, line 15: what exactly does this do? I think without some documentation I am confused. You are overriding the main huggingface DistilBert classifier, not sure what the intention is here.

modyn/models/fmownet/fmownet.py
```diff
@@ -0,0 +1,29 @@
...
+        self.model = FmowNetModel(**model_configuration)
+        self.model.to(device)
+
+
+class FmowNetModel(nn.Module):
+    def __init__(self, num_classes: int) -> None:
+        super().__init__()
+        self.num_classes = num_classes
+        self.enc = densenet121(pretrained=True).features
+        self.classifier = nn.Linear(1024, self.num_classes)
```
MaxiBoether, line 21: is there no need for a softmax?

modyn/models/fmownet/fmownet.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class FmowNet:
+    # pylint: disable-next=unused-argument
+    def __init__(self, model_configuration: dict[str, Any], device: str, amp: bool) -> None:
+        self.model = FmowNetModel(**model_configuration)
+        self.model.to(device)
+
+
+class FmowNetModel(nn.Module):
```
MaxiBoether, line 16: (my comment on some documentation on model architecture + where it is coming from applies to all models, not going to copy it to the other files)

modyn/models/tokenizers/distill_bert_tokenizer.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, max_token_length: int = 300) -> None:
+        self.max_token_length = max_token_length
+        self.tokenizer = DistilBertTokenizer.from_pretrained("distilbert-base-uncased")
+
+    def __call__(self, sample: str) -> torch.Tensor:
+        tokens = self.tokenizer(
+            sample, padding="max_length", truncation=True, max_length=self.max_token_length, return_tensors="pt"
+        )
+        data = torch.stack((tokens["input_ids"], tokens["attention_mask"]), dim=2)
+        data = torch.squeeze(data, dim=0)  # First shape dim is always 1
```
MaxiBoether, line 15: is it? should we assert that and add a short reason why we expect it to be that way?

modyn/models/yearbooknet/yearbooknet.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+import torch
+from torch import nn
+
+
+class YearbookNet:
```
MaxiBoether, line 7: (I suppose you will adjust all of these accordingly when introducing the ModynModel with deepcore, right?)

modyn/storage/internal/grpc/storage_grpc_servicer.py
```diff
@@ -72,7 +72,10 @@ def Get(self, request: GetRequest, context: grpc.ServicerContext) -> Iterable[Ge
                 return
 
             samples: list[Sample] = (
-                session.query(Sample).filter(Sample.sample_id.in_(request.keys)).order_by(Sample.file_id).all()
+                session.query(Sample)
+                .filter(and_(Sample.sample_id.in_(request.keys), Sample.dataset_id == dataset.dataset_id))
```
MaxiBoether, line 76: does that fix #286?

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -149,13 +156,23 @@ def _unpack_data_tuple(self, data_tuple: Tuple) -> Tuple[int, bytes, int, Option
...
+        return key, tranformed_sample, label
 
     def end_of_trigger_cleaning(self) -> None:
         self._key_source.end_of_trigger_cleaning()
 
+    def _instantiate_tokenizer(self, tokenizer: str) -> Callable:
+        tokenizer_module = dynamic_module_import("modyn.models.tokenizers")
+        if not hasattr(tokenizer_module, tokenizer):
+            raise ValueError("Requested tokenizer is not available.")
+        return getattr(tokenizer_module, tokenizer)()
```
MaxiBoether, line 174: We can use the instantiateclass function we introduced, right?

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -56,6 +58,11 @@ def __init__(
         self._key_source = SelectorKeySource(self._pipeline_id, self._trigger_id, self._selector_address)
         self._uses_weights = None
 
+        # tokenizer for NLP tasks
+        self._tokenizer = None
+        if tokenizer != "":
+            self._tokenizer = self._instantiate_tokenizer(tokenizer)
```
MaxiBoether, line 64: I think we might be able to adjust the _transform function accordingly to include the tokenizer, right? So not only the potential torchvision transformers get composed, but if the tokenizer is there, we also include that in this function. then we don't need to check this everytime we yield a sample.

modyn/trainer_server/internal/utils/training_info.py
```diff
@@ -71,4 +71,9 @@ def __init__(
         else:
             self.seed = None
 
+        if request.HasField("tokenizer"):
+            self.tokenizer = request.tokenizer.value
+        else:
+            self.tokenizer = ""
```
MaxiBoether, line 77: Shall we use None instead of empty string?

MaxiBoether (changes requested):
Thanks! I had some comments, I hope they make sense.

benchmark/wildtime_benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,53 @@
...
+import os
+import pathlib
+from datetime import datetime
+
+import gdown
+
+DAY_LENGTH_SECONDS = 24 * 60 * 60
+
+
+def maybe_download(drive_id: str, destination_dir: str, destination_file_name: str) -> None:
```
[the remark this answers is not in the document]
francescodeaglio: The function is taken from wildtimedata (and also the name). I can change if you prefer

benchmark/wildtime_benchmarks/data_generation_fmow.py
```diff
@@ -0,0 +1,102 @@
...
+    args = parser.parse_args()
+
+    logger.info(f"Downloading data to {args.dir}")
+
+    downloader = FMOWDownloader(args.dir)
+    downloader.store_data()
+    downloader.clean_folder()
+
+
+def parse_metadata(data_dir: str) -> list:
```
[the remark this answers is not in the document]
francescodeaglio: It does not use any method/attribute/self of the class

benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
```diff
@@ -0,0 +1,48 @@
...
+    maximum_keys_in_memory: 10000
+    config:
+      limit: -1
+      reset_after_trigger: True
+data:
+  dataset_id: arxiv
+  bytes_parser_function: |
+    def bytes_parser_function(data: bytes) -> str:
+      return data.decode("utf-8")
+  tokenizer: DistilBertTokenizerTransform
```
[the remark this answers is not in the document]
francescodeaglio: All the models are taken from the paper, preprocessing included

benchmark/wildtime_benchmarks/example_pipelines/fmow.yaml
```diff
@@ -0,0 +1,52 @@
...
+    name: NewDataStrategy
+    maximum_keys_in_memory: 1000
+    config:
+      limit: -1
+      reset_after_trigger: True
+data:
+  dataset_id: fmow
+  transformations: [
+                "transforms.ToTensor()",
+                "transforms.Normalize([0.485, 0.456, 0.406], [0.229, 0.224, 0.225])",
```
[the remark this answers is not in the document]
francescodeaglio: It's part of the pipeline. See here from line 163 https://github.com/wistuba/Wild-Time-Data/blob/main/wild_time_data/datasets.py

modyn/models/articlenet/articlenet.py
francescodeaglio: The models and names are taken from Wild-time

modyn/storage/internal/grpc/storage_grpc_servicer.py
francescodeaglio: The training works and I haven't seen any side effects but I don't know if it fixes it completely

modyn/trainer_server/internal/dataset/online_dataset.py
francescodeaglio: Yep, I tried to keep the structure similar on purpose

benchmark/wildtime_benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
+# Wild-time datasets
```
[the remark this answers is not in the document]
francescodeaglio: Ok done

benchmark/wildtime_benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Datasets description
+
+### Yearbook
+The goal is to predict the sex given a  yearbook picture. The dataset contains 37189 samples collected from 1930 to 2013. Since timestamps in Modyn are based on 
+Unix Timestamps (so 0 is 1/1/1970) we have to remap the years to days. Precisely, the timestamp for pictures from 1930 is
+1/1/1970, then 2/1/1970 for the ones taken in 1931 and so forth. 
+
+### FMoW
+The goal is to predict land use (62 classes) given a satellite image . The dataset contains more than 100.000 samples collected from 2002 to 2017. Every picture is stored
```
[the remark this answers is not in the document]
francescodeaglio: Added

benchmark/wildtime_benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
+# Wild-time datasets
+
```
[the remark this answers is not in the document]
francescodeaglio: ok

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
francescodeaglio: Yep, I think it remained CPU because it is the only one you can run locally without problems

modyn/config/examples/modyn_config.yaml
francescodeaglio: Yearbook images are extremely small (32x32, B/W) so having one image per file is not efficient. Since the images and labels have a fixed size, I used BinaryFIleWrapper.

francescodeaglio: I had tried both versions and BinaryFileWrapper was faster

modyn/models/articlenet/articlenet.py
francescodeaglio: I don't know exactly what happens here either. The code is taken from the benchmark library, it is not mine. See here https://github.com/huaxiuyao/Wild-Time/blob/main/wildtime/networks/article.py

modyn/models/fmownet/fmownet.py
francescodeaglio: Softmax is tipically hidden inside the CrossEntropy loss function. See [this discussion](https://discuss.pytorch.org/t/softmax-cross-entropy-loss/125383) for example

modyn/models/fmownet/fmownet.py
francescodeaglio: I added a link pointing to the paper's implementation of each network

modyn/models/tokenizers/distill_bert_tokenizer.py
francescodeaglio: This is also code that is not mine that I tend to trust

modyn/models/yearbooknet/yearbooknet.py
francescodeaglio: Yep, this is the plan

modyn/trainer_server/internal/utils/training_info.py
francescodeaglio: Usual problem of having GRPC in the middle. If a tokenizer with `request.tokenizer=None` is sent by the supervisor, it is received as `request.tokenizer=""` and consequently gives problems. If you prefer `None` I can use it only on the trainer server side and add something like
```
if self.tokenizer == "":
   selt.tokenizer = None
```

modyn/trainer_server/internal/dataset/online_dataset.py
MaxiBoether: ahh it is not merged in main yet. maybe just merge #277 first then, since we need to order the PRs anyways, and then update that

benchmark/wildtime_benchmarks/benchmark_utils.py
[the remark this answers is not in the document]
MaxiBoether: sure, that would be great. As outlined in the comment on README, if a file (partially) stems from the wildtimedata repo, please add a comment at the top, to ensure we give sufficient credit

benchmark/wildtime_benchmarks/data_generation_fmow.py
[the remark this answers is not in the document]
MaxiBoether: it could be a staticmethod, I think. since it is definitely related to the class and only used there, it should be a static method of the class

benchmark/wildtime_benchmarks/example_pipelines/fmow.yaml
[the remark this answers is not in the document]
MaxiBoether: ok, thanks. I hope wild-time-data is correct, since it does not stem from the wild time paper authors, does it?

benchmark/wildtime_benchmarks/example_pipelines/fmow.yaml
MaxiBoether: Ah no, that makes sense, but maybe add a short comment here this needs to be adjusted for the wanted granularity and time trigger

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
MaxiBoether: ok even if we use the binary file wrapper here, do we not need any kind of image transformations, like Normalize? I thought normalizing the input as with MNIST or the FMOW data was pretty standard

modyn/models/tokenizers/distill_bert_tokenizer.py
MaxiBoether: Well we should not blindly merge code. In any case, do these models have some form of test? I think for the existing DLRM and ResNes Models, we at least check we can initialize the models. For small models here, we might want to also assert the output is as expected  (shape-wise)

benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _get_year_data(self, year: int) -> list[Tuple]:
+        images = torch.FloatTensor(
+            np.array(
+                [
+                    img.transpose(2, 0, 1)[0].reshape(*self.input_dim)
+                    for img in self._dataset[year][0]["images"]
+                ]
+            )
+        )
```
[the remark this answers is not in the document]
MaxiBoether: ah ok, maybe add a short comment like # Convert HWC to CHW or so. thanks!

modyn/models/articlenet/articlenet.py
MaxiBoether: As mentioned in another comment, I think we should at least somewhat understand what is hapenning here and if that is truly doing what we want it to do. The problem with research code is that we build a chain of trust but nobody really understands what somebody else has written. It would be glad if we could figure this out since many thing can go wrong here with huggingface stuff.

modyn/trainer_server/internal/dataset/online_dataset.py
francescodeaglio: Ok so something like this?
image

Note that the tokenizer is already written to be compatible with torch compose (see [here](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html))

MaxiBoether: Yes, looks good. Although we should rename the function to something like _setup_composed_transform

MaxiBoether:
ok, after round 2 of answering, I lost a bit of track of if I should review again or not. Let's consider all comments, and then request a review in github when it makes sense that I go through it end-to-end again. Thanks!

modyn/trainer_server/internal/dataset/online_dataset.py
francescodeaglio: Ok done

modyn/trainer_server/internal/utils/training_info.py
francescodeaglio: Ok true. Changed to `Optional[str]`

modyn/models/fmownet/fmownet.py
francescodeaglio: Is all hidden inside the densenet. In total, FMoWNet has 7007156 parameters

modyn/config/examples/modyn_config.yaml
francescodeaglio: I added a brief comment to the README

environment.yml
francescodeaglio: Ok, didn't know, thank you :) Now we install huggingface:transformers

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
francescodeaglio: I thought so too, but based on WildTimeData it seems that the only transformation is to create the tensor

francescodeaglio: See [Here](https://github.com/wistuba/Wild-Time-Data/blob/main/wild_time_data/datasets.py) line 187 vs line 163 (FMoW)

benchmark/wildtime_benchmarks/example_pipelines/fmow.yaml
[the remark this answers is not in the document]
francescodeaglio: Exactly, but they also apply the same transformation in wildTime. See [here](https://github.com/huaxiuyao/Wild-Time/blob/main/wildtime/data/fmow.py) line 31

benchmark/wildtime_benchmarks/data_generation_yearbook.py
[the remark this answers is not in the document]
francescodeaglio: Yep done

benchmark/wildtime_benchmarks/data_generation_fmow.py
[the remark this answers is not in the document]
francescodeaglio: Now it's a static method

benchmark/wildtime_benchmarks/benchmark_utils.py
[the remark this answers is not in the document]
francescodeaglio: Done

modyn/models/tokenizers/distill_bert_tokenizer.py
francescodeaglio: I added some comments to the class

modyn/models/articlenet/articlenet.py
francescodeaglio: I have added some documentation. To answer your questions:
- Wildtime
- No, it is the name used for the two NLP tasks that both deal with articles
- Typically classification for NLP tasks uses a featurizer (in this case DistilBert) to exploit the [CLS] (classification) token and from there project with a linear layer. See [here](https://towardsdatascience.com/text-classification-with-bert-in-pytorch-887965e5820f) for example
- I don't think it is a _very simple NLP DNN_, it has about 60 million trainable parameters
- Same reason we use torchvision for resnet

modyn/models/articlenet/articlenet.py
francescodeaglio: I have added some documentation. Shall we add unit tests? The problem is that it has to download the weights from huggingface

modyn/models/tokenizers/distill_bert_tokenizer.py
francescodeaglio: Also some tests

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -349,6 +354,7 @@ def start_training(
             "grad_scaler_configuration": TrainerServerJsonString(value=json.dumps(grad_scaler_config)),
             "epochs_per_trigger": epochs_per_trigger,
             "seed": seed,
+            "tokenizer": PythonString(value=tokenizer) if tokenizer is not None else None,
```
francescodeaglio, line 357: Just using PythonString(value=tokenizer) is not working since it's mapped to PythonString(value="") and not filtered in cleaned_kwargs

modyn/trainer_server/internal/dataset/online_dataset.py
MaxiBoether: leaving that open so we do not forget until the other pRs are merged

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
MaxiBoether: Hmm, ok, then I suppose it is correct that way.

modyn/models/articlenet/articlenet.py
MaxiBoether: I didn't realize at first we were actually using DistilBert, makes sense. Thank you.

modyn/models/articlenet/articlenet.py
MaxiBoether: For DistilBert, if we add tests, we should be able to somehow make the number of parameters small, like we did for DLRM (so it can be run on a CPU in testing) with random weights. If that is not possible, we should not add a test. For the other simpler models, we should add tests

modyn/supervisor/internal/grpc_handler.py
MaxiBoether: ok, thanks for letting me know!

environment.yml
```diff
@@ -29,6 +30,7 @@ dependencies:
   - types-protobuf
   - types-psycopg2
   - types-PyYAML
+  - huggingface:transformers
```
MaxiBoether, line 33: I think two colons are required, no?

modyn/models/articlenet/articlenet.py
```diff
@@ -0,0 +1,54 @@
...
+        input_ids = data[:, :, 0]
+        attention_mask = data[:, :, 1]
+        # DistilBert's forward pass
+        hidden_state = super().__call__(
+            input_ids=input_ids,
+            attention_mask=attention_mask,
+        )[
+            0
+        ]  # 0: last hidden state, 1: hiddent states, 2: attentions
+        pooled_output = hidden_state[:, 0]  # first token is the pooled output, which is the aggregated representation
```
MaxiBoether, line 41: just for my understanding: distilbert's output is redundant in the sense that we have pooled on idx 0 and then the individual representations again?

MaxiBoether (approved):
Thanks, I just had a few questions left, approving already!  After these are resolved, we can merge. Have a great time at the camp!

benchmark/wildtime_benchmarks/data_generation_fmow.py
```diff
@@ -0,0 +1,101 @@
...
+    downloader = FMOWDownloader(args.dir)
+    downloader.store_data()
+    downloader.clean_folder()
+
+
+class FMOWDownloader(Dataset):
+    time_steps = list(range(16))
+    input_dim = (3, 224, 224)
+    num_classes = 62
+    drive_id = "1s_xtf2M5EC7vIFhNv_OulxZkNvrVwIm3"
```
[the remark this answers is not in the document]
francescodeaglio: Hosted by the original dataset authors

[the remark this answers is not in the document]
francescodeaglio: Here you can find the same pointers https://github.com/huaxiuyao/Wild-Time/blob/main/wildtime/data/utils.py

environment.yml
francescodeaglio: Right

benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -0,0 +1,89 @@
...
+    def create_binary_file(data, output_file_name: str, timestamp: int) -> None:
+        with open(output_file_name, "wb") as f:
+            for tensor1, tensor2 in data:
+                features_bytes = tensor1.numpy().tobytes()
+                label_integer = tensor2.item()
+
+                features_size = len(features_bytes)
+                assert features_size == 4096
+
+                f.write(int.to_bytes(label_integer, length=4, byteorder="big"))
```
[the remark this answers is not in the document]
francescodeaglio: It seems to work correctly. This is a loaded image from the binary file
image

modyn/trainer_server/internal/dataset/online_dataset.py
francescodeaglio: Changed accordingly

Issue #280, opened 2023-06-27
Connect Evaluator


robin-oester:
After #269 is merged, we want to connect the evaluator to the system. In particular we want to:
- let supervisor interact with evaluator via gRPC calls
- evaluate models after each trigger on configured (static) dataset
- store evaluation results persistently
- test pipeline end-to-end
- (format used to store results should be readable for a visualization tool like TensorBoard)

MaxiBoether:
Closed by #287.

PR #287, opened 2023-07-05
Connect Evaluator

robin-oester:
As discussed in #280, we want to connect the evaluator to the system. Evaluation should be done optionally after each training.


modyn/config/schema/pipeline-schema.yaml
```diff
@@ -398,19 +407,9 @@ properties:
...
-          The device the model should be put on. In the future (#131), we might want this to be either "cpu" or "gpu"
-          and let the evaluator figure out the exact device, but for now, this really is the identifier of the device.
-      amp:
-        type: boolean
-        description: |
-          If True, automatic mixed precision will be used.
     required:
       - datasets
       - device
-      - amp
```
MaxiBoether, line 413: Do we default to false if this is not given?

modyn/protos/storage.proto
```diff
@@ -66,6 +68,14 @@ message GetDataPerWorkerResponse {
   repeated int64 keys = 1;
 }
 
+message GetDatasetSizeRequest {
+  string dataset_id = 1;
+}
+
+message GetDatasetSizeResponse {
+  int32 num_keys = 1;
```
MaxiBoether, line 76: this should be int64. For example with Criteo, these can be billions of samples. probably even uint64_t + a boolean success field instead

modyn/storage/internal/grpc/storage_grpc_servicer.py
```diff
@@ -243,6 +245,25 @@ def GetDataPerWorker(
                 if len(batch) > 0:
                     yield GetDataPerWorkerResponse(keys=[value[0] for value in batch])
 
+    # pylint: disable-next=unused-argument,invalid-name
+    def GetDatasetSize(self, request: GetDatasetSizeRequest, context: grpc.ServicerContext) -> GetDatasetSizeResponse:
```
MaxiBoether, line 249: Could you please also add an integration test for this? That would be great since we are rewriting the storage currently and we need to keep track end2end if the C++ implementation has the same behavior as the Python implementation :) Thank you!

modyn/protos/evaluator.proto
```diff
@@ -40,6 +40,7 @@ message EvaluateModelRequest {
 message EvaluateModelResponse {
   bool evaluation_started = 1;
   int32 evaluation_id = 2;
+  int32 dataset_size = 3;
```
MaxiBoether, line 43: see comment on storage proto file

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -434,3 +462,173 @@ def seed_selector(self, seed: int) -> None:
...
+                transform_list = []
+
+            if "label_transformer_function" in dataset:
+                label_transformer = dataset["label_transformer_function"]
+            else:
+                label_transformer = ""
+
+            bytes_parser_function = dataset["bytes_parser_function"]
+            batch_size = dataset["batch_size"]
+            dataloader_workers = dataset["dataloader_workers"]
```
MaxiBoether, line 500: I think there might be an indentation problem here. Should this be within the for loop for the datasets?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -434,3 +462,173 @@ def seed_selector(self, seed: int) -> None:
...
+        else:
+            model_config = "{}"
+
+        device = pipeline_config["evaluation"]["device"]
+        amp = pipeline_config["evaluation"]["amp"] if "amp" in pipeline_config["evaluation"] else False
+
+        evaluations: dict[int, EvaluationStatusTracker] = {}
+
+        # enforce unique dataset_ids
+        dataset_ids_seen = set()
```
MaxiBoether, line 482: I think you could also do something like
```
dataset_ids = [dataset["dataset_id"] for dataset in  pipeline_config["evaluation"]["datasets"]]
if len(set(dataset_ids)) < len(dataset_ids):
    raise some exception

```

to decouple this check from preparing the input to the grpc request (and avoid the indentation struggle below)

modyn/supervisor/internal/utils/evaluation_status_tracker.py
```diff
@@ -0,0 +1,39 @@
+from enlighten import Counter, Manager
+
+
```
MaxiBoether, line 3: I haven't checked in detail yet - Do you think there is potential to make some inheritance with the status bar classes by Francesco or is there no iheritance possible?

modyn/supervisor/internal/triggers/amounttrigger.py
```diff
@@ -1,4 +1,4 @@
-from modyn.supervisor.internal.trigger import Trigger
+from modyn.supervisor.internal.triggers.trigger import Trigger
```
MaxiBoether, line 1: if we are in the process of refactoring this, maybe export all triggers in the triggers init py and write

modyn/supervisor/internal/utils/evaluation_status_tracker.py
MaxiBoether: (supervisor_counter)

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -434,3 +462,173 @@ def seed_selector(self, seed: int) -> None:
...
+                evaluations[evaluation_id] = EvaluationStatusTracker(dataset_id, response.dataset_size)
+
+        return evaluations
+
+    def wait_for_evaluation_completion(self, training_id: int, evaluations: dict[int, EvaluationStatusTracker]) -> None:
+        if not self.connected_to_evaluator:
+            raise ConnectionError("Tried to wait for evaluation to finish, but not there is no gRPC connection.")
+
+        self.status_bar.update(demo=f"Waiting for evaluation (training = {training_id})")
+
```
MaxiBoether, line 556: Could you maybe explain what is happening here? I am not sure why we need a deque here (compared to the wait training to finish function), and if we can somehow make a common wait function out of these since they are very similar. If you could add some comments/explain the ideas here that would be great, right now it is hard for me to understand the idea (especially coming from the training thing without a deque)

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -434,3 +462,173 @@ def seed_selector(self, seed: int) -> None:
...
+            res: EvaluationResultResponse = self.evaluator.get_evaluation_result(req)
+
+            if not res.valid:
+                logger.warning(f"Cannot get the evaluation result for evaluation {evaluation_id}")
+                continue
+
+            with open(eval_directory / file_name, "w+", encoding="utf-8") as file:
+                file.write(f"evaluation_id: {evaluation_id}\n")
+                file.write(f"dataset_size: {evaluations[evaluation_id].dataset_size}\n")
+                file.write("metrics:\n")
```
MaxiBoether, line 632: could you maybe store it in some json file instead of this raw text format? maybe similar to the jsons created by the criteo evaluation script? i think that would be super useful (and needed to visualize the results)

modyn/supervisor/supervisor.py
```diff
@@ -51,6 +53,9 @@ def __init__(
         if not self.validate_pipeline_config():
             raise ValueError("Invalid pipeline configuration")
 
+        if not is_directory_writable(self.eval_directory):
+            raise ValueError("No permission to write to the evaluation results directory.")
+
```
MaxiBoether, line 58: we might also want to check that if evaluation is given, there are no duplicate dataset ids (instead of doing this when making the gRPC call, we just fail early) + there is at least one dataset (non empty lsit) in the input validation

MaxiBoether (changes requested):
Thank you! I had some small comments regarding structure, one potential error in indentation, and the status bar was not 100% clear to me yet (in comparison to training) and maybe there is potential to unify, although not sure for this case. Before merging, it would also be great if you could run a pipeline again to check if everything works end to end. Thank you so much and have a good trip back to Zurich!

modyn/config/schema/pipeline-schema.yaml
robin-oester: Yes, exactly

modyn/supervisor/internal/grpc_handler.py
robin-oester: Sure! Compared to training, here we have a parallel setup. The evaluator might receive several evaluation requests (for different evaluation datasets) and works on them in parallel. Hence, we also fetch the status in the supervisor for all ongoing evaluations in this `wait_for_evaluation_completion` function. I used a dequeue here in order to fetch the status of all evaluations sequentially in a round-robin manner. (We can also think of adding a thread for each evaluation at this point, which might simplify the unification process for waiting in training and evaluation).

robin-oester: Speaking of a common wait function: We introduced optional attributes in the `trainer_server.proto` (see `TrainingStatusResponse`) in #256. Was there a particular reason doing this?

modyn/supervisor/internal/grpc_handler.py
robin-oester: Yes, I would say that `batch_size` and `dataloader_workers` are dataset-dependent. However, I can change this if it doesn't make sense.

modyn/supervisor/internal/utils/evaluation_status_tracker.py
robin-oester: I also thought about this (also to only have one class). But IMHO there is not much potential as the evaluation status tracker is pretty simple and created on a per-evaluation-dataset basis. The `supervisor_counter` on the other hand changes states (btw. training and downsampling) during training, has several epochs and is only created once during training-invocation.

modyn/supervisor/supervisor.py
robin-oester: Done, see that check the evaluation pipeline config similar to the training config.

modyn/supervisor/internal/grpc_handler.py
robin-oester: Yes, done.

modyn/supervisor/internal/triggers/amounttrigger.py
robin-oester: Done

modyn/supervisor/internal/grpc_handler.py
robin-oester: I agree, done in supervisor.

modyn/protos/evaluator.proto
robin-oester: Done

modyn/storage/internal/grpc/storage_grpc_servicer.py
robin-oester: Sure :) done in next commit.

modyn/protos/storage.proto
robin-oester: Done

robin-oester:
Thank you for the feedback Maxi!
I made the changes accordingly. We can discuss about the implementation of `wait_for_evaluation_completion`. A possible solution would be to create a thread per evaluation (i.e per `EvaluateModelRequest`), which periodically fetches the current status of the evaluation and updates the progress bar. Then, we should also have consistent StatusResponses in evaluator.proto and trainer_server.proto. This should enable us to create a common wait function.

modyn/supervisor/internal/grpc_handler.py
MaxiBoether: Ahh you're right, sorry I am not sure what I mixed up here.

modyn/supervisor/internal/grpc_handler.py
MaxiBoether: We had some discussions on having a field set to 0 + a success flag vs optional. I don't really remember any other reason than potential clarity in the code I think. Is that a problem?

modyn/supervisor/internal/utils/evaluation_status_tracker.py
MaxiBoether: Ok, makes sense. Then let's keep that seperate.

modyn/storage/internal/grpc/storage_grpc_servicer.py
MaxiBoether: thank you!

integrationtests/storage/integrationtest_storage.py
```diff
@@ -89,6 +93,43 @@ def check_dataset_availability() -> None:
...
+
+
+def check_dataset_size_invalid() -> None:
+    storage_channel = connect_to_storage()
+
+    storage = StorageStub(storage_channel)
+    request = GetDatasetSizeRequest(dataset_id="unknown_dataset")
+    response: GetDatasetSizeResponse = storage.GetDatasetSize(request)
+
+    assert not response.success, "Dataset is available (even tough it should not be)."
```
MaxiBoether, line 114:
```suggestion
    assert not response.success, "Dataset is available (even though it should not be)."
```

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -434,3 +462,169 @@ def seed_selector(self, seed: int) -> None:
...
+        else:
+            model_config = "{}"
+
+        device = pipeline_config["evaluation"]["device"]
+        amp = pipeline_config["evaluation"]["amp"] if "amp" in pipeline_config["evaluation"] else False
+
+        evaluations: dict[int, EvaluationStatusTracker] = {}
+
+        for dataset in pipeline_config["evaluation"]["datasets"]:
+            dataset_id = dataset["dataset_id"]
```
MaxiBoether, line 482: could you maybe put the body of that loop into an extra function? sorry, I think i just did not manage to read it well and hence made the wrong claim that there is an indentation issue. if we have one function to call per dataset it should be clearer

MaxiBoether (approved):
Thank you! I had very minor comments, after that feel free to merge. Looks good to me!

modyn/supervisor/internal/grpc_handler.py
robin-oester: I see, thank you! No, this is not a problem. I've used the same approach for evaluation. Note, however, that I did not make a common waiting function. They are indeed (very) similar but are using other classes for e.g status bar, training has downsampling and evaluation handles several `EvaluationStatusRequest` in parallel.

Comment done.

integrationtests/storage/integrationtest_storage.py
robin-oester: Done

modyn/supervisor/internal/grpc_handler.py
robin-oester: Sure, done!

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -544,18 +495,79 @@ def start_evaluation(self, trained_model_id: int, pipeline_config: dict) -> dict
 
         return evaluations
 
+    @staticmethod
+    def _get_evaluate_model_request(
```
MaxiBoether, line 499: maybe rename to prepare request?

MaxiBoether (approved):
One minor comment :) thanks for the last changes!

modyn/supervisor/internal/grpc_handler.py
robin-oester: Sure, done. I tested the pipeline end-to-end and everything seemed to work out.

PR #288, opened 2023-07-07
Introduce a super inefficient per class dataloader

francescodeaglio:
To support some downsampling strategies, we need to be able to get samples label by label. In this PR, an inefficient dataloader is proposed.
These are the updates in this PR:

- new `PerClassOnlineDataset` in which samples are yielded only if  `label == self.filtered_label`. All the other samples are ignored. This attribute can be changed to filter on a different label
- support in the sample then batch function (grpc call to get all the distinct labels, _for_ loop to work label by label ..)
- support in the AbstractRemoteDownsampler (attribute `requires_data_label_by_label` and abstract function `_downsampler.inform_end_of_current_label` to handle the end of the label


modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -398,3 +398,20 @@ def _persist_samples(self, keys: list[int], timestamps: list[int], labels: list[
...
+                .filter(
+                    SelectorStateMetadata.pipeline_id == self._pipeline_id,
+                    SelectorStateMetadata.seen_in_trigger_id >= self._next_trigger_id - self.tail_triggers
+                    if self.tail_triggers is not None
+                    else True,
+                )
+                .distinct()
+                .all()
+            )
+            available_labels = [result_tuple[0] for result_tuple in result]
```
MaxiBoether, line 415: I suppose it is not a problem that these lists all labels that have been seen and it could be (e.g. due to random presampling) that not all labels are actually part of the TriggerTrainingSet in the end?

modyn/protos/selector.proto
```diff
@@ -74,6 +76,14 @@ message GetNumberOfPartitionsRequest {
 
 message NumberOfPartitionsResponse { int32 num_partitions = 1; }
 
+message GetAvailableLabelsRequest {
```
MaxiBoether, line 79: Could you please add an integration test for this :)?

modyn/selector/internal/selector_manager.py
```diff
@@ -135,6 +135,12 @@ def get_number_of_partitions(self, pipeline_id: int, trigger_id: int) -> int:
 
         return self._selectors[pipeline_id].get_number_of_partitions(trigger_id)
 
+    def get_available_labels(self, pipeline_id: int) -> list[int]:
```
MaxiBoether, line 138: for this (and the other new functions), please add some unit tests as well. thank you!

modyn/trainer_server/internal/dataset/data_utils.py
```diff
@@ -47,3 +48,19 @@ def prepare_dataloaders(
     # TODO(#50): what to do with the val set in the general case?
     val_dataloader = None
     return train_dataloader, val_dataloader
+
+
```
MaxiBoether, line 52: please mention issue 289 here :)

modyn/trainer_server/internal/dataset/per_class_online_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+            trigger_id,
+            dataset_id,
+            bytes_parser,
+            serialized_transforms,
+            storage_address,
+            selector_address,
+            training_id,
+        )
+
+        self._filtered_label = None
```
MaxiBoether, line 36: any reason we do not set that in the constructor?

modyn/trainer_server/internal/dataset/per_class_online_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+            for idx, data_tuple in self._get_data_iterator(keys, data, labels, weights):
+                key, sample, label, weight = self._unpack_data_tuple(data_tuple)
+
+                if partition < self._num_partitions - 1 and idx == fetch_next_partition_idx:
+                    # TODO(#175) in case this blocks training
+                    new_keys, new_data, new_labels, new_weights = self._get_data(
+                        worker_id=worker_id, partition_id=partition + 1
+                    )
+                if label == self._filtered_label:
+                    yield self._yield_samples(key, sample, label, weight)
```
MaxiBoether, line 81: hmm alternatively, couldnt we use the regular onlinedataset, and give it the optional parameter filtered label, and if it is not none, then we do this check? then we avoid introducing a new class. anything speaks against that?

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsample_strategy.py
```diff
@@ -62,3 +66,7 @@ def inform_samples(self, sample_ids: list[int], forward_output: torch.Tensor, ta
     @abstractmethod
     def select_points(self) -> tuple[list[int], torch.Tensor]:
         raise NotImplementedError
+
+    def inform_end_of_current_label(self) -> None:
```
MaxiBoether, line 70: abstractmethod or not? you have a if that raises not implemented, not sure what this means.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -487,28 +495,32 @@ def downsample_trigger_training_set(self) -> None:
...
-                self._downsampler.inform_samples(sample_ids, model_output, target)
+        if self._downsampler.requires_data_label_by_label:
+            available_labels = self._get_available_labels_from_selector()
+            per_class_dataloader = prepare_per_class_dataloader_from_online_dataset(
+                self._train_dataloader.dataset, self._batch_size, self._num_dataloaders
+            )
+            number_of_samples = 0
+            batch_number = 0
+            for label in available_labels:
+                per_class_dataloader.dataset._filtered_label = label
```
MaxiBoether, line 515: ah now I see. usually a leading underscore indicates a private attribute. we should pass an initial filtered label in the constructor and make this attribute public

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -533,10 +545,39 @@ def downsample_trigger_training_set(self) -> None:
...
 
         self.update_queue(AvailableQueues.DOWNSAMPLING, batch_number, number_of_samples, training_active=True)
+        # set the model to train
+        self._model.model.train()
+
+    def _iterate_dataloader_and_compute_scores(
+        self,
+        dataloader: torch.utils.data.DataLoader,
+        previous_batch_number: int = 0,
+        previous_number_of_samples: int = 0,
```
MaxiBoether, line 555: could you add some documentation why this is needed? it appears some per class vs non per class thing but its not obvious to me what is happening here

MaxiBoether (changes requested):
Thank you! I had some small comments, I suppose after the next round we can merge :)

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
francescodeaglio: Not a problem, simply no points are returned

francescodeaglio: To actually know which classes are selected, we could have to use more inefficient presampling methods (like `last used in trigger` as in the other PR). I don't think it is worth it.

modyn/trainer_server/internal/dataset/per_class_online_dataset.py
francescodeaglio: In my opinion, it is not worth changing the online dataset for a temporary feature. It is already complex enough, I would avoid adding to it.

I have now changed the iteration loop to 
```
data_tuple = self._get_data_tuple(key, sample, label, weight)
if data_tuple is not None:
   yield data_tuple
```

So we just have to overload the method `_get_data_tuple` and filter there

modyn/trainer_server/internal/trainer/pytorch_trainer.py
francescodeaglio: Ok

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsample_strategy.py
francescodeaglio: Weird inheritance. Now should be clearer

modyn/trainer_server/internal/dataset/per_class_online_dataset.py
francescodeaglio: Now toy can specify the initial label in the constructor and then change it at runtime

modyn/trainer_server/internal/dataset/per_class_online_dataset.py
francescodeaglio: So basically this is the only difference
```
def _get_data_tuple(self, key: int, sample: bytes, label: int, weight: Optional[float]) -> Optional[Tuple]:
    assert self.filtered_label is not None

    if self.filtered_label != label:
        return None
    return super()._get_data_tuple(key, sample, label, weight)
```

modyn/trainer_server/internal/trainer/pytorch_trainer.py
francescodeaglio: Sure

modyn/protos/selector.proto
francescodeaglio: Added

modyn/selector/internal/selector_manager.py
francescodeaglio: done

MaxiBoether (approved):
LGTM - 2 very minor comments, feel free to merge afterwards!

integrationtests/selector/integrationtest_selector.py
```diff
@@ -624,10 +625,71 @@ def test_many_samples_unevenly_distributed():
     assert len(total_samples) == 10001
 
 
+def test_get_available_labels(reset_after_trigger: bool):
+    selector_channel = connect_to_selector_servicer()
+    selector = SelectorStub(selector_channel)
+    # We test without reset, i.e., after an empty trigger we get the same data
```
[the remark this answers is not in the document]
francescodeaglio: Yes, it was just `reset = false` at the beginning. Removed.

MaxiBoether:
thanks for fixing, feel free to merge. note that robin just merged the evaluator, which might come in handy for evaluating the pipelines and collecting metrics :)

Issue #281, opened 2023-06-27
Provide Visualization for Evaluation Metrics


robin-oester:
Implement basic functionality to provide the user with training/evaluation metrics visualization. Regarding tools, we need to think about what makes sense (options are Tensorboard, W&B, etc.). The basic version should probably include:
- graphical interface where we can select a specific trigger from a pipeline
- visualize evaluation metrics over time
- should be able to compare different model versions w.r.t evaluation metrics
Issue depends on #269 and #280.

PR #292, opened 2023-07-17
Add Tensorboard Support

robin-oester:
Implement basic functionality to provide the user with evaluation metrics visualization as described in #281. Tensorboard should run on the supervisor and visualize evaluation metrics over time.


modyn/supervisor/internal/utils/evaluation_result_writer.py
```diff
@@ -0,0 +1,85 @@
...
+    @staticmethod
+    @abstractmethod
+    def get_name() -> str:
+        """
+        Defines the name used in the pipeline config for this result writer
+        """
+        raise NotImplementedError()
+
+
+class JsonResultWriter(AbstractEvaluationResultWriter):
```
robin-oester, line 48: Should I put these `ResultWriters` in separate files? Since they are rather short and simple I put them together.

modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -351,6 +351,15 @@ properties:
         - port
         - ftp_port
         - offline_dataset_directory
+  tensorboard:
+    type: object
+    properties:
+      port:
+        type: string
+        description: |
+          The port on which tensorboard is run.
```
MaxiBoether, line 360: If not given, tensorboard is disabled?

MaxiBoether: (if the entire tensorboard object is not given, I mean. Port is required)

modyn/supervisor/internal/utils/evaluation_result_writer.py
MaxiBoether: I understand they are short but sometimes if you don't know where what is defined it is helpful if we maintain a class-file mapping I think. So if it's not too big of an effort, please move it to separate files. Thanks!

modyn/supervisor/supervisor.py
```diff
@@ -102,15 +117,21 @@ def validate_pipeline_config_schema(self) -> bool:
...
-    def _validate_evaluation_options(evaluation_config: dict) -> bool:
+    def _validate_evaluation_options(self, evaluation_config: dict) -> bool:
         is_valid = True
 
         dataset_ids = [dataset["dataset_id"] for dataset in evaluation_config["datasets"]]
         if len(set(dataset_ids)) < len(dataset_ids):
             logger.error("Dataset ids must be unique in evaluation")
             is_valid = False
 
+        if "result_writers" in evaluation_config:
```
MaxiBoether, line 128: is this not required? does evaluation without storing results make sense?

modyn/supervisor/supervisor.py
```diff
@@ -102,15 +117,21 @@ def validate_pipeline_config_schema(self) -> bool:
...
+            writer_names = set(evaluation_config["result_writers"])
+            supported_names = {writer.get_name() for writer in self.supported_evaluation_result_writers}
```
MaxiBoether, line 130: We could also handle this similary to some other cases within Modyn where we have an enum that maps to class names. Would that make sense? not sure what is the best approach

modyn/supervisor/supervisor.py
```diff
@@ -222,6 +243,13 @@ def get_dataset_selector_batch_size(self) -> None:
     def validate_system(self) -> bool:
         return self.dataset_available() and self.grpc.trainer_server_available()
 
+    def _run_tensorboard(self, port: int) -> None:
+        logger.info(f"Starting up tensorboard on port {port}.")
+        try:
+            subprocess.call(f"tensorboard --logdir {self.eval_directory} --bind_all --port={port}".split(" "))
```
MaxiBoether, line 249: Can we avoid subprocess? Rather, I would prefer to run tensorboard within Python, if possible. I did a quick search and found that: https://stackoverflow.com/questions/42158694/how-to-run-tensorboard-from-python-scipt-in-virtualenv

modyn/supervisor/supervisor.py
```diff
@@ -359,7 +387,17 @@ def _run_training(self, trigger_id: int) -> None:
...
             self.grpc.wait_for_evaluation_completion(self.current_training_id, evaluations)
-            self.grpc.store_evaluation_results(self.eval_directory, self.pipeline_id, trigger_id, evaluations)
+
+            writer_names: set[str] = set()
+            if "result_writers" in self.pipeline_config["evaluation"]:
+                writer_names.update(self.pipeline_config["evaluation"]["result_writers"])
+
+            writers: list[AbstractEvaluationResultWriter] = []
+            for writer_handler in self.supported_evaluation_result_writers:
+                if writer_handler.get_name() in writer_names:
```
MaxiBoether, line 397: hmm, here we iterate over all handlers, check if we want them, and instantiate them. an alternative would be to iterate over all specified handlers, and instantiate them (e.g. with a string -> class map). WDYT?

MaxiBoether (changes requested):
Thank you for the PR! I had some comments but overall looks good :)

modyn/supervisor/internal/utils/evaluation_result_writer.py
robin-oester: I see. It's also better if we want to extend the system at some point. Moved them in the next commit.

modyn/supervisor/supervisor.py
robin-oester: I agree, changed it to required and to contain at least one element.

modyn/supervisor/supervisor.py
robin-oester: I agree. Had some issues with excessive logging of tensorboard. This should now work :)

modyn/supervisor/supervisor.py
robin-oester: Sure! I did some changes in order to solve this differently (see also other comment).

modyn/supervisor/internal/evaluation_result_writer/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""Supervisor module. The supervisor initiates a pipeline and coordinates all components.
```
MaxiBoether, line 1: copy & pasted docstring :)

MaxiBoether (approved):
sorry for getting back so late :) LGTM

modyn/supervisor/internal/evaluation_result_writer/__init__.py
robin-oester: Yes, I see. But note that this docstring is used by every submodule of the supervisor (e.g supervisor/internal/triggers or /utils)

PR #304, opened 2023-10-04
Introduce policies from DeepCore

francescodeaglio:
With this PR, we introduce several new downsampling methods: `Craig`, `GradMatch`, `Submodular` (FacilityLocation, GraphCut and LogDeterminant), `KcenterGreedy` and `Uncertainty` (Margin, Entropy, and LeastConfidence). These methods are adapted from [DEEPCORE](https://github.com/PatrickZH/DeepCore)

The following functionalities are implemented:
- `CoresetSupportingModule` support in both BTS and STB mode. Embeddings are registered and provided in the `inform_samples` method if required.
- addition of the `device` in the constructor of the downsampling methods. For example, kcenter uses a different implementation depending on whether it runs on CPU or GPU.
- implementation of the techniques mentioned above. Many of which, having a common behaviour, are abstracted with the `AbstractMatrixDownsamplingStrategy` class
- fixing some bugs found by running the experiments (typically moving weights to the correct device)

All methods were tested by comparing the results obtained with deepcore in various controlled experiments. The available tests serve this purpose.


modyn/common/trigger_sample/trigger_sample_storage.py
```diff
@@ -232,6 +232,10 @@ def _parse_file(self, file_path: Path) -> np.ndarray:
         Returns:
             list[tuple[int, float]]: List of trigger samples.
         """
+
+        if not os.path.isfile(file_path):
+            return np.ndarray(0, dtype="i8,f8")
```
francescodeaglio, line 237: When there are few samples, it may happen that some workers don't have any samples to store. Therefore, they do not write anything; you get an error if you try to read their file. This way, it returns an empty array if the worker has not written anything.

modyn/trainer_server/internal/dataset/local_dataset_writer.py
```diff
@@ -62,14 +62,10 @@ def _store_triggersamples_impl(
     def inform_samples(self, sample_ids: list, sample_weights: torch.Tensor) -> None:
         # map the two input lists to the desired format
         assert self.output_samples_list is not None
-        samples_list = np.empty(len(sample_ids), dtype=np.dtype("i8,f8"))
-
-        for i, _ in enumerate(sample_ids):
-            samples_list[i] = (sample_ids[i], sample_weights[i])
```
francescodeaglio, line 65: The double iteration is useless

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -0,0 +1,110 @@
...
+import torch
+from modyn.trainer_server.internal.trainer.remote_downsamplers.abstract_per_label_remote_downsample_strategy import (
+    AbstractPerLabelRemoteDownsamplingStrategy,
+)
+from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.shuffling import _shuffle_list_and_tensor
+
+MatrixContent = Enum("MatrixContent", ["EMBEDDINGS", "GRADIENTS"])
+
+
+class AbstractMatrixDownsamplingStrategy(AbstractPerLabelRemoteDownsamplingStrategy):
```
francescodeaglio, line 15: This class is used to abstract the common behaviour of Kcentergreedy, Submodular and GradMatch

modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/README.md
```diff
@@ -0,0 +1,4 @@
+# DeepCore Utils
+The content of this folder is taken from DeepCore with minor adjustments.
```
francescodeaglio, line 2: Neither conda nor pip is available; you can only clone the repo. Some functions (like OMP) are changed.

modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/orthogonal_matching_pursuit.py
```diff
@@ -0,0 +1,105 @@
...
+            indices.append(index.item())
+            boolean_mask[index] = False
+
+            if indices.__len__() == 1:
+                A_i = A[:, index]
+                x_i = projections[index] / torch.dot(A_i, A_i).view(-1)
+                A_i = A[:, index].view(1, -1)
+            else:
+                A_i = torch.cat((A_i, A[:, index].view(1, -1)), dim=0)
+                temp = torch.matmul(A_i, torch.transpose(A_i, 0, 1)) + lam * torch.eye(A_i.shape[0], device="cuda:0")
```
francescodeaglio, line 49: Adapted DeepCore's version since lstsq has been updated in Pytorch. See [here](https://github.com/PatrickZH/DeepCore/issues/10)

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -0,0 +1,122 @@
...
+        self.matrix_elements = []
+        self.index_sampleid_map: list[int] = []
+
+    def select_points(self) -> tuple[list[int], torch.Tensor]:
+        if self.balance:
+            ids, weights = self.already_selected_samples, self.already_selected_weights
+        else:
+            ids, weights = self._select_from_matrix()
+
+        return _shuffle_list_and_tensor(ids, weights)
```
MaxiBoether, line 105: Could you briefly explain why we shuffle the selection again?

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
```diff
@@ -55,12 +58,22 @@ def __init__(self, pipeline_id: int, trigger_id: int, batch_size: int, params_fr
         # can use the following parameter
         self.requires_data_label_by_label = False
 
+        # Some methods require extra features (embedding recorder, get_last_layer) that are implemented in the class
+        # CoresetSupporingModule
```
MaxiBoether, line 62:
```suggestion
        # Some methods require extra features (embedding recorder, get_last_layer) that are implemented in the class
        # CoresetSupporingModule for model implementations.
```

modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/README.md
MaxiBoether: Could you maybe in the README file explain what you changed? (such that if DeepCore ever gets updated we know how to integrate the updates)

modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/README.md
```diff
@@ -0,0 +1,4 @@
...
+You can find the original code [here](https://github.com/PatrickZH/DeepCore/tree/main/deepcore/methods/methods_utils) and 
+the MIT license [here](https://raw.githubusercontent.com/PatrickZH/DeepCore/main/LICENSE.md)
```
MaxiBoether, line 4: Similarily to the other time, we need to inline the license here (literally, add a new subheading and copy + paste the content of the LICENSE file here)

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
```diff
@@ -0,0 +1,196 @@
...
+from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.shuffling import _shuffle_list_and_tensor
+from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.submodular_function import (
+    FacilityLocation,
+)
+from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.submodular_optimizer import (
+    OPTIMIZER_CHOICES,
+)
+
+
+class RemoteCraigDownsamplingStrategy(AbstractPerLabelRemoteDownsamplingStrategy):
```
MaxiBoether, line 20: Just for my understanding: Why do some downsamplers have their algorithm living in deepcore_utils (like kcenter_greedy), and others (like craig) have their algorithm here?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
```diff
@@ -0,0 +1,196 @@
...
+        target: torch.Tensor,
+        embedding: Optional[torch.Tensor] = None,
+    ) -> None:
+        assert embedding is not None
+
+        # Slightly different implementation for BTS and STB since in STB points are supplied class by class while in
+        # BTS are not. STB will always use the first branch, STB will typically (might use the first if all the points
+        # belong to the same class) use the second one
+        different_targets_in_this_batch = target.unique()
+        if len(different_targets_in_this_batch) == 1:
```
MaxiBoether, line 82: Is this your proxy for STB vs BTS? Because what if we only have samples of one class in BTS?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
```diff
@@ -0,0 +1,196 @@
...
+        self, sample_ids: list[int], forward_output: torch.Tensor, target: torch.Tensor, embedding: torch.Tensor
+    ) -> None:
+        embedding_dim = embedding.shape[1]
+        num_classes = forward_output.shape[1]
+        batch_num = target.shape[0]
+
+        loss = self.criterion(forward_output, target).mean()
+
+        # compute the gradient for each element provided
+        with torch.no_grad():
```
MaxiBoether, line 103: I may be stupid, but how can we compute gradients with torch.no_grad?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
```diff
@@ -0,0 +1,196 @@
...
+        self,
+        sample_ids: list[int],
+        forward_output: torch.Tensor,
+        target: torch.Tensor,
+        embedding: Optional[torch.Tensor] = None,
+    ) -> None:
+        assert embedding is not None
+
+        # Slightly different implementation for BTS and STB since in STB points are supplied class by class while in
+        # BTS are not. STB will always use the first branch, STB will typically (might use the first if all the points
```
MaxiBoether, line 79: Not sure what you mean by branch here and also, to my shame, i don't understand the problem here yet. Maybe you can elaborate a bit?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
```diff
@@ -16,9 +17,15 @@ class RemoteGradNormDownsampling(AbstractRemoteDownsamplingStrategy):
     """
 
     def __init__(
-        self, pipeline_id: int, trigger_id: int, batch_size: int, params_from_selector: dict, per_sample_loss: Any
+        self,
```
MaxiBoether, line 20: We could also add a reference here to the paper proposing it

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
```diff
@@ -31,30 +38,38 @@ def get_scores(self, forward_output: torch.Tensor, target: torch.Tensor) -> torc
...
+                # pylint: disable=E1102
+                one_hot_targets = F.one_hot(target, num_classes=num_classes)
                 scores = torch.norm(probs - one_hot_targets, dim=-1)
         else:
             sample_losses = self.per_sample_loss_fct(forward_output, target)
             last_layer_gradients = torch.autograd.grad(sample_losses.sum(), forward_output, retain_graph=False)[0]
             scores = torch.norm(last_layer_gradients, dim=-1)
 
-        return scores
+        return scores.cpu()
```
MaxiBoether, line 51: i am not sure I understand yet the score device handling. How exactly did you change that? In this PR, you change the device in the trainer, and here you add the .cpu() call, but I am not sure why

MaxiBoether (changes requested):
Thank you! This is very solid work. I had a few comments but hopefully they should not be too much

modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/README.md
francescodeaglio: Ok done

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
francescodeaglio: CRAIG's selection algorithm is in deepcore utils ( ``` from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.submodular_function import FacilityLocation ``` ). All the selection algorithms are in a separate file, except for the trivial ones (like top-k or weighted sampling). This follows the design of Deepcore and makes sense since the same algorithm can be used for different strategies (ex, submodularFL is used by Craig and Submodular)

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -606,14 +608,29 @@ def downsample_batch(
...
+            embeddings = None
+
+        self._downsampler.inform_samples(sample_ids, big_batch_output, target, embeddings)
         # TODO(#218) Persist information on the sample IDs/weights when downsampling is performed
         selected_indexes, weights = self._downsampler.select_points()
         selected_data, selected_target = get_tensors_subset(selected_indexes, data, target, sample_ids)
         sample_ids, data, target = selected_indexes, selected_data, selected_target
         # TODO(#219) Investigate if we can avoid 2 forward passes
-        return data, sample_ids, target, weights
+        return data, sample_ids, target, weights.to(self._device)
```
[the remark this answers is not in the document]
francescodeaglio: Line 390 is for STB, where we retrieve the samples from the disk. This function is for BTS, where weights are in memory. Hence, we need both `.to` since they operate on different tensors.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -714,7 +743,15 @@ def _iterate_dataloader_and_compute_scores(
...
                 # compute the scores and accumulate them
                 model_output = self._model.model(data)
-                self._downsampler.inform_samples(sample_ids, model_output, target)
+                # supply the embeddings if required by the downsampler
+                if self._downsampler.requires_coreset_supporting_module:
+                    assert isinstance(self._model.model, CoresetSupportingModule)
+                    assert self._model.model.embedding_recorder.record_embedding
+                    embeddings = self._model.model.embedding_recorder.embedding
+                else:
+                    embeddings = None
```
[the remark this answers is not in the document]
francescodeaglio: Added start, get, and end functions for embedding recording.

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
francescodeaglio: Training is done class by class if we do not shuffle with the balance version (first $N_0$ points of class 0, then $N_1$ points of class 1 and so on). In non-balance, it is less important, but I think it is still useful (otherwise, the training follows the order of importance of the selection, i.e. the first batches have all the most important points and then progressively less and less important points come in).

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
francescodeaglio: Added

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
francescodeaglio: `Torch.no_grad()` is applied to the operations within the context manager (matrix multiplication) and not to the previous ones (forward pass). Hence, we can still compute the gradient we need, and we avoid computing the gradient of the product (kind of gradient of the gradient), which would be useless.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
francescodeaglio: There's a reference for each downsampler. It is a few lines above for this one. 
```
"""
    Method adapted from
    Not All Samples Are Created Equal: Deep Learning with Importance Sampling (Katharopoulos, Fleuret)
    The main idea is that the norm of the gradient up to the penultimate layer can be used to measure the importance
    of each sample. This is particularly convenient if the loss function is CrossEntropy since a closed form solution
    exists and thus no derivatives are needed (so it's marginally more expensive than computing the loss)
"""
```

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
francescodeaglio: If the device is the GPU, the scores are computed in GPU (the tensors are there, and pytorch automatically performs the operations in GPU). However, to select the samples, we have to keep all the scores, and keeping them all in GPU does not make sense. Therefore, the scores are calculated in GPU, moved to CPU memory and saved in a list.
If the device is the CPU, everything happens already in the CPU, and that line is useless.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
francescodeaglio: The problem with CRAIG is that the algorithm considers class when selecting points. This is not a problem in STB since there is a dedicated dataloader to receive the samples as desired.
In BTS it is different because potentially in the same batch, there could be samples from different classes. If this happens, the algorithm is used in a different way (receiving samples class by class, emulating what happens in STB).

It is not a proxy for STB vs. BTS because there is no problem if there are samples from a single class in BTS. It is just to handle the special case of multiple classes in the same batch.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
francescodeaglio: Branch refers to the if/else. See the comment above for the explanation.

modyn/common/trigger_sample/trigger_sample_storage.py
francescodeaglio: Sure

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
MaxiBoether: Thank you!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
MaxiBoether: Ahh, could you add this as a comment here as well? Keeping this knowledge for the future. Thank you!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
MaxiBoether: sorry, I was braindead here apparently

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
MaxiBoether: ahh now I understand this comment. maybe you can extend this a bit with the comment you wrote above. I think there is also a typo here because you mention STB twice in the second sentence

MaxiBoether (approved):
Thank you. The last open thing is just migrating the one comment for Craig into the file, everything else looks good!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
francescodeaglio: Yes, you're right. Fixed

PR #305, opened 2023-10-07
Enhancements for wild time benchmarks

francescodeaglio:
This PR introduces three enhancements for the WildTime benchmarks. Here they are:

1. So far, all datasets except fMoW save data using only the year as the timestamp (finer information is unavailable). Since years can be pre-1970, these are mapped to days from 1/1/1970 (e.g. in YB 1930->1/1/1970, 1931->2/1/1970). fMoW, on the other hand, has the exact timestamps available and, before this PR, was used to save files. Now, to ensure consistency, fMoW also considers only the year. To use the entire timestamp, you can use the option `--daily`
2. In addition to the training dataset, wildtime provides a validation and test dataset. Since we do not have these concepts in modyn, all datasets can be used for training by specifying the `--all` option.
3. In Modyn, training is initiated by a sample that satisfies the triggering condition. Consequently, samples from the last year are never used since no sample from the following year initiates its training. By specifying the `--dummyyear` parameter, it is possible to add a sample to the year following the last one to enable training on the last year's data. For example, arXiv has data up to 2022. As a result, since there is no 2023 point, training on the 2022 data never happens. With this addition, a 2023 sample is added so that the 2022 data can be used.

### DIFF
--- a/.github/actions/mamba/action.yml
+++ b/.github/actions/mamba/action.yml
@@ -6,6 +6,10 @@ runs:
         with:
           python-version: '3.10'
 
+      - name: Install prerequisites for pystack
+        run: sudo apt-get install --no-install-recommends -qy libdw-dev libelf-dev
+        shell: bash
+
       - name: Set cache date
         run: echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
         shell: bash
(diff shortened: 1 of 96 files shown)


MaxiBoether (approved):
Thanks for the changes. Please only merge after testing the fMoW changes as you mentioned on Element. Thanks!

PR #316, opened 2023-10-17
Add CMake to project, add C++ CI checks, allow for C++ extensions in Python

MaxiBoether:
This PR is the basic setup for C++ infrastructure within Modyn. We add various CI stages using gcc and clang, including sanitizers, to test the setup. We provide an example extension to demonstrate the setup including unit tests.

See also
https://medium.com/@iftimiealexandru/bridging-the-gap-between-python-and-c-c-libraries-an-exploration-of-integrating-with-cython-cc3bbfcfe539

https://martinopilia.com/posts/2018/09/15/building-python-extension.html


modyn/common/cpp/src/utils/utils.cpp
vGsteiger, line 1: Willst du hier nicht gerade das `FAIL` und `ASSERT` einfügen? Dann haben wir diese direkt

MaxiBoether: das ist eigentlich schon drin im header :D

Issue #317, opened 2023-10-19
Make the supervisor a server and create a separate modyn_client


jenny011:
Currently, we instantiate a supervisor per pipeline.\
We want to make an architectural change to the supervisor:
1. Make the supervisor a grpc server that registers, executes and monitors pipelines upon StartPipeline requests from the client. First, we need to finish #302. 
2. Create a separate modyn_client that can submit pipelines to the supervisor and receive execution results.

MaxiBoether:
An open question we have to resolve is how to do the visualization. Currently, we use enlighten at the legacy supervisor. To achieve the same result with the Modyn CLI and Supervsior server, we would need to send status progress messages frequently to the CLI client (via gRPC), and use enlighten on the CLI. Maybe there also is an alternative

robinholzi:
Seems to be implemented, feel free to reopen if needed

Issue #251, opened 2023-05-10
Remove fragments of initial pass


MaxiBoether:
The initial pass proposed in #128 is not considered for now. We should remove the code fragments related to it from the supervisor.

MaxiBoether:
@jenny011: When changing to the supervisor server, we can get rid of the initial pass.

Issue #124, opened 2023-02-01
Pipeline deregistration


MaxiBoether:
After #111, we don't offer any gRPC functionality to deregister a pipeline at the selector. We should offer that function and think about what we do with existing data in the database.

MaxiBoether:
Can be addressed as part of creating the supervisor server that @jenny011  is working on.

Issue #107, opened 2023-01-20
Rethink command line flags for supervisor


MaxiBoether:
I see - I think it's a minor problem and can certainly be left this way for now. However, since your concern is decoupling the experiment from the pipeline, one thing I might bring up is perhaps we could include them in the pipeline config under `experiment_config`, and then if you want to run it in experiment mode, you can add an --experiment flag, which if absent, makes the system ignore the `experiment_configs`. Anyway as before it's a minor point, I just wanted to bring it up. Feel free to resolve!

_Originally posted by @K-Shao in https://github.com/eth-easl/dynamic_datasets_dsl/pull/79#discussion_r1082736114_

When we switch the supervisor to a continously running service to which we submit pipelines to, we will probably adjust that.

MaxiBoether:
Since @jenny011 is working on the modyn client / supervisor server, this will be handled as part of implementing that.

PR #322, opened 2023-10-31
Implement a Supervisor Server gRPC component

jenny011:
This PR implements the supervisor server part of #317. It also addresses #107 and #251.
1. The supervisor becomes a server running in its own container. It entrypoint takes only modyn_config as cmd line arg.
2. Validate configs and register pipeline in the supervisor.
3. Wrap pipeline execution into a `PipelineExecutor` class.
4. Init GRPCHandler connection to cluster components outside of GRPCHandler init due to ForkingPickler error.
5. Remove progress_mgr and status_bar for now because they are not picklable.
6. Run a pipeline monitor thread in the supervisor to periodically polls for pipeline termination.
7. Use a system-wide seed for randomness.
8. Implement a dummy model (one Identity layer) to replace resnet in supervisor integration tests.

TODOs (in separate PRs)
1. Implement the modyn client part of #317. The client can connect to the supervisor and send StartPipelineRequest.
2. Implement supervisor get_pipeline_status proto.
3. Implement client progress bar.
4. Implement pipeline deregistration #124.
5. Implement supervisor integration test. 1.a. #124 so that we can cleanup pipelines after testing. 1.b. Check that pipeline execution process terminated without error.
6. Use per-pipeline seeds if possible.


modyn/protos/supervisor.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package supervisor;
+
+service Supervisor {
+  rpc start_pipeline(StartPipelineRequest) returns (PipelineResponse) {}
+}
+
+message Empty {}
```
MaxiBoether, line 9: Do we need this empty message?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -66,20 +66,30 @@
 class GRPCHandler:
     # pylint: disable=too-many-instance-attributes
 
+    # pylint: disable=unused-argument
     def __init__(
         self,
         modyn_config: dict,
         progress_mgr: Optional[enlighten.Manager] = None,
         status_bar: Optional[enlighten.StatusBar] = None,
```
MaxiBoether, line 74: Do you plan to remove this within a follow PR? I am fine with either (removing it now or in a follow up PR)

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+
+def pipeline_monitor(pipeline_process_dict: dict[int, PipelineInfo]) -> None:
+    logger.info(f"[{os.getpid()}][pipeline_monitor] start")
+    while True:
+        num_pipelines = len(pipeline_process_dict)
+        if num_pipelines > 0:
+            logger.info(f"[{os.getpid()}][pipeline_monitor] {num_pipelines} pipelines registered")
+
+        # join all the terminated procs
+        num_active_pipeline_processes = len(mp.active_children())
```
MaxiBoether, line 34: I am a bit unsure about the use of mp.active_children here. Maybe it does exactly what it should, but for example gRPC may spawn threads etc - does this truly only list the threads/processes we spawned using multiprocessing? Maybe we should use something like `len([p_id for  for p_id, p_info in pipeline_process_dict.items() if p_info.process_handler.is_alive()])`? Not sure...

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+        if num_active_pipeline_processes > 0:
+            logger.info(f"[{os.getpid()}][pipeline_monitor] {num_active_pipeline_processes} pipeline processes running")
+
+        for p_id, p_info in pipeline_process_dict.items():
+            p_info.update_status()
+            if p_info.pipeline_exited:
+                proc_alive = p_info.process_handler.is_alive()
```
MaxiBoether, line 41: I haven't read the source of `PipelineInfo` yet, but can we have an exited pipeline where the process is still running?

tmuxp.yaml
```diff
@@ -3,6 +3,7 @@ windows:
   - window_name: supervisor
     layout: tiled
     panes:
+      - docker logs $(docker compose ps -q supervisor) && docker attach $(docker compose ps -q supervisor)
       - shell_command:
         - docker exec -it $(docker compose ps -q supervisor) /bin/bash
         - clear && mamba activate modyn && echo "Welcome to the supervisor pane. You have access to the 'modyn-supervisor' command here to submit pipelines."
```
MaxiBoether, line 9: I think we might comment this out OR already prepare a pane for the client. This pane will be what the client needs, right now the supervisor server does not need that anymore

scripts/run_integrationtests.sh
```diff
@@ -22,6 +22,8 @@ docker build -t modynbase -f docker/Base/Dockerfile --build-arg MODYN_BUILDTYPE=
 docker compose up --build tests --abort-on-container-exit --exit-code-from tests
 exitcode=$?
 
+docker logs supervisor
```
MaxiBoether, line 25: Leftover debugging?

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+        return state
+
+    def init_metadata_db(self) -> None:
+        with MetadataDatabaseConnection(self.modyn_config) as database:
+            database.create_tables()
+
+    def init_cluster_connection(self) -> None:
+        logging.info("Setting up connections to cluster components.")
+        self.grpc.init_cluster_connection()
+        self.init_metadata_db()
```
MaxiBoether, line 102: We might be able to init the metadata db in the constructor, since we only need to do that once

modyn/supervisor/internal/supervisor.py
MaxiBoether: (My main "concern" is that we have multiple supervisors due to multiprocessing of the gRPC server and I am not sure how much they interfere with each other here)

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+
+        exception_queue: mp.Queue[str] = mp.Queue()  # pylint: disable=unsubscriptable-object
+        status_query_queue: mp.Queue[str] = mp.Queue()  # pylint: disable=unsubscriptable-object
+        status_response_queue: mp.Queue[dict[str, Any]] = mp.Queue()  # pylint: disable=unsubscriptable-object
+
+        start_timestamp = self.grpc.get_time_at_storage()
+        pipeline_id = self.register_pipeline(pipeline_config)
+        logger.info(f"Pipeline {pipeline_id} registered, start executing.")
+
+        logger.info(f"start_pipeline: start method is {mp.get_start_method()}")
```
MaxiBoether, line 322: Debug leftovers?

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+        # join all the terminated procs
+        num_active_pipeline_processes = len(mp.active_children())
+        if num_active_pipeline_processes > 0:
+            logger.info(f"[{os.getpid()}][pipeline_monitor] {num_active_pipeline_processes} pipeline processes running")
+
+        for p_id, p_info in pipeline_process_dict.items():
+            p_info.update_status()
+            if p_info.pipeline_exited:
+                proc_alive = p_info.process_handler.is_alive()
+                logger.info(
```
MaxiBoether, line 42: Do we also want to check the exception queue here or is this something we only check when the client asks for a status update?

modyn/supervisor/internal/supervisor.py
```diff
@@ -0,0 +1,368 @@
...
+                         exit code {p_info.pipeline_exit_code.name},
+                         proc alive {proc_alive}"""
+                )
+                if not proc_alive:
+                    # TODO(#317): unregister pipeline when process terminates
+                    pass
+            else:
+                logger.info(f"[{os.getpid()}][pipeline_monitor] pipeline {p_id} still running")
+
+        time.sleep(10)
```
MaxiBoether, line 53: Let's make this a global in this class, i.e., `MONITORING_INTERVAL = 10` or so

modyn/supervisor/internal/supervisor.py
MaxiBoether: Ahhh, I see. The PipelineInfo checks the for the exception queue? My only thought is that, similar to the trainer server, we might want to forward the exception to the client when the client asks for the pipeline status update, and I am not 100% sure how we will implement that if we already consume the exception, but maybe that's something we can figure out in the client PR

modyn/supervisor/internal/utils/pipeline_info.py
```diff
@@ -0,0 +1,55 @@
...
+        if not self.pipeline_exited:
+            self.exception_msg = self.check_for_exception()
+            # Mark pipeline as exited when exception occurs
+            # But, process is not dead right after exception, it is alive until joined with the parent process
+            if self.exception_msg is not None:
+                self.pipeline_exited = True
+                self.pipeline_exit_code = PipelineExitCode.ERROR
+            # Mark pipeline as exited when process is joined w.o. exception
+            elif not self.process_handler.is_alive():
+                self.pipeline_exited = True
```
MaxiBoether, line 54: We might want to check the exit code of the proecss handler (exitcode should be the name of the parameter here). It could potentially be that for soem reason we do not get an exception but have a non zero exit code. In that case, we still should set `self.pipeline_exit_code = PipelineExitCode.ERROR`

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -0,0 +1,390 @@
+import json
+import logging
+import multiprocessing as mp
+import os
+import pathlib
+import traceback
+from time import sleep
+from typing import Any, Optional
+
+# import enlighten
```
MaxiBoether, line 10: I don't think we'll need enlighten here, it will only be at the client

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -0,0 +1,94 @@
...
+
+if __name__ == "__main__":
+    dataset_helper = DatasetHelper()
+    try:
+        dataset_helper.setup_dataset()
+        test_start_one_experiment_pipeline()
+        test_start_two_experiment_pipelines()
+        time.sleep(TIMEOUT)
+        # test_mnist()
+        # time.sleep(TIMEOUT * 5)
```
MaxiBoether, line 90: leftovers?

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -0,0 +1,94 @@
...
+    print(f"pipeline ids: {pipeline1_id}, {pipeline2_id}")
+    assert pipeline1_id >= 0 and pipeline2_id >= 0
+
+
+if __name__ == "__main__":
+    dataset_helper = DatasetHelper()
+    try:
+        dataset_helper.setup_dataset()
+        test_start_one_experiment_pipeline()
+        test_start_two_experiment_pipelines()
```
MaxiBoether, line 87: In the current implementation, we don't have any way to wait for training to finish, right? I think we should still somehow make training here non stressful on CI resources. Could we maybe implement a model that just does "nothing" on the forward pass, (e.g., outputs a constant or so), and change the example pipeline we test to use that model instead of a resnet? because otherwise we acutally train a resnet on github CI servers, I would like to avoid that, if possible. Thank you!

MaxiBoether (changes requested):
Thank you very much for the work - this again is a very clean PR! A pleasure to review :) I had a few minor comments, after this round we can probably merge this :)

integrationtests/supervisor/integrationtest_supervisor.py
jenny011: It was used to test mnist. We could also run mnist instead of the random dataset. What would you prefer?

scripts/run_integrationtests.sh
jenny011: Yes, I plan to remove it as the last step before merging this PR. It helps me understand the github checks results.

modyn/supervisor/internal/supervisor.py
jenny011: The process is alive if its parent does not call join on it, even if it has exited due to an exception. That's also why I call `mp.active_children()` to periodically join all the exited processes (related to comment https://github.com/eth-easl/modyn/pull/322#discussion_r1390474318).

modyn/supervisor/internal/supervisor.py
jenny011: I'll double check

tmuxp.yaml
jenny011: For this PR, I can make the supervisor window the same as the other servers.

modyn/supervisor/internal/supervisor.py
jenny011: Yes, will remove it when the PR is ready to merge.

modyn/supervisor/internal/supervisor.py
jenny011: It is a temporary solution for me to check that pipelines run or break as expected in this PR. Yeah, we probably need a better design in the client PR.

modyn/protos/supervisor.proto
jenny011: done.

modyn/supervisor/internal/grpc_handler.py
jenny011: removed.

modyn/supervisor/internal/supervisor.py
jenny011: done.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
jenny011: removed.

modyn/supervisor/internal/utils/pipeline_info.py
jenny011: I removed PipelineExitCode for now and let pipeline_monitor check process.exitcode instead. I added sys.exit(custom_code) in the exception handling part of the execute_pipeline() runnable.
related discussions: https://github.com/eth-easl/modyn/pull/322#discussion_r1390490520, https://github.com/eth-easl/modyn/pull/322#discussion_r1390488907, https://github.com/eth-easl/modyn/pull/322#discussion_r1390489253

modyn/supervisor/internal/supervisor.py
jenny011: done

modyn/supervisor/internal/supervisor.py
jenny011: done

modyn/supervisor/internal/supervisor.py
jenny011: removed update_status().

tmuxp.yaml
jenny011: done

modyn/supervisor/internal/utils/pipeline_info.py
jenny011: done

integrationtests/supervisor/integrationtest_supervisor.py
jenny011: done

modyn/supervisor/internal/supervisor.py
jenny011: done

modyn/protos/supervisor.proto
MaxiBoether: Thanks!

modyn/supervisor/internal/grpc_handler.py
MaxiBoether: Thanks

modyn/supervisor/internal/supervisor.py
MaxiBoether: Thank you!

tmuxp.yaml
MaxiBoether: thx

modyn/supervisor/internal/supervisor.py
MaxiBoether: thx

integrationtests/supervisor/integrationtest_supervisor.py
jenny011: Will open a separate PR

modyn/supervisor/internal/supervisor.py
jenny011: done

scripts/run_integrationtests.sh
jenny011: done

integrationtests/supervisor/integrationtest_supervisor.py
jenny011: merged

MaxiBoether (approved):
Looks great, thank you! My only comment is that `check_for_exception` on PipelineInfo is currently never called, but I assume we will only use that when the client asks for the status. For now, it's all looking good!

Issue #262, opened 2023-05-23
Use Binary File Format to Store Simple Selector State


MaxiBoether:
The insertion process at the selector is a bottleneck. Postgres inserts with 80-100k rows/second, which means waiting for 30+ minutes for a day of Criteo data to be inserted.

For many simple strategies (retraining/finetuning/random subsampling/…), we actually do not need any complicated Postgres query. We can implement this on a simple binary file format. We can leverage a file structure such as `/tmp/<pipeline id>/<trigger id>/<files containing binary data>`

These files can be binary fixed size tables with at least the sample ID as a column. We need to build a nice abstraction around this, such that strategies can add other fields as well. For example, GDumb might require to fetch and store the labels of all data from the storage. With #261 in mind, this would be the place where we go from intervals to sample level information.

For other strategies that require more complex queries or state across components (e.g., storing information from the training), we can still keep the `SelectorStateMetadata` table. For example, a strategy similar to RHO-LOSS might be a use case of such a table.

Superseeds #204. Will probably include #250.

MaxiBoether:
This is implementing the LocalStorageBackend, which @Sipondo started to work on - thanks!

MaxiBoether:
Solved by #364!

PR #364, opened 2024-01-15
Local Storage Backend

Sipondo:
#262


modyn/selector/internal/storage_backend/local/local_storage_backend.py
```diff
@@ -1,27 +1,127 @@
 import logging
-from typing import Any, Iterable, Optional
+from pathlib import Path
+from typing import Iterable, Optional
 
+import numpy as np
+from modyn.common.local_storage_backend import LocalStorageBackend as LocalStorageBackendCPP
```
MaxiBoether, line 6: Is there a reason we have the LocalStorageBackend in common and then import it here? I am considering either having the CPP wrapper in the storage component or alternatively, both within the same class here. I think it's not a common utility but rather specific for the storage (unlike the triggersamplestorage, where this historaically probably stems from). I'd be happy if you could adjust it in one of those two ways (and move the tests appropriately)

modyn/selector/internal/storage_backend/local/local_storage_backend.py
```diff
@@ -1,27 +1,127 @@
...
 from modyn.selector.internal.storage_backend import AbstractStorageBackend
 
 logger = logging.getLogger(__name__)
 
+MAX_SAMPLES_IN_FILE = 1000000
```
MaxiBoether, line 11: should we make this a modyn system parameter?

modyn/tests/selector/internal/storage_backend/local/test_local_storage_backend.py
```diff
@@ -0,0 +1,142 @@
+import shutil
```
MaxiBoether, line 1: Thanks for the tests here - we need to slightly adjust this depending on how we decide to change the structure: If we have the CPP Python wrapper still in a separate file, let's test that individually and then the LocalStorageBackend, which uses this wrapper, separately.

Additionally, similarily to the triggersamplestorage, it would be great to have a couple of C++ tests as well to have the sanitizer check this as well.

MaxiBoether (approved):
Thanks Ties! I am pre-approving this but have two comments:

1) the structure of the CPP wrapper (it should go into the storage component instead of common I think, since it's not a common utility, unlike the TSS). 

2) a couple of C++ tests similar to the TSS would be amazing

Thank you so much. After adjusting this feel free to merge.

modyn/common/local_storage_backend/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+This module contains classes shared among several components.
```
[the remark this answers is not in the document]
Sipondo: Done

modyn/selector/internal/storage_backend/local/local_storage_backend.py
Sipondo: yeah we should - where should I add it?

Sipondo: done

modyn/selector/internal/storage_backend/local/local_storage_backend.py
Sipondo: done

modyn/tests/selector/internal/storage_backend/local/test_local_storage_backend.py
Sipondo: Done

MaxiBoether (approved):
Thank you! Only the Dockerfile thing I think we need to change, everything else LGTM! Thanks!

docker/Storage/Dockerfile
```diff
@@ -10,6 +10,11 @@ COPY ./modyn/common/example_extension /src/modyn/common/example_extension
 COPY ./modyn/common/trigger_sample /src/modyn/common/trigger_sample
 COPY ./modyn/protos/storage.proto /src/modyn/protos/storage.proto
 
+COPY ./modyn/selector/internal/storage_backend/local /src/modyn/selector/internal/storage_backend/local
+COPY ./modyn/selector/CMakeLists.txt /src/modyn/selector/CMakeLists.txt
+COPY ./modyn/selector/internal/CMakeLists.txt /src/modyn/selector/internal/CMakeLists.txt
+COPY ./modyn/selector/internal/storage_backend/CMakeLists.txt /src/modyn/selector/internal/storage_backend/CMakeLists.txt
```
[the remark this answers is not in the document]
Sipondo: done

PR #370, opened 2024-03-07
ignore jetbrains files in .gitignore

XianzheMa:
I added an entry in `.gitignore` to ignore JetBrains related files.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -95,8 +95,6 @@ jobs:
   unittests:
     timeout-minutes: 25
     runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
     steps:
       - name: Check out code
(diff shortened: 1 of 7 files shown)


MaxiBoether (approved):
Thanks! Just need to wait for CI to finish

.github/workflows/workflow.yaml
```diff
@@ -95,6 +95,8 @@ jobs:
   unittests:
     timeout-minutes: 25
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
XianzheMa, line 99: so that the commentator can comment in the PR
Not sure why it passed previously; possibly the default permissions of `GITHUB_TOKEN` has changed?

XianzheMa:
changes incorporated in https://github.com/eth-easl/modyn/pull/372.

PR #372, opened 2024-03-11
Fix failing CI, supplement gitignore and add dockerignore

XianzheMa:


modyn/tests/evaluator/internal/grpc/test_evaluator_grpc_servicer.py
```diff
@@ -173,6 +174,18 @@ def test_init(test_connect_to_model_storage, test_connect_to_selector, test_conn
...
+def is_writable(path):
+    try:
+        testfile = tempfile.TemporaryFile(dir=path)
+        testfile.close()
+    except OSError as e:
+        if e.errno == errno.EACCES:  # 13
+            return False
+        e.filename = path
+        raise
+    return True
```
MaxiBoether, line 186: that was just WIP can can go :)

modyn/tests/evaluator/internal/grpc/test_evaluator_grpc_servicer.py
```diff
@@ -5,6 +5,7 @@
 import pathlib
 import platform
 import tempfile
+import errno
```
MaxiBoether, line 8: was old stuff

.github/workflows/workflow.yaml
```diff
@@ -95,7 +95,6 @@ jobs:
   unittests:
     timeout-minutes: 25
     runs-on: ubuntu-latest
-
```
MaxiBoether, line 98: maybe recover that line :D

MaxiBoether (approved):
thanks for consolidating all of this - let's hope it runs through now...

modyn/tests/supervisor/internal/test_supervisor.py
```diff
@@ -438,9 +438,18 @@ def test_get_pipeline_status_exit(
...
+
+    # iteratively query the pipeline status until it is no longer running
+    while time.time() - start_time < 30:
+        msg = sup.get_pipeline_status(res["pipeline_id"])
+        if msg["status"] != PipelineStatus.RUNNING:
+            break
+        time.sleep(2)
+    else:
+        raise TimeoutError("Pipeline did not finish in time")
+
```
XianzheMa, line 452: On my local computer, the original test fails because sleeping 1 second is not enough for the pipeline to finish (it is still running after 1s). I made an iterative loop to only check the pipeline status when it finishes running.

.dockerignore
```diff
@@ -0,0 +1,14 @@
+libbuild/
```
XianzheMa, line 1: Without this file, some caches built on local computer will be included in the image build environment, and then the docker build will fail because of cache inconsistency.

modyn/tests/supervisor/internal/test_supervisor.py
MaxiBoether: Thanks for cleaning all of this up, Xianzhe. Much appreciated!

.dockerignore
MaxiBoether: Yeah, we were really missing a dockerignore file. Thank you so much for creating one

PR #373, opened 2024-03-12
reduce files copied into image

XianzheMa:
Per the title, I reduced files copied into the docker image.

Currently almost all files in the repository are copied into the image. This has the downside that whenever I changed a irrelevant file (a file not used to build the image environment), docker invalidates previous image caches and has to build the image from scratch, which takes 1-2 unnecessary minutes. Now this time is saved.

Also a port mistake is corrected

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,7 +142,7 @@ services:
 #      - .:/modyn_host
     container_name: supervisor
     ports:
-      - 3000:50062
+      - "3000:50063"
   tests:
     depends_on:
       supervisor:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -142,7 +142,7 @@ services:
 #      - .:/modyn_host
     container_name: supervisor
     ports:
-      - 3000:50062
+      - "3000:50063"
```
XianzheMa, line 145: 50062 is never the port for supervisor... I was not sure how other people correctly ran Modyn before

PR #374, opened 2024-03-14
Fix broken args eval_dir on modyn-client

XianzheMa:
## What is the current case
Currently, the `eval_dir` arg on modyn-client is practically unusable: it is a required arg; the client throws an exception if this path does not exist on host system; supervisor throws an exception if this same path does not exist on supervisor's container. Therefore user can practically set nothing to `eval_dir` instead `.` (a reference to the current working directory, possibly the only value that can pass both checks)

Then users need to use `docker cp <more args>` to manually copy evaluation files from supervisor container to host.

## What it would be after this PR

Supervisor has a default location, `eval_directory` to store evaluation results (on supervisor container), whose value is in `modyn_config.yaml`. It is up to the users to decide how to map this path to host filesystem, e.g. via volume mapping or named volume in `docker-compose.yaml`. The `eval_dir` arg on modyn-client is removed.


modyn/config/schema/modyn_config_schema.yaml
```diff
@@ -376,6 +376,25 @@ properties:
...
+          The port to use for the supervisor.
+      eval_base_path:
+        type: string
+        description: |
+          The base path to store the evaluation results.
+    required:
+      - hostname
+      - port
+      - eval_base_path
 required:
```
XianzheMa, line 398: is `metadata_processor` not required? please confirm

MaxiBoether: metadata_processor currently is not used at all, indeed. that's a relict from old days, currently not needed.

MaxiBoether:
Pinging @jenny011 here - see my message on Element about this :-)

jenny011 (approved):
Looks good to me!

MaxiBoether (changes requested):
Thanks, Xianzhe! However, I would like to go with the FTP solution that we initially thought of (I hope what I mean by that is clear from my message in element - if not, let me know!).

The reason is that the modyn-client might not live on the same machine as the supervisor server. Right now, that is very much likely the case, but the general idea is that you can have Modyn running, e.g., on a sgs-gpu machine, and start a pipeline using modyn-client on your local laptop. In this case, there is no way to access the mounted directory. This is why we should transfer the logfiles via FTP instead. If you could implement that instead, that would be very much appreciated. I guess you could also reach out to @jenny011, but I think the biggest issue was that we need to have a FTP passive port _range_ and active ports. Using https://pyftpdlib.readthedocs.io/en/latest/api.html#pyftpdlib.handlers.FTPHandler.passive_ports, we should be able to set that.

XianzheMa:
> Thanks, Xianzhe! However, I would like to go with the FTP solution that we initially thought of (I hope what I mean by that is clear from my message in element - if not, let me know!).
> 
> The reason is that the modyn-client might not live on the same machine as the supervisor server. Right now, that is very much likely the case, but the general idea is that you can have Modyn running, e.g., on a sgs-gpu machine, and start a pipeline using modyn-client on your local laptop. In this case, there is no way to access the mounted directory. This is why we should transfer the logfiles via FTP instead. If you could implement that instead, that would be very much appreciated. I guess you could also reach out to @jenny011, but I think the biggest issue was that we need to have a FTP passive port _range_ and active ports. Using https://pyftpdlib.readthedocs.io/en/latest/api.html#pyftpdlib.handlers.FTPHandler.passive_ports, we should be able to set that.

Thanks for explaining the vision! I agree with you on the direction that the modyn-client should be run on any machine apart from the modyn server, but I think enabling the FTP for now is an overkill, because:
1. Having a solution that works locally is better than leaving the current bug. And with this PR, it will make our experimentations later much more efficient. We can still implement FTP later after this PR, but solving the bug and making our experiment-workflow more convenient is more important for now. Let's build up the solution iteratively.
2. Without FTP and with the client on a different machine, the client can still download logs from the server by, for example, scp.

Therefore I am still inclined towards merging this PR now to make experiments easier. I can create an issue to discuss FTP/how we should implement the file transfer in a remote-server setting. What do you think?

docs/EXAMPLE.md
```diff
@@ -33,12 +33,15 @@ After building the containers, run `tmuxp load tmuxp.yaml` to have access to all
...
 ### Starting the pipeline
 You can now submit a pipeline to the supervisor container using the `modyn-client` command in `modynclient/client/`.\
-For example, you can run `modyn-client --start-replay-at 0 --maximum-triggers 1 <pipeline config file> <modyn client config file> .`\
+For example, you can run `modyn-client --start-replay-at 0 --maximum-triggers 1 <pipeline config file> <modyn client config file> <eval relative path>`\
 pipeline config file example: modynclient/config/examples/mnist.yaml\
 modyn client config file example:
 - on your local machine: modynclient/config/examples/modyn_client_config.yaml
 - in one of the containers: modynclient/config/examples/modyn_client_config_container.yaml
 
+eval relative path: specifies where the evaluation results are stored in supervisor's container. The absolute path will be
```
MaxiBoether, line 42: Please mention the issue of FTP here

modynclient/client/README.md
```diff
@@ -2,8 +2,10 @@
 
 This is the modynclient module 
 
-The client can connect to the modyn supervisor at the ip and port specified in the modyn_client_config file. ML engineers start a pipeline using `modyn-client pipeline.yaml config.yaml`″.
-The first configuration file describes the pipeline setup, while the second configuration file describes the modyn system setup.
+The client can connect to the modyn supervisor at the ip and port specified in the modyn_client_config file. ML engineers start a pipeline using `modyn-client pipeline.yaml config.yaml <eval_relative_path>`″.
+The first configuration file describes the pipeline setup, while the second configuration file describes the modyn system setup. `eval_relative_path` specifies where the evaluation results are stored in supervisor's container.
+The absolute path will be `modyn_config['supervisor']['eval_base_path']/eval_relative_path`.
```
MaxiBoether, line 7: please mention the ftp issue here

MaxiBoether (approved):
Okay, then please go ahead and create and mention the FTP issue in the adjusted docs, so we know this is just a temporary solution

PR #376, opened 2024-03-26
Default prefetch configs for evaluation

XianzheMa:
`num_prefetched_partitions` and `parallel_prefetch_requests` are two optional parameters in pipeline config, but required when `--evaluation-matrix` is specified. This PR makes them 1 (the default value according to the schema) when they are not specified but needed.

Additionally this PR adds a small test to increase test coverage.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -105,7 +105,7 @@ jobs:
 
       - name: Pytest
         run: |
-          micromamba run -n modyn pytest modyn --cov-reset --cache-clear --cov-fail-under=90
+          micromamba run -n modyn pytest modyn --cache-clear --cov-report term --cov-fail-under=90
           micromamba run -n modyn pytest > pytest-coverage.txt
 
       - name: Comment coverage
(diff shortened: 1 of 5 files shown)


.github/workflows/workflow.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
 
       - name: Pytest
         run: |
-          micromamba run -n modyn pytest modyn --cov-reset --cache-clear --cov-fail-under=90
+          micromamba run -n modyn pytest modyn --cache-clear --cov-report term --cov-fail-under=90
```
XianzheMa, line 108: why do we need --cov-reset? This is the first time we use pytest and no previous configurations are accumulated

MaxiBoether: This stems from times when we were using internal Github runners (non public) where the cache was not properly cleaned. I would suggest to just keep it, if it's not causing trouble.

MaxiBoether (approved):
LGTM, thanks Xianzhe - but I would just keep `cov-reset` since it resolved some issue some time ago when we were using internal runners. If you could change that, feel free to merge.

.github/workflows/workflow.yaml
XianzheMa: I need to investigate into `--cov-reset`. Theoretically it has nothing to do with cache 
https://github.com/pytest-dev/pytest-cov/issues/460

PR #377, opened 2024-03-27
Fix yearbook channel bug

XianzheMa:


benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -44,7 +43,7 @@ def _get_year_data(self, year: int, store_all_data: bool) -> list[Tuple]:
             np.array(
                 [   # transpose to transform from HWC to CHW (H=height, W=width, C=channels).
                     # Pytorch requires CHW format
-                    img.transpose(2, 0, 1)[split].reshape(*self.input_dim)
+                    img.transpose(2, 0, 1)
```
MaxiBoether, line 46: I understand the necessity to change `num_input_channels` to 3. Could you just briefly comment why we do _not_ filter by `split` anymore and not reshape, especially given the comment above? The comment mentions we need to go from HWC to CHW, do we still do that? Sorry, I did not write that code, but I am trying to validate the changes.

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
```diff
@@ -44,7 +44,7 @@ data:
     import torch
     import numpy as np
     def bytes_parser_function(data: bytes) -> torch.Tensor:
-      return torch.from_numpy(np.frombuffer(data, dtype=np.float32)).reshape(1, 32, 32)
```
MaxiBoether, line 47: similar question here on why this is necessary

benchmark/wildtime_benchmarks/data_generation_yearbook.py
XianzheMa: Sorry, I forgot explaining the motive behind. I believe this is even a bug.
Suppose `splits = [0, 1]`, this means we want to download both train and validation datasets.

img itself has shape (32, 32, 3): it has size 32 x 32 and 3 channels. The comment has right and transpose is needed. 

If we filter by `split`, the dataset will first have a huge chunk of pictures of only red scale (when split = 0), then a chunk of pictures of only green scale. And we train our model on red scale pictures but validate on green scale ones. This does not make sense to me.

MaxiBoether:
pinging @francescodeaglio here - if you see this and have some minutes and this interests, I am curious. Are we missing something here? It seems like the split / data channel shenanagans might be wrong, but maybe we also misunderstand something. Thank you so much already if you take a look at this!

benchmark/wildtime_benchmarks/data_generation_yearbook.py
XianzheMa: @francescodeaglio Another question, why do we concatenate the training set and validation set into a single dataset, instead of generating two independent ones?

MaxiBoether: @XianzheMa The training / validation thing was not really existing in Francesco's thesis (or done manually), IIRC. In Modyn itself, we mostly had the dataset for training, and francesco had his own evaluation scripts. For us, we can change the splitting logic

MaxiBoether:
just as a note here: if we don't hear back from @francescodeaglio here or via mail, let's merge this by end of this week :)

benchmark/wildtime_benchmarks/data_generation_yearbook.py
francescodeaglio: Hi Xianzhe and Maxi, I searched a bit because your version made sense to me. I think [split] is wrong and comes from a wrong generalisation, it should just be [0]. Consequently, the remaining code is taken from [here](https://github.com/wistuba/Wild-Time-Data/blob/35aa927dd292e09e6115ea463583f1cff6602c47/wild_time_data/datasets.py#L146)  . I don't remember more but if I recall correctly I evaluated end-to-end from storage to training loop and everything was working.

XianzheMa: I see. Thanks a lot for clarifying!

XianzheMa:
In case @MaxiBoether  is on the vacation, @jenny011 could you approve this? Maxi already approves this orally; I just realized this PR was previously a draft, so I need someone to approve it to merge

MaxiBoether (approved):

PR #379, opened 2024-03-28
Time-slice the EvaluationDataset

XianzheMa:
This is the first step towards evaluating on a separate dataset: to time-slice the `EvaluationDataset`.

Specifically, we add two parameters to the class `EvaluationDataset`: `start_timestamp`, `end_timestamp`,
to only select samples falling into this interval from the evaluation dataset.

We hence also add these two parameters to a storage grpc API `GetDataPerWorker`.

We also add these parameters to another API `GetDatasetSize` because it's used in evaluator


modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -186,6 +186,130 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
     return {StatusCode::OK, "Data retrieved."};
   }
 
+  template <typename WriterT>
+  Status GetDataPerWorker_Impl(  // NOLINT (readability-identifier-naming)
+      ServerContext* /*context*/, const modyn::storage::GetDataPerWorkerRequest* request, WriterT* writer) {
```
XianzheMa, line 191: copy-pasted from storage_service.cpp

modyn/tests/storage/storage_test_utils.hpp
```diff
@@ -34,6 +34,8 @@ class MockServerWriter : public grpc::ServerWriterInterface<T> {
 
   std::vector<T> get_responses() { return responses_; }
 
+  void clear_responses() { responses_.clear(); }
```
XianzheMa, line 37: so that I can reuse this writer across grpc requests

modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -186,6 +186,138 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
...
+              fmt::format("SELECT sample_id FROM samples WHERE dataset_id = {} ORDER BY sample_id LIMIT {} OFFSET {}",
+                          dataset_id, limit, start_index);
+        } else {
+          query = fmt::format(
+              "SELECT samples.sample_id "
+              "FROM samples INNER JOIN files "
+              "ON samples.file_id = files.file_id AND samples.dataset_id = files.dataset_id "
+              "WHERE samples.dataset_id = {} AND {} "
+              "ORDER BY sample_id LIMIT {} OFFSET {}",
+              dataset_id, timestamp_filter, limit, start_index);
```
XianzheMa, line 248: Since `GetDataPerWorker` API is only used in evaluation, which is off the critical path of training.
I consider this as okay

MaxiBoether: the `send_sample_id_and_label` implementation does the same so should be ok, however, I still wonder whether we can deduplicate the code somehow.

MaxiBoether (changes requested):
thanks Xianzhe. Sorry, this is not an in-depth review. My biggest comments are to double check whether we can reduce code duplication somehow and to extend the integration tests (besides addressing the memory leaks, of course :)). 

Thanks already!

modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
```diff
@@ -130,6 +131,7 @@ std::tuple<int64_t, int64_t> add_non_existing_sample(const YAML::Node& config, c
   if (!session.get_last_insert_id("samples", inserted_sample_id_ll)) {
     FAIL("Failed to insert sample into database");
   }
+  session.close();
```
MaxiBoether, line 134: One additional comment here: The tests ran through without leaks beforehand, so I hope the destructor of the session would work correctly... but I hope that this might fix the leak issues you encountered.

modyn/protos/storage.proto
```diff
@@ -63,6 +63,9 @@ message GetDataPerWorkerRequest {
   string dataset_id = 1;
   int32 worker_id = 2;
   int32 total_workers = 3;
+  // value unset or set with default value means no limit
+  int64 start_timestamp = 4;
```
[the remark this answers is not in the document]
XianzheMa: In `proto3` there is no "required" field. A field without `optional` prefix means if the field is unset, the default value is returned every time we read this `field`.

I don't have a specific reason: I just find it very handy to automatically have the default value when unspecified, rather than having to check if the field is present on the reader side.

More info https://protobuf.dev/programming-guides/field_presence/#c-example

modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -186,6 +186,138 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
     return {StatusCode::OK, "Data retrieved."};
   }
 
+  template <typename WriterT>
+  Status GetDataPerWorker_Impl(  // NOLINT (readability-identifier-naming)
```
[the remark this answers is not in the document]
XianzheMa: I probably used the wrong word :D
It is not `copy-pasted`, but `cut-pasted`, meaning the implementation is moved here.

This `GetDataPerWorker` API is not unit-tested before, and to unit-test it, I have to inject a mock writer (it is a `stream` grpc API). So as the other `stream` APIs like `GetDataSince`, `GetDataInInterval`, I have to move the implementation to a template to accept a mock writer.

modyn/protos/storage.proto
```diff
@@ -63,6 +63,9 @@ message GetDataPerWorkerRequest {
   string dataset_id = 1;
   int32 worker_id = 2;
```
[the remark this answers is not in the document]
XianzheMa: Of course!

modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
XianzheMa: Yes it fixed and I think maybe the leak is only marked as error when the amount is over a threshold?

But the `session` **is** related to the memory leak. see https://github.com/eth-easl/modyn/blob/61d39e58dc0e24a93f1498c987e4689ed788efac/modyn/storage/include/internal/grpc/storage_service_impl.hpp#L123

modyn/storage/include/internal/grpc/storage_service_impl.hpp
XianzheMa: Yes correct! I noticed it as well. I didn't merge the implementation because
1. in `send_sample_id_and_label` it is ordered by `updated_at` and here by `sample_id`.
2. the response of `sned_sample_id_and_label` is not sharded by `worker_id`.

It is possible to merge the implementation but considering the logic itself is less than 10 lines, and the merge has to take care of the above differences (not a trivial merge), so I thought it is not worth the effort.

XianzheMa:
> thanks Xianzhe. Sorry, this is not an in-depth review. My biggest comments are to double check whether we can reduce code duplication somehow and to extend the integration tests (besides addressing the memory leaks, of course :)).
> 
> Thanks already!
> 
> oh and one more thing: Before merging, please change the PR title to something more comprehensive than the branch name :) I will double check if I get the time before my flight goes. Thanks so much Xianzhe!

Thanks for reviewing! I addressed all the comments.

XianzheMa:
closed in favor of https://github.com/eth-easl/modyn/pull/388, for the reason of branch renaming

PR #382, opened 2024-04-06
Fix Wildtime benchmark_utils create_timestamp() timezone to be UTC

jenny011:
# Bug
The `create_timestamp()` function in `benchmark/wildtime_benchmarks/benchmark_utils.py` uses the **local** timezone. The generated timestamp thus is **not UTC**. 
However, modyn Storage assumes a UTC timestamp so that 1970/1/1 corresponds to 0.
Therefore, if the files are generated in Zurich local time, then 1970/1/1 is converted to -3600 in the `files` table. 

# Fix
Fix tzinfo to be UTC timezone in `create_timestamp()`.

### DIFF
--- a/benchmark/wildtime_benchmarks/benchmark_utils.py
+++ b/benchmark/wildtime_benchmarks/benchmark_utils.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 import pathlib
-from datetime import datetime
+from datetime import datetime, timezone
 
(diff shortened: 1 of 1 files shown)


XianzheMa (approved):
Thanks for spotting and fixing the bug!!!

XianzheMa:
let's merge this!

PR #383, opened 2024-04-07
Generate test/train datasets for yearbook and arxiv

XianzheMa:
This PR modifies the files to generate yearbook and arxiv datasets, so that both train and test datasets are generated.


benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -85,9 +114,9 @@ def create_binary_file(data, output_file_name: str, timestamp: int) -> None:
                 label_integer = tensor2.item()
 
                 features_size = len(features_bytes)
-                assert features_size == 4096
+                assert features_size == 12288
```
XianzheMa, line 118: 4096*3 == 12288, because we have 3 channels now

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
```diff
@@ -42,9 +42,8 @@ data:
   transformations: []
   bytes_parser_function: |
     import torch
-    import numpy as np
     def bytes_parser_function(data: bytes) -> torch.Tensor:
-      return torch.from_numpy(np.frombuffer(data, dtype=np.float32))
+      return torch.frombuffer(bytearray(data), dtype=torch.float32).reshape(3, 32, 32)
```
XianzheMa, line 47: This is to remove a warning. During training this constantly appears:
image

Because the `bytes` itself is not writable and `np.frombuffer` just re-interprets the data as a numpy array, and presents it (as writable data) to pytorch.

The correctness of the modified one is tested by:
```python
import torch
import numpy as np
data = torch.rand(size=(3, 32, 32), dtype=torch.float32)
# save the tensor to a file
with open('test.bin', 'wb') as f:
    f.write(data.numpy().tobytes())


# read the tensor from the file
with open('test.bin', 'rb') as f:
    data_bytes = f.read()

# the old way which produces a warning
# read_data = torch.from_numpy(np.frombuffer(bytes, dtype=np.float32).reshape(3, 32, 32))
# the new way
read_data = torch.frombuffer(bytearray(data_bytes), dtype=torch.float32).reshape(3, 32, 32)

assert torch.allclose(data, read_data)

```

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -17,19 +15,25 @@ def main():
...
+corrupted_idx_dict = {
+    2007: [33213],
+    2008: [22489],
+    2009: [64621, 165454],
+    2015: [42007, 94935],
+    2016: [111398],
+    2019: [41309, 136814],
+    2020: [102074],
+    2021: [32013, 55660]
+}
```
XianzheMa, line 27: there are some lines in the file that are corrupted, i.e. the csv file wrapper cannot properly read the data. This is to remove those lines.

benchmark/wildtime_benchmarks/benchmark_utils.py
```diff
@@ -57,7 +57,7 @@ def setup_logger():
 
 
 def create_fake_timestamp(year: int, base_year: int) -> int:
-    timestamp = ((year - base_year) * DAY_LENGTH_SECONDS) + 1
+    timestamp = ((year - base_year) * DAY_LENGTH_SECONDS)
```
XianzheMa, line 60: not sure why previously there is a `+1`. Without it every timestamp is neater.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -17,19 +15,25 @@ def main():
...
     num_classes = 172
     drive_id = "1H5xzHHgXl8GOMonkb6ojye-Y2yIp436V"
     file_name = "arxiv.pkl"
 
-    def __getitem__(self, idx):
-        return self._dataset["title"][idx], torch.LongTensor([self._dataset["category"][idx]])[0]
-
-    def __len__(self):
-        return len(self._dataset["category"])
-
```
XianzheMa, line 32: they are not used

benchmark/wildtime_benchmarks/benchmark_utils.py
MaxiBoether: Hmm. IIRC, it has something to do with the trigger. It might have been some reasoning like "we need a day + 1 so that the trigger fires" but I think that is actually not true. If this works for the time trigger, then we can get rid of the + 1 :)

benchmark/wildtime_benchmarks/data_generation_arxiv.py
MaxiBoether: Can we add a brief comment here in the file explaining this? And thanks for catching this! Not sure why this did not come up before.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -1,10 +1,8 @@
 import os
 import pickle
 
-import torch
 from benchmark_utils import create_timestamp, download_if_not_exists, setup_argparser_wildtime, setup_logger
 from torch.utils.data import Dataset
-from tqdm import tqdm
```
MaxiBoether, line 7: Why no tqdm anymore :D?

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -44,39 +48,89 @@ def __init__(self,  data_dir):
         self.path = data_dir
 
     def store_data(self, store_all_data: bool, add_final_dummy_year: bool):
-        for year in tqdm(self._dataset):
+        # create directories
+        if not os.path.exists(self.path):
+            os.mkdir(self.path)
+
+        train_dir = os.path.join(self.path, "train")
+        os.makedirs(train_dir, exist_ok=True)
```
MaxiBoether, line 56: As a general note, I prefer `pathlib.Path` for such operations (esp. joining paths), it reads a bit more idiomatic. But since this is just data generation code, we can keep os.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -44,39 +48,89 @@ def __init__(self,  data_dir):
...
                     csv_row = f"{text}\t{label}"
-                    year_rows.append(csv_row)
+                    rows.append(csv_row)
+                return rows
 
-            # store the year file
-            text_file = os.path.join(self.path, f"{year}.csv")
-            with open(text_file, "w", encoding="utf-8") as f:
-                f.write("\n".join(year_rows))
+            train_year_rows = get_one_split(0)
```
MaxiBoether, line 77: This more like a general question: If I understand correctly, the pre-supplied arXiv data already has a test/train split, and we use that, correct? Have we actually validated that this split is balanced, in some sense? What I mean is that whoever created the dataset might not have created a balanced test split, it could contain many samples from e.g. 2007 and 0 samples from 2009 or something like that. 

If this is the case, we might need to merge the train+test data supplied from the arxiv dataset, and then manually create a test split per year. Actually, I thought this would happen in this PR. It's okay to use the pre-defined split, but we should validate that the split behaves as expected

benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -37,21 +37,33 @@ def __init__(self, data_dir: str):
         self._dataset = datasets
         self.data_dir = data_dir
```
MaxiBoether, line 39: The same comment about the pre-defined split applies here - we should validate this.

benchmark/wildtime_benchmarks/data_generation_yearbook.py
```diff
@@ -85,9 +122,9 @@ def create_binary_file(data, output_file_name: str, timestamp: int) -> None:
                 label_integer = tensor2.item()
 
                 features_size = len(features_bytes)
-                assert features_size == 4096
+                assert features_size == 12288
 
-                f.write(int.to_bytes(label_integer, length=4, byteorder="big"))
+                f.write(int.to_bytes(label_integer, length=4, byteorder="little"))
```
MaxiBoether, line 127: I think after merging #398 we should default to big since it's cheaper to parse the big endian files at the storage (no reverse iterator needed)

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
MaxiBoether: I am a bit confused here, unfortunately. What gets passed is a `memoryview`, not `bytes` (and this PR only changes the type hint, which is wrong now). Because the OnlineDataset will always pass a memoryview. Changing the type hint to bytes here should not actually impact the code, it's just a hint, but in this case (from my understanding) a wrong hint, since the parsing function always gets passed a memoryview.

The reason for the old code is that we want to avoid unnecessary in-memory copies. We will never write to this tensor (so it will not cause undefined behavior). With your new version, we create a new `bytearray`, which copies the data out of the memoryview. That's exactly what the old code tries to avoid. I think a better approach would be a way to supress the log this line causes and use the (more efficient) old code:

```
import warnings
import torch

def bytes_parser_function(data: bytes) -> torch.Tensor:
    with warnings.catch_warnings():
        warnings.simplefilter("ignore", category=UserWarning)
        return torch.frombuffer(data, dtype=torch.float32).reshape(3, 32, 32)
```

MaxiBoether (changes requested):
Thank you Xianzhe! I had some comments/feedback, just to double check we are doing the right thing :) But the general changes look good!

benchmark/wildtime_benchmarks/benchmark_utils.py
XianzheMa: Yeah it works for my past experiments. Also @jenny011 confirmed as well without `+1` everything is easier to manage.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
XianzheMa: This whole process is very quick so we don't need to keep track of how much we processed and how much is left. On the contrary the `tqdm` just adds noise to the terminal.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -44,39 +48,89 @@ def __init__(self,  data_dir):
...
     def store_data(self, store_all_data: bool, add_final_dummy_year: bool):
-        for year in tqdm(self._dataset):
+        # create directories
+        if not os.path.exists(self.path):
+            os.mkdir(self.path)
+
+        train_dir = os.path.join(self.path, "train")
+        os.makedirs(train_dir, exist_ok=True)
+
+        if store_all_data:
```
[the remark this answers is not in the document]
XianzheMa: If not given, the test data is not stored.
Yeah you are right I will change the name.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
```diff
@@ -44,39 +48,89 @@ def __init__(self,  data_dir):
...
+                test_file = os.path.join(test_dir, f"{year}.csv")
+                with open(test_file, "w", encoding="utf-8") as f:
+                    f.write("\n".join(test_year_rows))
+
+                # set timestamp
+                os.utime(test_file, (year_timestamp, year_timestamp))
+                print(f"for year {year} train size {len(train_year_rows)} test size {len(test_year_rows)}")
+                stats[year] = {"train": len(train_year_rows), "test": len(test_year_rows)}
+            else:
+                print(f"for year {year} train size {len(train_year_rows)}")
```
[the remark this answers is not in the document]
XianzheMa: I removed it. Anyway I also stored an `overall_stats.json` which stores these statistics.

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
XianzheMa: Sorry, I should change back the type hint.

I was just not sure if the warning really hurts anything or it is okay. But your explanation makes sense! If we are sure what returns from this function is never overwritten, we should keep the old implementation.

benchmark/wildtime_benchmarks/data_generation_arxiv.py
XianzheMa: Yes the pre-supplied dataset already has a test/train split, and we use it (also I think the wildtime paper uses it as well). It has a ~10:1 ratio (train vs test), so it's quite balanced

XianzheMa: fyi
image

benchmark/wildtime_benchmarks/data_generation_yearbook.py
XianzheMa: It's also a 10:1 ratio, well balanced!

benchmark/wildtime_benchmarks/data_generation_arxiv.py
MaxiBoether: Ok, thank you, those numbers look reasonable!

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
```diff
@@ -41,10 +41,12 @@ data:
...
   transformations: []
   bytes_parser_function: |
+    import warnings
     import torch
-    import numpy as np
     def bytes_parser_function(data: memoryview) -> torch.Tensor:
-      return torch.from_numpy(np.frombuffer(data, dtype=np.float32))
+      with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+      return torch.frombuffer(data, dtype=torch.float32).reshape(3, 32, 32)
```
MaxiBoether, line 49: I try to remember why we use torch.from_numpy and np.from_buffer instead of torch.frombuffer. From a quick test, they seem to behave identical:

```
>>> byte_array = bytearray('WXYZ', 'utf-8')
>>> mv = memoryview(byte_array)
>>> torch.frombuffer(mv, dtype=torch.float32)
tensor([1.5294e+16])
>>> torch.from_numpy(np.frombuffer(mv, dtype=np.float32))
tensor([1.5294e+16])
```

But there should be a reason I did this (we played around with it quite a bit, but tbh I do not remember right now what the problem was). If you tested this within the online dataset and it works, I guess it should be fine. I am a tad confused :D

MaxiBoether (approved):
Thank you Xianzhe! Feel free to merge, although I am sure there was a reason I did the intermediate numpy step...

benchmark/wildtime_benchmarks/example_pipelines/yearbook.yaml
MaxiBoether: https://github.com/search?q=repo%3Aeth-easl%2Fmodyn%20frombuffer&type=code

For Criteo, I also used frombuffer. I am confused. What did I do here :D

PR #384, opened 2024-04-07
Correct assertion on yearbook image size

XianzheMa:
In a previous PR https://github.com/eth-easl/modyn/pull/377 the number of channels in an image is changed from 1 to 3.
However the corresponding assertion on image size was not changed.
This PR changes it: `4096 = 12288`


XianzheMa:
not a mature PR

Issue #275, opened 2023-06-23
Common Dataset Abstraction


robin-oester:
After we merge #256 which induces several structural changes to the `OnlineDataset` and #269 (introducing `ValidationDataset`), we can create a common abstraction for Datasets. We should probably put this in the `common` folder.

MaxiBoether:
In the context of this, we should also implement integration tests for the different datasets.

PR #388, opened 2024-04-09
Time-slice the EvaluationDataset

XianzheMa:
This is the first step towards evaluating on a separate dataset: to time-slice the `EvaluationDataset`.

Specifically, we add two parameters to the class `EvaluationDataset`: `start_timestamp`, `end_timestamp`,
to only select samples falling into this interval from the evaluation dataset.

We hence also add these two parameters to a storage grpc API `GetDataPerWorker`.

We also add these parameters to another API `GetDatasetSize` because it's used in evaluator


integrationtests/storage/integrationtest_storage.py
```diff
@@ -122,15 +125,59 @@ def check_data_per_worker() -> None:
...
+
+        assert len(responses) == 1, f"Received batched response or no response, shouldn't happen: {responses}"
+
+        response_keys_size = len(responses[0].keys)
+
+        assert response_keys_size == 5 if worker_id <= 1 else response_keys_size == 4
+
+    split_ts1 = IMAGE_UPDATED_TIME_STAMPS[9] + 1
+    split_ts2 = IMAGE_UPDATED_TIME_STAMPS[19] + 1
+
```
XianzheMa, line 141: The rest tests are against `GetDataPerWorkerRequest` with `start_timestamp` or `end_timestamp` set.

modyn/storage/src/internal/filesystem_wrapper/local_filesystem_wrapper.cpp
```diff
@@ -66,22 +66,15 @@ uint64_t LocalFilesystemWrapper::get_file_size(const std::string& path) {
...
 
-  const auto modified_time = std::filesystem::last_write_time(path);
-  const auto cftime = to_time_t(modified_time);
-  return static_cast<int64_t>(cftime);
-
+  // there is no portable way to get the modified time of a file in C++17 and earlier
+  struct stat file_attribute {};
+  stat(path.c_str(), &file_attribute);
+  return static_cast<int64_t>(file_attribute.st_mtime);
   /* C++20 version, not supported by compilers yet */
```
XianzheMa, line 78: this is needed as the previous C++ way is not accurate. Sometimes a different modified time is read into the system (due to inaccurate data type converstion). I rewrite it in C because prior to C++20 there is no absolutely correct way in C++.

modyn/evaluator/internal/dataset/evaluation_dataset.py
```diff
@@ -34,6 +34,8 @@ def __init__(
         storage_address: str,
```
MaxiBoether, line 34: Now that the evaluation dataset is becoming more and more different from the onlinedataset, I wonder whether we should have a separate EvaluationDataset integrationtest, similar to test_online_dataset.py. Does not need to be part of this PR, but maybe we should create an issue to write a test here

modyn/storage/src/internal/filesystem_wrapper/local_filesystem_wrapper.cpp
MaxiBoether: I actually don't understand the problem yet. What do you mean by "Sometimes a different modified time is read into the system (due to inaccurate data type converstion)"? At the time, I researched this quite a bit and this seemed to be the most reliable implementation. We should maybe also check whether compiler support for C++20 chrono has improved since writing this. 

Anyways, I don't understand yet what the problem is with the std::filesytem approach, since that should be realiable. Could you maybe clarify why you intend to change this?

modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -520,6 +650,30 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
...
+    int64_t total_keys = 0;
+    std::string timestamp_filter;
+    if (start_timestamp > 0 && end_timestamp == 0) {
+      timestamp_filter = fmt::format("updated_at >= {}", start_timestamp);
+    } else if (start_timestamp == 0 && end_timestamp > 0) {
+      timestamp_filter = fmt::format("updated_at < {}", end_timestamp);
+    } else if (start_timestamp > 0 && end_timestamp > 0) {
+      timestamp_filter = fmt::format("updated_at >= {} AND updated_at < {}", start_timestamp, end_timestamp);
+    } else if (start_timestamp == 0 && end_timestamp == 0) {
+      timestamp_filter = "1 = 1";
```
MaxiBoether, line 665: Mhm, alternatively, we could have the empty string and append AND to the filter in other cases, because this reads a bit weird at first. At least, we should add a small comment that we need some True statement in this case to keep valid SQL below

modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -186,6 +186,136 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
...
+            timestamp_filter = fmt::format("updated_at >= {}", start_timestamp);
+          } else if (start_timestamp == 0 && end_timestamp > 0) {
+            timestamp_filter = fmt::format("updated_at < {}", end_timestamp);
+          } else if (start_timestamp > 0 && end_timestamp > 0) {
+            timestamp_filter = fmt::format("updated_at >= {} AND updated_at < {}", start_timestamp, end_timestamp);
+          } else if (start_timestamp == 0 && end_timestamp == 0) {
+            timestamp_filter = "1 = 1";
+          } else {
+            FAIL(fmt::format("Invalid timestamps: start = {}, end = {}", start_timestamp, end_timestamp));
+          }
```
MaxiBoether, line 238: There is quite some code duplication here with the counting function. Could we extract this into a common function?

MaxiBoether (changes requested):
Thank you (also for keeping the scope of this PR small - really helps in reviewing)! Looks good, just had minor comments, after which we can merge :)

modyn/storage/src/internal/filesystem_wrapper/local_filesystem_wrapper.cpp
XianzheMa: We here need to convert `std::filesystem::file_time_type` to `time_t`.
The previous way resorts to `TP::clock::now()` and `system_clock::now()`, which are two separate calls to the current time and might produce a difference. So it is inaccurate. Also @jenny011 confirmed this inaccuracy in her experiments, where the some files with modified time `**00` is read as `**99`.

I believe the previous implementation consulted this [stackoverflow QA](https://stackoverflow.com/questions/61030383/how-to-convert-stdfilesystemfile-time-type-to-time-t) because the impl is exactly the same.

modyn/evaluator/internal/dataset/evaluation_dataset.py
XianzheMa: I didn't get it? I think this PR just augments two storage APIs, i.e. `GetDataPerWorkerRequest` and `GetDatasetSizeRequest`, which are on the storage server side, so the current integration tests are in `integrationtest_storage.py`.

modyn/storage/include/internal/grpc/storage_service_impl.hpp
XianzheMa: Thanks for spotting this out!

modyn/storage/include/internal/grpc/storage_service_impl.hpp
XianzheMa: I added a comment, because it looks neater to have `AND` outside
image

modyn/storage/src/internal/filesystem_wrapper/local_filesystem_wrapper.cpp
MaxiBoether: Okay, then we use the C implementation.

modyn/evaluator/internal/dataset/evaluation_dataset.py
MaxiBoether: I extended issue #275. Just a note, (integration) tests are not necessarily driven by complexity. In a perfect world, unit tests mock away everything except the function-under-test. The idea of an integration test is to verify whether a component works end-to-end. They are two different perspectives on testing. Anyways, nothing that affects this PR.

MaxiBoether (approved):
Sorry for this taking so long, looks good to me!

XianzheMa:
> Sorry for this taking so long, looks good to me!

No worries all good! This PR is a bit long and my progress is not delayed by anything :)
Thanks for taking time to review it!

PR #389, opened 2024-04-11
fix volume path

XianzheMa:
Actually where evaluation results are stored is `/tmp/evaluation_results`... This is a bug.

The bug causes that the pipeline log is not persistent if supervisor container exits.


jenny011 (approved):
LGTM

PR #393, opened 2024-04-17
fix: Load file labels with appropriate `byteorder`

robinholzi:
# Motivation

`BinaryFileWrapper::int_from_bytes` previously used hard-coded preprocessor definitions for determining the endianness. We now want to use the `byteorder` specified in the `filesystem_wrapper_config`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     command: postgres -c config_file=/etc/postgresql/postgresql.conf
 #    shm_size: 24gb
     volumes:
-      - ./metadata-postgres-data:/var/lib/postgresql/data
+      - .data/metadata-postgres-data:/var/lib/postgresql/data
       - ./conf/metadata_postgresql.conf:/etc/postgresql/postgresql.conf
       - ./conf/pg_hba.conf:/tmp/pg_hba.conf
       - ./conf/init_pg_hba.sh:/docker-entrypoint-initdb.d/init_pg_hba.sh
@@ -38,7 +38,7 @@ services:
     command: postgres -c config_file=/etc/postgresql/postgresql.conf
 #    shm_size: 8gb
     volumes:
-      - ./storage-postgres-data:/var/lib/postgresql/data
+      - .data/storage-postgres-data:/var/lib/postgresql/data
       - ./conf/storage_postgresql.conf:/etc/postgresql/postgresql.conf
       - ./conf/pg_hba.conf:/tmp/pg_hba.conf
       - ./conf/init_pg_hba.sh:/docker-entrypoint-initdb.d/init_pg_hba.sh
@@ -57,6 +57,7 @@ services:
       dockerfile: docker/Storage/Dockerfile
     volumes:
       - storage-data:/app/storage
+      - ./modyn/config/examples:/src/modyn/config/examples/
 #      - /mnt/datasets:/datasets
 #      - .:/modyn_host
   metadata_processor:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -57,6 +57,7 @@ services:
       dockerfile: docker/Storage/Dockerfile
     volumes:
       - storage-data:/app/storage
+      - ./modyn/config/examples:/src/modyn/config/examples/
 #      - /mnt/datasets:/datasets
```
XianzheMa, line 61: why are these three changes are needed?

modyn/storage/src/internal/file_wrapper/binary_file_wrapper.cpp
```diff
@@ -61,13 +58,25 @@ std::vector<int64_t> BinaryFileWrapper::get_all_labels() {
...
+      label_vec.clear();
+    }
+  }
+  else {
+    for (uint64_t i = 0; i < num_samples; ++i) {
+      get_stream()->seekg(static_cast<int64_t>(i * record_size_), std::ios::beg);
+      get_stream()->read(reinterpret_cast<char*>(label_vec.data()), static_cast<int64_t>(label_size_));
+      labels.push_back(int_from_bytes_big_endian(label_vec.data(), label_vec.data() + label_size_));
+      label_vec.clear();
+    }
```
XianzheMa, line 79: so I guess previously in `get_all_labels` they even didn't take endianness into account right? Thanks for the catch!

XianzheMa (changes requested):
Looks great! Can you add some unit tests against the changes? (and also why changes in docker-compose.yml are needed)
 
We have a CI system that type-checks and style-checks the code, and also execute the tests. Let me know if you have issues with passing any check.

docker-compose.yml
robinholzi: I found it useful for debugging to be able to adjust the `modyn_config.yaml` and have it reflected in the container. I guess this mapping doesn't do any harm. If you don't find it useful we don't have to include it in the commit :)

modyn/storage/src/internal/file_wrapper/binary_file_wrapper.cpp
robinholzi: Previously get_all_labels was agnostic of the `endianness` and relied on `int_from_bytes` to handle it. I could have done it there and let get_all_labels the same. However, I guess it makes more sense to not do the check for `byteorder` in every iteration of the for loop (potentially every 4-byte block) read-in.

robinholzi:
@MaxiBoether could you please add me as a codebase collaborator, I cannot trigger CI workflows. Thank you! :)

MaxiBoether:
Sorry about that, you should have received an invite!

docker-compose.yml
XianzheMa: I think it takes more steps than just mapping the config: you need to restart the servers in the container. (which also may not always work currently because with the current impl, some other servers may just crash when failing to connect to the restarting ones...)

Also it might confuse others as we already copy the files into the container [here](https://github.com/eth-easl/modyn/blob/ed861c92cc05f7ab99b2694df4f1281234539639/docker/Base/Dockerfile#L4).

Anyway let's not do this in this commit

robinholzi: Okay, I did a `docker exec` into the container and triggered the script there. In that context having a mapped file made sense. But I agree that it might not be useful in other situations, let's remove it!

PR #394, opened 2024-04-17
Remove noisy logs in `file_watchers.cpp`

XianzheMa:
FileWatcher periodically monitors file status and hence is very active. Its highly frequent logs will hide the more important logs from the storage server (i.g. when processing a request.) This PR removes many noisy logs.

### DIFF
--- a/modyn/storage/include/internal/grpc/storage_service_impl.hpp
+++ b/modyn/storage/include/internal/grpc/storage_service_impl.hpp
@@ -230,6 +230,8 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
     session.close();
 
     if (file_ids.empty()) {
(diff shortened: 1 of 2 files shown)


modyn/storage/include/internal/grpc/storage_service_impl.hpp
```diff
@@ -230,6 +230,8 @@ class StorageServiceImpl final : public modyn::storage::Storage::Service {
     session.close();
 
     if (file_ids.empty()) {
+      SPDLOG_INFO("No files found for dataset {} with start_timestamp = {} and end_timestamp = {}", dataset_id,
+                  start_timestamp, end_timestamp);
```
XianzheMa, line 234: when no data is found, e.g. due to bug in CSVFileWrapper for corrupted lines, this log is very useful.

MaxiBoether (approved):
Thanks Xianzhe!

PR #395, opened 2024-04-18
Correct arxiv settings in modyn_config and corresponding pipeline config

XianzheMa:
Currently in `modyn_config.yaml` somehow there are two definitions on the arxiv dataset, (where clearly we should have only one...) I removed the wrong one as the arxiv dataset is in `csv` format.

Also, a bug on `byte_parser_function` is corrected.

### DIFF
--- a/benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
+++ b/benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
@@ -40,7 +40,7 @@ training:
 data:
   dataset_id: arxiv
   bytes_parser_function: |
-    def bytes_parser_function(data: bytes) -> str:
+    def bytes_parser_function(data: memoryview) -> str:
       return str(data, "utf8")
   tokenizer: DistilBertTokenizerTransform
 
--- a/benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
+++ b/benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
@@ -15,7 +15,7 @@ training:
   dataloader_workers: 2
   use_previous_model: True
   initial_model: random
-  batch_size: 96
+  batch_size: 128
   optimizers:
     - name: "default"
       algorithm: "SGD"
@@ -41,7 +41,7 @@ data:
   dataset_id: arxiv
   bytes_parser_function: |
     def bytes_parser_function(data: bytes) -> str:
-      return data.decode("utf-8")
+      return str(data, "utf8")
   tokenizer: DistilBertTokenizerTransform
 
 trigger:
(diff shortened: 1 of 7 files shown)


benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
```diff
@@ -15,7 +15,7 @@ training:
   dataloader_workers: 2
   use_previous_model: True
   initial_model: random
-  batch_size: 96
+  batch_size: 128
   optimizers:
```
XianzheMa, line 19: 128 is the largest size that fits on our gpus. We set 128 as default to make it convenient for new people to use

robinholzi: What would be the implications of using smaller `batch_sizes` like `96`?

modyn/config/examples/modyn_config.yaml
```diff
@@ -82,22 +82,6 @@ storage:
         file_watcher_interval: 5,
         selector_batch_size: 1024,
       },
-      {
```
robinholzi, line 85: Makes sense!

benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
```diff
@@ -41,7 +41,7 @@ data:
   dataset_id: arxiv
   bytes_parser_function: |
     def bytes_parser_function(data: bytes) -> str:
-      return data.decode("utf-8")
+      return str(data, "utf8")
```
robinholzi, line 44: Any particular reason to change this, in my experience `bytes.decode(...)` is more idiomatic?https://stackoverflow.com/a/14473492

robinholzi (approved):
lgtm, just two non blocking questions👍

benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
XianzheMa: Usually it's a common practice to use as a large batch size as possible (until the gpu cannot accommodate it), to speed up training. It could, in some cases, come with a cost of losing accuracy in the end, but as we compare different policies and keep the same batch size, it's fair. (i.e. we don't care how high the absolute accuracy can be, but how higher the accuracy one policy can reach compared to another).

benchmark/wildtime_benchmarks/example_pipelines/arxiv.yaml
XianzheMa: hmm I should change the type hint...
the reason is that `data.decode` does not work here as the input is not bytes but a `memoryView` object that has no `.decode` method...

PR #397, opened 2024-04-19
add the missing tokenizer in evaluation setting

XianzheMa:
We miss the tokenizer in the evaluation so that the NLP datasets cannot be evaluated against.
This PR supplements it.


jenny011 (approved):
LGTM!

Issue #368, opened 2024-02-13
C++ Storage: `BinaryFileWrapper` ignores `byteorder` option


MaxiBoether:
The Python BinaryFileWrapper used to have a configuration option `byteorder` which states whether the binary file stores the label in little or big endian.  

The C++ file wrapper currently ignores this option completely. Right now, it assumes that the stored file uses little endian file order (which works for Criteo and is why we have not noticed it, but yearbook uses big endian). We should re-add support for big endian file order. For big endian, we do not need the reverse logic that we have for little endian data.

PR #398, opened 2024-04-19
fix: Load file labels with appropriate `byteorder`

robinholzi:
# Motivation

`BinaryFileWrapper::int_from_bytes` previously used hard-coded preprocessor definitions for determining the endianness. We now want to use the `byteorder` specified in the `filesystem_wrapper_config`.

successor of #393 (this doesn't use branch in repo fork)


modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
```diff
@@ -48,6 +49,7 @@ class BinaryFileWrapperTest : public ::testing::Test {
 TEST_F(BinaryFileWrapperTest, TestGetNumberOfSamples) {
   const std::shared_ptr<std::ifstream> stream_ptr = std::make_shared<std::ifstream>();
   stream_ptr->open(file_name_, std::ios::binary);
+  ASSERT_TRUE(stream_ptr->is_open());
```
robinholzi, line 52: On the sg's machines I ran into the issue that the test file was never created due to a permissions error (another user has already created the test directory). The file descriptor was then never opened leading `binary_file_wrapper.cpp:get_stream` to call the mocked `filesystem_wrapper_->get_stream` more than once as every `stream_->is_open()` returned `false`. To avoid this happening I propose assertion that with a test precondition.

.gitignore
```diff
@@ -78,3 +78,9 @@ plots/
 cmake-build-debug/
 clang-tidy-build/
 libbuild/
+
+
+# Data & config files
+
+.data/
+.env
```
XianzheMa, line 86: why is this needed?

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
XianzheMa: I didn't really get this. How can this `ASSERT_TRUE` help with creating the test file?

XianzheMa (changes requested):
Looks great, but where is the unit test on the new endianness-aware method?

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
robinholzi: It cannot help create it, but it can prevent the tested logic from starting. That saves some debugging time as the error message in the storage component doesn't indicate the origin of this issue.

.gitignore
robinholzi: Just a quality-of-life improvement for developers like me who like to store data and environment variables in the repos directory without adding it to git vcs.

robinholzi:
@XianzheMa I added unittests now, even though there weren't any tests on the `byteorder` before. Thanks for the reminder :)

MaxiBoether:
I will have a deeper look tomorrow, but one general note:

The idea behind the preprocessor definitions was to efficiently load little endian files depending on the architecture at hand. As described in #368, we only cover the two cases: file is LE and machine is LE, and file is LE and machine is BE. We missed the functionality in case the file is stored in big endian. I just briefly skimmed over it and you seem to remove the macros to check for LE/BE machines. 

I think this might be problematic since how to read a LE/BE file depends on the machine endianess. The two functions were for reading LE files on LE/BE machines. It might be that this also works for reading BE files on a LE machine (need to think about it - sorry too late right now :D) but I think we need a function `read_le_file` that internally has a macro to check for machine endianess and a function `read_be_file` that does the same and then dispatch the call according to the config whether the file was stored using LE or BE. I might be wrong though, need to take a deeper look tomorrow :)

robinholzi:
> I will have a deeper look tomorrow, but one general note:
> 
> The idea behind the preprocessor definitions was to efficiently load little endian files depending on the architecture at hand. As described in #368, we only cover the two cases: file is LE and machine is LE, and file is LE and machine is BE. We missed the functionality in case the file is stored in big endian. I just briefly skimmed over it and you seem to remove the macros to check for LE/BE machines.
> 
> I think this might be problematic since how to read a LE/BE file depends on the machine endianness. The two functions were for reading LE files on LE/BE machines. It might be that this also works for reading BE files on a LE machine (need to think about it - sorry too late right now :D) but I think we need a function `read_le_file` that internally has a macro to check for machine endianness and a function `read_be_file` that does the same and then dispatch the call according to the config whether the file was stored using LE or BE. I might be wrong though, need to take a deeper look tomorrow :)

@MaxiBoether Thanks for the feedback! Let's talk about that in person tomorrow. I had the same concerns during implementation. 

Afaik these two parsing functions should are agnostic of the system's byteorder. The bit-shift operators on 8-byte types handle the variables' (like accumulator) byte sequence without us having to care about the endian. 
The byte order of the files (i.e. the char sequences) is relevant though. As we accumulate byte-wise by shifting the accumulator to the higher valued bytes, we have to accumulate bytes in descending order (most significant byte first). By doing so the most significant bytes will end up in the higher valued bytes in memory (regardless of what the system endian is). As big endian files are already in most significant byte first order we don't have to reverse there and vice versa.

In case we knew the label size in advance though, we could replace the byte-wise `std::accumulate` with a reinterpret cast IFF `system byte order` = `file byte order`. Presumably, that would be more efficient.

modyn/storage/include/internal/file_wrapper/binary_file_wrapper.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <spdlog/spdlog.h>
 
+#include <cctype>
```
MaxiBoether, line 5: For which function is this needed?

MaxiBoether:
Okay, let's chat after lunch. But yes, you're right. Actually, reading the data is independent of the machine endianess. Since we just build up an int, the only thing that matters is how the file was saved. So I agree with you. But let's chat briefly, and then I will do the full review of this PR.

MaxiBoether:
> > Okay, let's chat after lunch. But yes, you're right. Actually, reading the data is independent of the machine endianess. Since we just build up an int, the only thing that matters is how the file was saved. So I agree with you. But let's chat briefly, and then I will do the full review of this PR.
> 
> Yes how data is stored should be (and actually is) independent of the machine arch. We actually specify the data endianness in the [modyn config](https://github.com/eth-easl/modyn/blob/caec6226c1ac228fdbbdbae7df6793affae2ffdc/modyn/config/examples/modyn_config.yaml#L42), and when [storing the generated dataset](https://github.com/eth-easl/modyn/blob/caec6226c1ac228fdbbdbae7df6793affae2ffdc/benchmark/wildtime_benchmarks/data_generation_yearbook.py#L90), we specify the endianess

What I was referring to is that reading a file that was stored in LE follows the same logic on both LE and BE machines :D But yes, also what you are referring to (the machine arch should not define the endianess of the file) is also true.

modyn/storage/src/internal/file_wrapper/binary_file_wrapper.cpp
```diff
@@ -40,10 +37,13 @@ int64_t BinaryFileWrapper::get_label(uint64_t index) {
 
   get_stream()->seekg(static_cast<int64_t>(label_start), std::ios::beg);
 
-  std::vector<unsigned char> label_vec(label_size_);
-  get_stream()->read(reinterpret_cast<char*>(label_vec.data()), static_cast<int64_t>(label_size_));
+  std::vector<unsigned char> tmp_label_buffer(label_size_);
+  get_stream()->read(reinterpret_cast<char*>(tmp_label_buffer.data()), static_cast<int64_t>(label_size_));
 
-  return int_from_bytes(label_vec.data(), label_vec.data() + label_size_);
+  if (this->little_endian_) {
```
MaxiBoether, line 43: No need for the `this` pointer (I guess it is a coding style question, but we never used it in the codebase)

modyn/storage/src/internal/file_wrapper/binary_file_wrapper.cpp
```diff
@@ -61,13 +61,24 @@ std::vector<int64_t> BinaryFileWrapper::get_all_labels() {
...
-    get_stream()->seekg(static_cast<int64_t>(i * record_size_), std::ios::beg);
-
-    std::vector<unsigned char> label_vec(label_size_);
-    get_stream()->read(reinterpret_cast<char*>(label_vec.data()), static_cast<int64_t>(label_size_));
-
-    labels.push_back(int_from_bytes(label_vec.data(), label_vec.data() + label_size_));
+  std::vector<unsigned char> tmp_label_buffer(label_size_);
+
+  // Duplicating the loop here to avoid doing the endianess check in every loop iteration;
+  // there might be more idomatic ways to do this though.
```
MaxiBoether, line 67: We should use a function pointer here instead. Something like

```
auto int_from_bytes = little_endian_ ? int_from_bytes_little_endian : int_from_bytes_big_endian
``` 

and then just call `int_from_bytes`. No need to duplicate the loop, then :)

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
```diff
@@ -24,30 +25,70 @@ class BinaryFileWrapperTest : public ::testing::Test {
...
     file_name_ = tmp_dir_ + "/test.bin";
+    file_name_endian_ = tmp_dir_ + "/test_endian.bin";
   }
 
   void SetUp() override {
     std::filesystem::create_directory(tmp_dir_);
 
     std::ofstream file(file_name_, std::ios::binary);
-    const std::vector<std::pair<uint32_t, uint16_t>> data = {{42, 12}, {43, 13}, {44, 14}, {45, 15}};
+    const std::vector<std::pair<uint16_t, uint16_t>> data = {{42, 12}, {43, 13}, {44, 14}, {45, 15}};
```
MaxiBoether, line 35: Why this change?

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
```diff
@@ -24,30 +25,70 @@ class BinaryFileWrapperTest : public ::testing::Test {
...
     }
     file.close();
+
+    // tmp test file with 2 byte labels
+    std::ofstream file_endian(file_name_endian_, std::ios::binary);
+    const std::vector<std::pair<uint16_t, uint16_t>> data_endian = {
+        {(1u << 8u) + 2u, 0}, {(3u << 8u) + 4u, 1}, {(5u << 8u) + 6u, 2}, {(7u << 8u) + 8u, 3}};
+    for (const auto& [payload, label] : data_endian) {
+      // note: on macOS & linux, the architecture's endian is little
+      payload_to_file(file_endian, payload, label);
```
MaxiBoether, line 47: This does not depend on the OS, but on the processor. For example, you can run Linux on IBM POWER CPUs that are big endian. I am not sure whether the test currently assumes that it's always run on a LE machine, but in the best case, we would not have such implicit assumptions

MaxiBoether (changes requested):
I actually took a look already since the big discussion was already resolved. I had some comments/questions though. After addressing these, we can merge (although I really wonder about the yearbook sample size that changes in this PR)

modyn/config/examples/modyn_config.yaml
```diff
@@ -58,7 +58,7 @@ storage:
         file_wrapper_config:
           {
             byteorder: "big",
-            record_size: 4100,
+            record_size: 12292,
```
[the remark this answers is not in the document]
XianzheMa: It faced a problem. I also changed this. This change comes from the a [previous merged PR](https://github.com/eth-easl/modyn/pull/377) which increases the channel of a picture from 1 to 3. The previous PR does not fully incorporate all the necessary changes (for example this one here)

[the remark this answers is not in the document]
MaxiBoether: Ahhh, okay, that's good to hear :)

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
robinholzi: The only place we a using those values is in the `payload_to_file` function. It takes `uint16_t` inputs. As we don't need 4 bytes for our tests, I went for `uint16_t` over `uint32_t` everywhere.

modyn/storage/include/internal/file_wrapper/binary_file_wrapper.hpp
robinholzi: Does not seem to be needed anymore, thanks!

modyn/storage/src/internal/file_wrapper/binary_file_wrapper.cpp
robinholzi: That's neat, wasn't sure if this works in c++ :)

modyn/tests/storage/internal/file_wrapper/binary_file_wrapper_test.cpp
robinholzi: That's a valid point, the test file endian depends on the system running the tests. Running the test file creation on big-endian would yield different results. I will make the file creation endian independent by writing bytes instead of `uint16_t`s

robinholzi:
@MaxiBoether Could you take a final look here? Would be cool if you could approve before I resolve the merge conflicts - I'll do that tonight. With these long CI runs it's hard to keep a branch updated when other people keep merging things :)

MaxiBoether:
I think @XianzheMa needs to approve here, since you also requested a review from him :)

XianzheMa (approved):
LGTM

XianzheMa:
I think we just need to specify the protobuf version. In general I suppose first passing all the tests before merging because later the CI will fail on main branch again (see my PRhttps://github.com/eth-easl/modyn/commit/3038049d4eac1d7c4995573122d11109c50af440)

PR #400, opened 2024-04-25
test: Add user specific global `tmpdir`

robinholzi:
# Motivation

On the Systems group's machines, I found that running the tests suite locally didn't work as creating temporary files in the temporary directories (`/tmp/binary_file_wrapper_test`, ...) failed. These file creation issue were silent though so you wouldn't know that the tests file don't exist until the test fixtures were called more than once (see [more details here](https://github.com/eth-easl/modyn/pull/398#discussion_r1574539778)).
The reason leading to the files not being created is that the tmp directories (`tmp/binary_file_wrapper_test`, ...) were already created by another user leading my user to miss write access on them.

# Changes

Therefore I propose to use user-specific tmp directories `/tmp/<username>/<tmpdir_name like binary_file_wrapper_test>`.


modyn/tests/cpp_test_main.cpp
```diff
@@ -1,6 +1,23 @@
-#include "gtest/gtest.h"
+#include <gtest/gtest.h>
+
+#include <filesystem>
+#include <iostream>
+
+#include "test_utils.hpp"
+
+// Global fixtures run before unit tests are executed
+class GlobalTestEnvironment : public ::testing::Environment {
```
MaxiBoether, line 9: Could we not move this logic into `get_tmp_testdir`?  Then we do not need the global test environment. Although I am not sure what the cleaner option is here. I am just wondering because currently, all tests have their own subdirectory (`/tmp/user/bla`). The tests need to create that directory anyways, and hence could also just recursively create `/tmp/user` if it does not exist. In this case, we would not to create the /tmp/user directory explicitly. Maybe I am also missing something^^

modyn/tests/common/trigger_sample/test_trigger_sample_storage.cpp
```diff
@@ -4,6 +4,7 @@
 #include <array>
 #include <filesystem>
 
+#include "../../utils/test_utils.hpp"
```
MaxiBoether, line 7: `#include "test_utils.hpp"` should be sufficient, due to our CMake setup.

MaxiBoether (changes requested):
Thank you! Just two small comments, after that we can merge.

modyn/tests/cpp_test_main.cpp
robinholzi: I don't have a strong preference for where to put this logic. Making a couple of extra `create_directory` isn't harmful within the tests. Coming from a Python background I would have searched for some global fixture in a conftest file. I guess `::testing::Environment` is the closest thing to that in C++. 
We can't rely on recursive parent construction though:

> the parent directory must already exist

https://en.cppreference.com/w/cpp/filesystem/create_directory

MaxiBoether: Okay, if we need to create the parent_directory explicitly anyways, this is okay :-) Thank you!

MaxiBoether (approved):
Thanks, feel free to merge!

PR #401, opened 2024-04-26
Fix unit test failure on main

XianzheMa:
I suspect that it is because unfixed version of PyTorch on how to persist the `fused` option.


robinholzi (approved):
Thanks @XianzheMa!

PR #409, opened 2024-04-30
Add DataDriftTrigger: supports one Evidently metric

jenny011:
This is a clean version of PR#367.
1. Add DataDriftTrigger class to supervisor. Supports one configurable Evidently metric. Launches drift detection every N new data points. Data used in detection are data trained in the previous trigger and all the untriggered new data.
2. Update Trigger interface. `Trigger.inform()` returns a Generator instead of List.
3. Add a generic ModelDownloader in supervisor.
4. Add example pipelines using DataDriftTrigger.
5. Add Evidently to pylint known third party.
6. Change ModelDownloader to embedding encoder utils. The downloader sets up and returns the model. The DataDriftTrigger owns the model. 

Future
1. Support multiple configurable Evidently metric. #416
2. Support Alibi-Detect. #414 
3. Support custom embedding encoder. #417
4. Support different windowing for detection data, e.g. compare with all previously trained data. #418
5. Common DataLoaderInfo #415


modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -189,25 +191,23 @@ def _handle_new_data(self, new_data: list[tuple[int, int, int]]) -> bool:
...
-
-        self._sw.start("selector_inform", overwrite=True)
-        selector_log = self.grpc.inform_selector(self.pipeline_id, batch)
-        self.pipeline_log["supervisor"]["selector_informs"].append(
-            {"total_selector_time": self._sw.stop(), "selector_log": selector_log}
-        )
+        if num_triggers == 0:
+            self._sw.start("selector_inform", overwrite=True)
+            selector_log = self.grpc.inform_selector(self.pipeline_id, batch)
+            self.pipeline_log["supervisor"]["selector_informs"].append(
```
MaxiBoether, line 206: Why only log this if there are no triggers?

modyn/supervisor/internal/triggers/trigger.py
```diff
@@ -1,20 +1,37 @@
...
 
 class Trigger(ABC):
     def __init__(self, trigger_config: dict) -> None:
         assert trigger_config is not None, "trigger_config cannot be None."
 
     @abstractmethod
-    def inform(self, new_data: list[tuple[int, int, int]]) -> list[int]:
-        """The supervisor informs the trigger about new data.
-        In case the concrete trigger implementation decides to trigger, we return a list of _indices into new_data_.
+    def init_trigger(self, pipeline_id: int, pipeline_config: dict, modyn_config: dict, base_dir: pathlib.Path) -> None:
```
MaxiBoether, line 11: I think we can make this a non-abstract method with just a `pass` implementation, since only the drift trigger uses this. Then we don't need the empty overrides in the other triggers.

modyn/supervisor/internal/triggers/trigger.py
```diff
@@ -1,20 +1,37 @@
...
+        The list might be empty or only contain a single element, which concrete Triggers need to respect.
 
              Parameters:
                      new_data (list[tuple[str, int, int]]): List of new data (keys, timestamps, labels). Can be empty.
 
              Returns:
                      triggering_indices (list[int]): List of all indices that trigger training
         """
+
+    @abstractmethod
```
MaxiBoether, line 30: Same comment as above - can probably be non abstract methods to avoid empty overrides

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
```diff
@@ -0,0 +1,106 @@
...
+from modyn.model_storage.internal.grpc.generated.model_storage_pb2 import FetchModelRequest, FetchModelResponse
+from modyn.model_storage.internal.grpc.generated.model_storage_pb2_grpc import ModelStorageStub
+from modyn.models.coreset_methods_support import CoresetSupportingModule
+from modyn.utils import dynamic_module_import
+from modyn.utils.utils import grpc_connection_established
+
+logger = logging.getLogger(__name__)
+
+
+# TODO(366) Unify similar code in trainer_server and evaluator. Create common.utils.ModelDownloader
```
MaxiBoether, line 23: 366 might be the wrong issue number. I guess it would be cleaner to add this downloader as a common utility beforehand, but we can also do it afterwards. But we should definitly do it :)

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
```diff
@@ -0,0 +1,106 @@
...
+        self._amp = amp
+        model_module = dynamic_module_import("modyn.models")
+        self.model_handler = getattr(model_module, model_class_name)
+        assert self.model_handler is not None
+
+        self.model_configuration_dict = json.loads(model_config)
+        self._model = self.model_handler(self.model_configuration_dict, self._device, amp)
+
+        assert self._model is not None
+        assert isinstance(self._model.model, CoresetSupportingModule)
```
MaxiBoether, line 87: Why this assert? I think not every model has to be a coreset supporting moudle, or?

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
```diff
@@ -0,0 +1,106 @@
...
+from modyn.model_storage.internal.grpc.generated.model_storage_pb2_grpc import ModelStorageStub
+from modyn.models.coreset_methods_support import CoresetSupportingModule
+from modyn.utils import dynamic_module_import
+from modyn.utils.utils import grpc_connection_established
+
+logger = logging.getLogger(__name__)
+
+
+# TODO(366) Unify similar code in trainer_server and evaluator. Create common.utils.ModelDownloader
+class ModelDownloader:
```
MaxiBoether, line 24: This is a ModelDownloader + a class representing the downloaded model. To make the object oriented abstraction a bit cleaner, could we instead just `return` a model at the end? From my understanding, the model downloader should download + instantiate, but not _own_ the model. That would be cleaner. Thank you!

modyn/supervisor/internal/triggers/trigger_datasets/dataloader_info.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+
+# TODO(366): Unify with similar classes in trainer_server and evaluator
```
MaxiBoether, line 4: I think that might be the wrong issue number. Might this be 275 instead?

modyn/supervisor/internal/triggers/trigger_datasets/online_trigger_dataset.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Generator, Optional
+
+from modyn.trainer_server.internal.dataset.online_dataset import OnlineDataset
+from torch.utils.data import IterableDataset
+
+logger = logging.getLogger(__name__)
+
+
+# TODO(#275): inherit common abstraction of dataset
+class OnlineTriggerDataset(OnlineDataset, IterableDataset):
```
MaxiBoether, line 12: Could you add a docstring here explaining the idea of this dataset (sampling with some probability) and why we need it)?

modyn/supervisor/internal/triggers/trigger_datasets/fixed_keys_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    instantiate_class,
+)
+from torch.utils.data import IterableDataset, get_worker_info
+from torchvision import transforms
+
+logger = logging.getLogger(__name__)
+
+
+# TODO(#275): inherit common abstraction of dataset
+class FixedKeysDataset(IterableDataset):
```
MaxiBoether, line 25: Can you add a docstring here what this does and why we need it? For the sampling dataset, I remember, but here some docs would be great. Thank you!

modyn/supervisor/internal/triggers/utils.py
```diff
@@ -0,0 +1,140 @@
...
+from modyn.supervisor.internal.triggers.trigger_datasets import DataLoaderInfo, FixedKeysDataset, OnlineTriggerDataset
+from torch.utils.data import DataLoader
+
+logger = logging.getLogger(__name__)
+
+
+def get_evidently_metrics(
+    column_mapping_name: str, trigger_config: Optional[dict] = None
+) -> list[EmbeddingsDriftMetric]:
+    """Evidently metric configurations follow exactly the four DriftMethods defined in embedding_drift_methods:
```
MaxiBoether, line 19: I think this needs an introductory sentence on what this function is doing (like instantiating an evidently metric)

MaxiBoether: Also, if you want to do a separate PR for your TODO 5 (which i think applies here, right), we might want to open an issue and link it here

modyn/supervisor/internal/triggers/utils.py
```diff
@@ -0,0 +1,140 @@
...
+                    all_embeddings = embeddings
+                else:
+                    all_embeddings = torch.cat((all_embeddings, embeddings), 0)
+
+    model_downloader._model.model.embedding_recorder.end_recording()
+
+    return all_embeddings
+
+
+def get_embeddings_evidently_format(
```
MaxiBoether, line 134: I think it would be cleaner to pass the tensor here and have a convert tensor to evidently embedding function instead

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+        current_keys, _, _ = zip(*self.data_cache[idx_start:idx_end])  # type: ignore
+        current_dataloader = prepare_trigger_dataloader_fixed_keys(
+            self.previous_trigger_id + 1,
+            self.dataloader_info,
+            current_keys,
+            sample_size=self.sample_size,
+        )
+
+        # Download previous model as embedding encoder
+        # TODO(366) Support custom model as embedding encoder
```
MaxiBoether, line 200: Might need a new issue here

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+        reference_embeddings_df = get_embeddings_evidently_format(self.model_downloader, reference_dataloader)
+        current_embeddings_df = get_embeddings_evidently_format(self.model_downloader, current_dataloader)
+
+        drift_detected = self.run_detection(reference_embeddings_df, current_embeddings_df)
+
+        return drift_detected
+
+    def inform(self, new_data: list[tuple[int, int, int]]) -> Generator[int, None, None]:
+        """
+        Use Generator here because this data drift trigger
```
MaxiBoether, line 215: I think this needs a longer docstring, sorry. Could you provide some longer documentation here, especially on the sliding window approach and what is happening? Right now, the docstring mostly concerns the return type

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+        untriggered_data_points = unvisited_data_points
+        # the sliding window of data points for detection
+        detection_idx_start = 0
+        detection_idx_end = 0
+
+        while unvisited_data_points >= self.detection_interval:
+            unvisited_data_points -= self.detection_interval
+            detection_idx_end += self.detection_interval
+            if detection_idx_end <= self.leftover_data_points:
+                continue
```
MaxiBoether, line 231: Should we move this at the end of the loop? Semenatically, we have not visited any new points here, have we? (Generally, I think some description of the algorithm as mentioned in the docstring comment would be super helpful)

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+            (x["result"]["drift_score"], x["result"]["method_name"], x["result"]["drift_detected"])
+            for x in result["metrics"]
+        ]
+        logger.info(
+            f"[DataDriftDetector][Prev Trigger {self.previous_trigger_id}][Dataset {self.dataloader_info.dataset_id}]"
+            + f"[Result] {result_print}"
+        )
+
+        with open(self.drift_dir / f"drift_{self.previous_trigger_id}.json", "w", encoding="utf-8") as f:
+            json.dump(result, f)
```
MaxiBoether, line 173: Is this for logging/debugging purposes?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+
+        return result["metrics"][0]["result"]["drift_detected"]
+
+    def detect_drift(self, idx_start: int, idx_end: int) -> bool:
```
MaxiBoether, line 177: Can you also add a docstring here with some more context (e.g., we always compare against the previous trigger in the current implementation, right?)?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -0,0 +1,260 @@
...
+        self._init_dataloader_info()
+        self._init_model_downloader()
+        self._create_dirs()
+
+    def run_detection(self, reference_embeddings_df: pd.DataFrame, current_embeddings_df: pd.DataFrame) -> bool:
+        assert self.dataloader_info is not None
+        assert self.drift_dir is not None
+
+        # Run Evidently detection
+        column_mapping = ColumnMapping(embeddings={self.evidently_column_mapping_name: reference_embeddings_df.columns})
```
MaxiBoether, line 155: Can you add a comment on what the column mapping name is?

MaxiBoether (changes requested):
Thanks Jingyi! I did not have big comments, mostly smaller code-style changes and documentation issues. I think it might still be sensible to merge the downloader model thing first instead of extracting it later (probably overall less work) but we can do it however you prefer. As mentioned on Element, let's discuss how you proceed with the other TODOs. We definitely need the unit tests in this PR before merging :)

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
jenny011: Now, only the data drift trigger uses the model downloader, so the models should be able to record embeddings. Will deal with it in the general model downloader PR.

modyn/supervisor/internal/triggers/trigger_datasets/dataloader_info.py
jenny011: 275 is about a common Dataset class, not DataLoaderInfo. I can create an issue about a common DataLoaderInfo.

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: Was used for experiments. It logs all drift detection metrics and results. Can remove it or make it debug-only.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
jenny011: This is the same as before supervisor refactorization. If there are triggers, the log is added inside _handle_new_data_batch().

modyn/supervisor/internal/triggers/trigger.py
jenny011: done

modyn/supervisor/internal/triggers/trigger.py
jenny011: done

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
jenny011: ModelDownloader Issue #433.

modyn/supervisor/internal/triggers/model_downloader/model_downloader.py
jenny011: The current model class does not store additional info as attributes. E.g. amp, device, device type. I have this model downloader class as a unified interface to access such additional info required to run the forward pass. A better way would probably be to implement an abstract model class that contains those additional info. I think it would better be dealt with in the ModelDownloader Issue #433. What do you think?

modyn/supervisor/internal/triggers/trigger_datasets/online_trigger_dataset.py
jenny011: done.

modyn/supervisor/internal/triggers/trigger_datasets/fixed_keys_dataset.py
jenny011: done.

modyn/supervisor/internal/triggers/utils.py
jenny011: Comment updated. Links to issues see the PR description.

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: Issue #417

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: done.

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: done

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: done

modyn/supervisor/internal/triggers/datadrifttrigger.py
jenny011: It must be handled at the top because it is including all the remaining untriggered samples from the previous new data batch. The remaining untriggered samples have been processed in the previous batch, so there's no need to process them again. Explained in the docstring.

modyn/supervisor/internal/triggers/utils.py
jenny011: Removed get_embeddings_evidently_format(). Added convert_tensor_to_df(tensor, column_name_prefix).

MaxiBoether:
Thanks for the work. Let me know when I should review again.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -305,17 +317,16 @@ def _handle_triggers_within_batch(self, batch: list[tuple[int, int, int]], trigg
                 )
             else:
                 logger.info(f"Skipping training on empty trigger {trigger_id}]")
-            self._persist_pipeline_log()
 
             self.num_triggers = self.num_triggers + 1
             if self.maximum_triggers is not None and self.num_triggers >= self.maximum_triggers:
-                break
+                return num_triggers
```
XianzheMa, line 323: just nitpicking: I think `num_triggers` is not needed as we can replace every occurrence of it with `len(triggering_idx_list)` right?

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -305,17 +317,16 @@ def _handle_triggers_within_batch(self, batch: list[tuple[int, int, int]], trigg
                 )
             else:
                 logger.info(f"Skipping training on empty trigger {trigger_id}]")
-            self._persist_pipeline_log()
```
XianzheMa, line 308: why do we remove this?
I'd like to keep it because this position is immediately after a trigger training / evaluation. We can persist the training / evaluation data so long as we have it.

modyn/supervisor/internal/triggers/amounttrigger.py
```diff
@@ -14,12 +16,12 @@ def __init__(self, trigger_config: dict):
 
         super().__init__(trigger_config)
 
-    def inform(self, new_data: list[tuple[int, int, int]]) -> list[int]:
+    def inform(self, new_data: list[tuple[int, int, int]]) -> Generator[int, None, None]:
         assert self.remaining_data_points < self.data_points_for_trigger, "Inconsistent remaining datapoints"
```
XianzheMa, line 20: Why do we make it a `Generator[int, None, None]` instead of `Generator[int]`?

XianzheMa, line 20: Why do we make it a `Generator[int, None, None]` instead of `Generator[int]`?

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa, line 323: `num_triggers` is not needed as we can replace every occurrence of it with `len(triggering_idx_list)`.

It's not about nitpicking: better to remove it to not confuse others with `self.num_triggers`, which is a different number

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa, line 308: why do we remove this?
I'd like to keep it because this position is immediately after a trigger training / evaluation. We can persist the training / evaluation data so long as we have it.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -177,15 +180,15 @@ def _handle_new_data(self, new_data: list[tuple[int, int, int]]) -> bool:
 
     def _handle_new_data_batch(self, batch: list[tuple[int, int, int]]) -> bool:
         self._sw.start("trigger_inform", overwrite=True)
-        triggering_indices = self.trigger.inform(batch)
-        num_triggers = len(triggering_indices)
-        self.pipeline_log["supervisor"]["num_triggers"] += len(triggering_indices)
+        triggering_indices: Generator[int, None, None] = self.trigger.inform(batch)
```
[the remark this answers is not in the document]
XianzheMa: Thanks for keeping me in the loop. It looks good to me except for small changes!

XianzheMa (approved):
LGTM and can you address the small changes I requested before merging! Thanks!

modyn/supervisor/internal/triggers/trigger.py
jenny011: oops, missed it. Now it's fixed.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
jenny011: done.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
jenny011: It's probably lost in merging. Added back.

PR #410, opened 2024-05-01
Add option to cap training by the number of samples instead of number of epochs

XianzheMa:
Per the title, sometimes we want to equalize the number of training steps across different policies, this PR adds an option in the pipeline config file to limit the number of samples to train on.

The PR is part of the larger PR https://github.com/eth-easl/modyn/pull/407. I single it out for easier review.


environment.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
   - pip
   - tqdm
   - conda-forge::enlighten
-  - protobuf
```
XianzheMa, line 21: failure in dockernized unittest because of unspecified protobuf version:
image

robinholzi: Wouldn't it be better to pin the protobuf version rather than removing the dependency completely?

XianzheMa: I did intend to pin a version but made a mistake... definitely we can't remove protobuf

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -488,6 +500,9 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
             self._log["epochs"][epoch]["OnBatchEnd"] = stopw.measurements.get("OnBatchEnd", 0)
 
             self._persist_pipeline_log()
+            if 0 < self.num_samples_to_pass <= self._num_samples:
+                self._info("reached the threshold of samples to pass; break out of epoch loop to stop training")
+                break
```
MaxiBoether, line 505: Why do we need to check this twice?

MaxiBoether (approved):
Just some nitpicky comments, otherwise LGTM! (But please do not merge with removing protobuf)

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
```diff
@@ -107,13 +107,15 @@ def teardown():
...
+    if pipeline_config is None:
+        pipeline_config = get_minimal_pipeline_config()
     pipeline_executor = PipelineExecutor(
         START_TIMESTAMP,
         PIPELINE_ID,
         get_minimal_system_config(),
-        get_minimal_pipeline_config(),
-        EVALUATION_DIRECTORY,
+        pipeline_config,
+        str(EVALUATION_DIRECTORY),
```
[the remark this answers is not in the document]
XianzheMa: because this function accepts a string but `EVALUATION_DIRECTORY` is a `Path` object. Previously there is an implicit type conversion and my IDE complains about it.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -452,6 +461,9 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
                         self._model.model, self._optimizers, batch_number, sample_ids, data, target, output, loss
                     )
                 stopw.stop()
+                if 0 < self.num_samples_to_pass <= self._num_samples:
```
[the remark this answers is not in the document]
XianzheMa: When `self.num_samples_to_pass == 0` we don't even limit training by this field, i.e. we use the number of epochs. So in that case we shouldn't execute the logic under if.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
XianzheMa: We have two loops here. One for each epoch and one for each batch in an epoch. When we reach the limit we need to break out of both loops.

environment.yml
XianzheMa: I revert this change because of the dedicated PR https://github.com/eth-easl/modyn/pull/412

PR #411, opened 2024-05-01
chore: Support pre-build grpc artefacts for CI jobs

robinholzi:
# Motivation

I want to speed up the CI. I want to build a workflow that periodically compiles all grpc binaries via a CI matrix. The build artifacts can then be used for linking grpc in the other CI workflows.

However, to build up this CI workflow I will need a couple of iterations. ~~Therefore I want to merge this dummy workflow so I can have the workflow triggered on `pull_request`.~~

** Update: Seems as if repo settings allow CI workflow runs on freshly introduced workflows.

# Open questions

- do we need Debug builds when compiling grpc independently?

# Bugfixes

- Allows building modyn from pre-compiled binaries locally by following the adjusted documentation. CMake config previously led to linking errors.
- C++ coverage tests were run on main branch ignoring all changes in the PR (see checkout action), change needed to allow DMODYN_TRY_LOCAL_GRPC=ON (with cmake fixes)

### DIFF
--- a/modyn/storage/src/CMakeLists.txt
+++ b/modyn/storage/src/CMakeLists.txt
@@ -59,7 +59,7 @@ protobuf_generate(
     OUT_VAR PROTO_GENERATED_FILES
     LANGUAGE grpc
     GENERATE_EXTENSIONS .grpc.pb.h .grpc.pb.cc
-    PLUGIN "protoc-gen-grpc=\$<TARGET_FILE:grpc_cpp_plugin>"
+    PLUGIN "protoc-gen-grpc=\$<TARGET_FILE:${MODYN_GRPC_CPP_PLUGIN}>"
     # PLUGIN_OPTIONS "generate_mock_code=true"
     IMPORT_DIRS ../../protos
     PROTOC_OUT_DIR "${PROTO_BINARY_DIR}")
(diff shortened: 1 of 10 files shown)


modyn/storage/src/CMakeLists.txt
```diff
@@ -68,51 +68,10 @@ message(STATUS "Proto generated files in ${PROTO_BINARY_DIR}: ${PROTO_GENERATED_
...
-        absl::strings
-        absl::synchronization
-        absl::time
-        absl::type_traits
-        absl::utility
-        absl::variant
-        absl::random_random
-    )
-
-    target_link_libraries(modyn-storage-proto PUBLIC protobuf::libprotobuf grpc_unsecure gpr libaddress_sorting.a libupb.a libcares.a libz.a utf8_range ${protobuf_ABSL_USED_TARGETS})
```
MaxiBoether, line 113: Just a note: At the time, this all was somewhat necessary to make it work with local gRPC because otherwise we were running into tons of linking issues. It could be that something changed about abseil/grpc, but there was a reason that we did all of this shenanigans :D

MaxiBoether:
If you can actually cache the gRPC binaries, that would be amazing. Thanks so much for the effort here.

modyn/storage/src/CMakeLists.txt
robinholzi: I ran the full build pipeline from scratch using both local and pre-built grpc. I didn't encounter any linking errors so I assumed this wouldn't be needed (also in documentations online this isn't needed)

robinholzi:
Closing for now as I don't intend to spend more time on this.

PR #419, opened 2024-05-05
feat: Add pydantic config models

robinholzi:
To improve typing within the codebase w.r.t. to the config objects, I transitioned the `jsonschema`s to pydantic models.

Those model will make the supervisor a lot more concise as all the validation logic will be stripped out and arguments will have proper typing. As a result it will be a lot easier to have uniform benchmarking across several pipeline steps without significant code duplication (needed for policy evaluation)


MaxiBoether:
> This PR revealed several (>100) mypy typing issues that weren't caught (apparently) because of missing `__init__.py` files. I propose addressing those in a separate PR before continuing here.

Oh, wow. So would you suggest fixing those in a separate PR that gets merged into this PR here or into main first?

robinholzi:
> > This PR revealed several (>100) mypy typing issues that weren't caught (apparently) because of missing `__init__.py` files. I propose addressing those in a separate PR before continuing here.
> 
> Oh, wow. So would you suggest fixing those in a separate PR that gets merged into this PR here or into main first?

#420 should be merged into main first :) I will fix the broken CI there tonight 👍

PR #422, opened 2024-05-06
Xianzhe ma/modern eval strategy

XianzheMa:


modyn/supervisor/internal/evaluation_result_writer/__init__.py
```diff
@@ -6,7 +6,6 @@
 
 from .abstract_evaluation_result_writer import AbstractEvaluationResultWriter  # noqa: F401
 from .json_result_writer import JsonResultWriter  # noqa: F401
-from .log_result_writer import LogResultWriter  # noqa: F401
 from .tensorboard_result_writer import TensorboardResultWriter  # noqa: F401
```
XianzheMa, line 9: `LogResultWriter` is introduced in the [previous matrix eval PR](https://github.com/eth-easl/modyn/pull/334/files#diff-f444432c981ee0f1b1e27538f2aceb5d7ceb344df60c2791f8d118e17e7b7ee7) which is nothing different from `JsonResultWriter`.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -124,6 +130,65 @@ def _persist_pipeline_log(self) -> None:
...
+            else:
+                logger.info(f"Evaluation started for model {model_id} on split {previous_split} to {current_split}.")
+                reporter = EvaluationStatusReporter(
+                    self.eval_status_queue, response.evaluation_id, eval_dataset_id, response.dataset_size
+                )
+                reporter.create_tracker()
+                evaluation = {response.evaluation_id: reporter}
+                assert self.current_training_id is not None, "Training ID not set."
+                self.grpc.wait_for_evaluation_completion(self.current_training_id, evaluation)
+
```
XianzheMa, line 184: This json writer is actually not needed as we store the results in the pipeline log rather than in a separate eval file. We keep it for the moment to re-use its code to issue the `EvaluationResultRequest` and parse the `EvaluationResultReponse`.

We should have a further PR to clean up the logic here.

modyn/supervisor/internal/supervisor.py
```diff
@@ -81,27 +76,22 @@ def init_cluster_connection(self) -> None:
...
+            is_valid = False
 
-            dataloader_workers = dataset["dataloader_workers"]
-            if dataloader_workers < 1:
-                logger.error(f"Invalid dataloader worker amount: {dataloader_workers}.")
-                is_valid = False
+        dataloader_workers = dataset["dataloader_workers"]
+        if dataloader_workers < 1:
+            logger.error(f"Invalid dataloader worker amount: {dataloader_workers}.")
+            is_valid = False
```
XianzheMa, line 94: We change in the pipeline schema the list of evaluation datasets to a single dataset. Hence here we change the validation logic.

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
```diff
@@ -583,48 +590,6 @@ def test__run_training_set_num_samples_to_pass(
...
-    test_store_trained_model.assert_called_once()
-
-    test_start_evaluation.assert_called_once_with(101, evaluation_pipeline_config)
-    test_wait_for_evaluation_completion.assert_called_once_with(1337, evaluations)
-    test_store_evaluation_results.assert_called_once()
-    assert len(test_store_evaluation_results.call_args[0][0]) == 1
-    result_writer: AbstractEvaluationResultWriter = test_store_evaluation_results.call_args[0][0][0]
-    assert result_writer.eval_directory == EVALUATION_DIRECTORY
-    assert result_writer.pipeline_id == 42
-    assert result_writer.trigger_id == 21
```
XianzheMa, line 625: remove old unit test

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -124,6 +130,64 @@ def _persist_pipeline_log(self) -> None:
...
+                self._sw.stop("evaluate_model")
+            else:
+                logger.info(f"Evaluation started for model {model_id} on split {interval_start} to {interval_end}.")
+                reporter = EvaluationStatusReporter(
+                    self.eval_status_queue, response.evaluation_id, eval_dataset_id, response.dataset_size
+                )
+                reporter.create_tracker()
+                evaluation = {response.evaluation_id: reporter}
+                assert self.current_training_id is not None, "Training ID not set."
+                self.grpc.wait_for_evaluation_completion(self.current_training_id, evaluation)
```
XianzheMa, line 183: What is the use of JsonResultWriter / AbstractResultWriter? We seem to always inspect information from the pipeline log. Can I remove all the result writers?

This json writer is actually not needed as we store the results in the pipeline log rather than in a separate eval file. We keep it for the moment to re-use its code to issue the EvaluationResultRequest and parse the EvaluationResultReponse.

We should have a further PR to clean up the logic here.

modyn/config/examples/example-pipeline.yaml
```diff
@@ -69,30 +69,36 @@ trigger:
...
+          def evaluation_transformer_function(model_output: torch.Tensor) -> torch.Tensor:
+            return torch.argmax(model_output, dim=-1)
+      - name: "F1-score"
+        config:
+          num_classes: 10
+          average: "macro"
+        evaluation_transformer_function: |
+          import torch
+          def evaluation_transformer_function(model_output: torch.Tensor) -> torch.Tensor:
+            return torch.argmax(model_output, dim=-1)
```
XianzheMa, line 104: we don't need multiple evaluation datasets now.

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -524,37 +523,14 @@ def seed_selector(self, seed: int) -> None:
...
-            response: EvaluateModelResponse = self.evaluator.evaluate_model(req)
-
-            if not response.evaluation_started:
-                logger.error(f"Starting evaluation for dataset {dataset_id} did go wrong: {response}.")
-            else:
-                evaluation_id = response.evaluation_id
-                logger.info(f"Started evaluation {evaluation_id} on dataset {dataset_id}.")
-                evaluations[evaluation_id] = EvaluationStatusReporter(
-                    self.eval_status_queue, evaluation_id, dataset_id, response.dataset_size
-                )
```
XianzheMa, line 551: remove old eval implementation.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -230,22 +294,7 @@ def _run_training(self, trigger_id: int) -> None:
...
-            # TODO(#300) Add evaluator to pipeline log
-            evaluations = self.grpc.start_evaluation(model_id, self.pipeline_config)
-            self.grpc.wait_for_evaluation_completion(self.current_training_id, evaluations)
-
-            self._update_pipeline_stage_and_enqueue_msg(
-                PipelineStage.STORE_EVALUATION_RESULTS, MsgType.ID, id_submsg(IdType.TRIGGER, trigger_id)
-            )
-            writer_names: set[str] = set(self.pipeline_config["evaluation"]["result_writers"])
-            writers = [self._init_evaluation_writer(name, trigger_id) for name in writer_names]
-            self.grpc.store_evaluation_results(writers, evaluations)
```
XianzheMa, line 248: make evaluation logic a separate function rather than included in training.

integrationtests/model_storage/integrationtest_model_storage.py
```diff
@@ -96,13 +98,13 @@ def insert_triggers_into_database(
...
 
-        trigger_parent = Trigger(trigger_id=0, pipeline_id=pipeline_id)
-        trigger_child = Trigger(trigger_id=1, pipeline_id=pipeline_id)
+        trigger_parent = Trigger(trigger_id=parrent_trigger_id, pipeline_id=pipeline_id)
+        trigger_child = Trigger(trigger_id=child_trigger_id, pipeline_id=pipeline_id)
         database.session.add(trigger_parent)
         database.session.add(trigger_child)
         database.session.commit()
 
-        return pipeline_id, trigger_parent.trigger_id, trigger_child.trigger_id
```
XianzheMa, line 105: return statement shouldn't be in the `with` context because otherwise `__exit__()` is not executed.

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -16,15 +16,15 @@
 from modyn.supervisor.internal.grpc.generated.supervisor_pb2 import StartPipelineRequest
 from modyn.supervisor.internal.grpc.generated.supervisor_pb2_grpc import SupervisorStub
 
-POLL_TIMEOUT = 1
+POLL_INTERVAL = 1
```
XianzheMa, line 19: better naming

integrationtests/utils.py
```diff
@@ -278,16 +276,15 @@ def create_random_image(self) -> Image:
     def add_image_to_dataset(self, image: Image, name: str) -> None:
         image.save(self.dataset_path / name)
 
-    def add_images_to_dataset(self, start_number: int, end_number: int, images_added: list[bytes]) -> None:
+    def add_images_to_dataset(self, start_number: int, end_number: int) -> None:
```
XianzheMa, line 279: remove a useless argument

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -9,15 +9,21 @@
...
 from modyn.supervisor.internal.grpc.enums import CounterAction, IdType, MsgType, PipelineStage
 from modyn.supervisor.internal.grpc.template_msg import counter_submsg, dataset_submsg, id_submsg, pipeline_stage_msg
 from modyn.supervisor.internal.grpc_handler import GRPCHandler
 from modyn.supervisor.internal.triggers import Trigger
+from modyn.supervisor.internal.utils import EvaluationStatusReporter
 from modyn.utils import dynamic_module_import
 
 logger = logging.getLogger(__name__)
 EXCEPTION_EXITCODE = 8
+EVALUATION_RESULTS = "evaluation_results"
```
MaxiBoether, line 26: we need a different name for this global. Is this a directory name? Something else? Some suffix might be needed

MaxiBoether: ok I see it's a key - maybe `EVAL_RESULTS_KEY` or `EVAL_RESULTS_IDENTIFIER` or something

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
MaxiBoether: Hm, we can clean this up. I think where the `LogResultWriter` came from  was "just log to the pipeline log". IIRC we were implementing the pipeline log and the eval result writer somewhat in parallel so there might be historic reasons. We can either decide to have everything in the pipeline log or all eval results separately

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -127,6 +133,64 @@ def _persist_pipeline_log(self) -> None:
...
+                self._sw.stop("evaluate_model")
+            else:
+                logger.info(f"Evaluation started for model {model_id} on split {interval_start} to {interval_end}.")
+                reporter = EvaluationStatusReporter(
+                    self.eval_status_queue, response.evaluation_id, eval_dataset_id, response.dataset_size
+                )
+                reporter.create_tracker()
+                evaluation = {response.evaluation_id: reporter}
+                assert self.current_training_id is not None, "Training ID not set."
+                self.grpc.wait_for_evaluation_completion(self.current_training_id, evaluation)
```
MaxiBoether, line 186: So right now we cannot do parallel evaluation, correct? We always wait for one evaluation to start before starting the next one?

One way to do it would be to implement parallel evaluation at the evaluator itself, where it has a queue of incoming requests, executing as much in parallel as possible, and the pipeline executor will just submit a bunch of requests and wait for all of them to finish.

MaxiBoether: anyways, not relevant for this pr but let's keep this in mind for the future

MaxiBoether (approved):
LGTM generally! I had some general questions and especially I am not sure if we truly need to remove support for multiple eval datasets, but generally this looks clean. Thanks so much!

integrationtests/model_storage/integrationtest_model_storage.py
XianzheMa: Ah I see thanks. but this is not a good practice anyway

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa: I think still `EVALUATION_RESULTS` is better because under this entry, the sub-object includes everything about evaluations -- different intervals, different metrics, different eval datasets (I will restore the setting of multi eval datasets).

modyn/config/schema/pipeline.py
```diff
@@ -386,6 +386,17 @@ class TriggerConfig(BaseModel):
 # ---------------------------------------------------- EVALUATION ---------------------------------------------------- #
 
 
+EvalStrategyName = Literal["MatrixEvalStrategy", "OffsetEvalStrategy"]
+
+
+class EvalStrategy(BaseModel):
+    name: EvalStrategyName = Field(description="The name of the evaluation strategy that should be used.")
+    config: Dict[str, Any] = Field(
```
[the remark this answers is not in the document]
MaxiBoether: @robinholzi just merged a PR taking care about this in many different places (avoid passing around raw dictionaries), so let's try to keep this approach if we can :D I think these will just be 2 tiny classes instead of dicts, so it should not be too much bloat

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa: Yes now it's one after another. And we should measure if we can get speed up by executing multiple eval jobs in parallel (and I doubt it), because as far as I know, if we don't enable [Nvidia Multi-Process Service](https://docs.nvidia.com/deploy/mps/index.html), it's still executed one after another on GPU, or in best case time-sliced ([source](https://stackoverflow.com/questions/31643570/running-more-than-one-cuda-applications-on-one-gpu)).

For this PR let's do it one-after-another.

modyn/config/examples/example-pipeline.yaml
XianzheMa: You are right! I have restored this part.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
MaxiBoether: hmm, okay, I will also chat with Foteini about this. Thanks for the pointers!

modyn/config/schema/pipeline.py
```diff
@@ -128,14 +131,10 @@ class MultiDownsamplingConfig(BaseModel):
...
-    def validate_downsampling_thresholds(
-        cls,
-        data: "MultiDownsamplingConfig",
-    ) -> "MultiDownsamplingConfig":
-        if len(data.downsampling_thresholds) != len(data.downsampling_list) - 1:
+    def validate_downsampling_thresholds(self) -> Self:
+        if len(self.downsampling_thresholds) != len(self.downsampling_list) - 1:
             raise ValueError("The downsampling_thresholds list should have one less item than the downsampling_list.")
-        return data
+        return self
```
XianzheMa, line 138: @robinholzi We should add tests against non-trivial validations because I found some not-working code...
Here, when `model_validator` is used with `after` mode, it is passed with an instance of the class.
See https://docs.pydantic.dev/latest/concepts/validators/#model-validators

modyn/config/schema/pipeline.py
```diff
@@ -364,17 +362,16 @@ def validate_gpus(cls, value: int) -> int:
         return value
 
     @model_validator(mode="after")
-    @classmethod
-    def validate_num_samples_to_pass(cls, data: "TrainingConfig") -> "TrainingConfig":
-        if data.initial_model == "pretrained":
-            if not data.use_previous_model:
-                raise ValidationError(
+    def validate_pretrained(self) -> Self:
+        if self.initial_model == "pretrained":
```
XianzheMa, line 366: the previous name is confusing as this method is not acted on the field `num_samples_to_pass`

modyn/config/schema/pipeline.py
```diff
@@ -480,16 +534,6 @@ class EvaluationConfig(BaseModel):
         min_length=1,
     )
 
-    # Additional validation
-
-    @model_validator(mode="after")
```
robinholzi, line 485: Thanks for improving the pipeline model! 🥇 

Why do we want to drop the `dataset_id` uniqueness constraint though?

modyn/config/schema/pipeline.py
```diff
@@ -364,17 +362,16 @@ def validate_gpus(cls, value: int) -> int:
...
+            if not self.use_previous_model:
+                raise ValueError(
                     "Cannot have use_previous_model == False and use a pretrained initial model."
                     "Initial model would get lost after first trigger."
                 )
-            if not data.initial_model_id:
-                raise ValidationError("Initial model set to pretrained, but no initial_model_id given")
-        return data
+            if not self.initial_model_id:
+                raise ValueError("Initial model set to pretrained, but no initial_model_id given")
```
XianzheMa, line 373: Also `ValidationError` should not used, see https://docs.pydantic.dev/latest/concepts/validators/#handling-errors-in-validators. Actually `ValidationError` even does not have a constructor with a string.

modyn/config/schema/pipeline.py
robinholzi: Personally, I wouldn't add tests for the validation as it will fail loudly when sth. is wrong in almost all cases. Then one can simply adjust the model (e.g. if a field type is off). The same applies to the usage of @field_validator and @model_validator. But I agree, if there's complex logic inside the custom validation functions it makes sense to write a small unit test though

modyn/tests/config/schema/test_pipeline.py
```diff
@@ -0,0 +1,92 @@
...
+            {
+                "offsets": ["10d10s"],
+            }
+        )
+
+    TypeAdapter(OffsetEvalStrategyConfig).validate_python(
+        {
+            "offsets": [OffsetEvalStrategy.INFINITY, OffsetEvalStrategy.NEGATIVE_INFINITY, "10s", "10d", "-10s", "0s"],
+        }
+    )
```
XianzheMa, line 92: We can consider splitting the `pipeline.py` into multiple files if a combined one is too long. But for this PR it will do the job.

modyn/config/schema/pipeline.py
robinholzi: Good catch! :) That might have been a copy & paste issue

modyn/config/schema/pipeline.py
XianzheMa: I didn't find a working solution for the moment. The `model_validator` does not work as I checked, but I cannot find a correct one.

modyn/config/schema/pipeline.py
XianzheMa: I didn't get your argument? You mean we cannot find out which file/test function fails in the `pytest` output because there are too many errors? I think the summary is easy to analyze.

I believe it's worth adding unit tests to these custom validation functions, otherwise we might make mistakes without being aware of it (for example here)

PR #423, opened 2024-05-07
feat: Add pydantic config models

robinholzi:
To improve typing within the codebase w.r.t. to the config objects, I transitioned the `jsonschema`s to pydantic models.

Those model will make the supervisor a lot more concise as all the validation logic will be stripped out and arguments will have proper typing. As a result it will be a lot easier to have uniform benchmarking across several pipeline steps without significant code duplication (needed for policy evaluation)


modyn/storage/src/CMakeLists.txt
```diff
@@ -135,4 +135,4 @@ target_compile_definitions(modyn-storage-library PUBLIC ${MODYN_COMPILE_DEFINITI
 set(INCLUDE_EXPR "$<TARGET_PROPERTY:modyn-storage-library,INCLUDE_DIRECTORIES>")
 set(INCLUDE_FILTER "$<FILTER:${INCLUDE_EXPR},INCLUDE,.*/.*>")
 set(INCLUDE_JOINED "$<JOIN:${INCLUDE_FILTER},,>")
-target_compile_definitions(modyn-storage-library PRIVATE "INCLUDE_DIRECTORIES=\"${INCLUDE_JOINED}\"")
```
MaxiBoether, line 138: let's not touch this file with this commit :)

MaxiBoether (approved):
I validated the pipeline config, client config and an excerpt of the modyn config. Since the tests also run through, I think this is good to go. Thank you!

I think your IDE might be using different settings from our flake/black config since this PR introduces many (unrelated) linebreaks (but to me, that's ok, can merge them) :-)

modyn/storage/src/CMakeLists.txt
robinholzi: Ok, I reverted these changes. Going forward, would you prefer file endings with or without newlines?

XianzheMa:
Großartig! Thank you for this pain saver!

PR #425, opened 2024-05-07
Make the reason for evaluation failure explicit

XianzheMa:
This is the last irrelevant PR I singled out from https://github.com/eth-easl/modyn/pull/422.

One evaluation request can fail to start due to many reasons. Previously the reasons are not explicitly reflected. This has two downsides:
1. In unit test when we want to test that "condition A" must be satisfied to start evaluation, we can only assert the evaluation is not started but the failure could be because of "condition B", i.e. we have no way to verify the failure is because of "condition A".
2. On supervisor side, we don't know why the evaluation fails.

This PR makes the failure reason explicit.


modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -16,6 +16,7 @@
     EvaluateModelRequest,
     EvaluateModelResponse,
     EvaluationData,
+    EvaluationNotStartReason,
```
MaxiBoether, line 19: Can we rename this to `EvaluationNotStartedReason`?

MaxiBoether: or something smaller, like `EvaluationCancelReason` or `EvaluationAbortReason` or `EvaluationFailureReason` or `EvaluationFailureReason `

That would be better to read.

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -100,12 +101,16 @@ def evaluate_model(self, request: EvaluateModelRequest, context: grpc.ServicerCo
 
             if not trained_model:
                 logger.error(f"Trained model {request.model_id} does not exist!")
-                return EvaluateModelResponse(evaluation_started=False)
+                return EvaluateModelResponse(
+                    evaluation_started=False, not_start_reason=EvaluationNotStartReason.MODEL_NOT_EXIST_IN_METADATA
```
MaxiBoether, line 105: Similar here, depending on what name you decide for, can we rename the argument to `not_started_reason` or `failure_reason` or `cancel_reason` or sth like this please? Thanks

MaxiBoether (approved):
Thanks, my only comment is on the naming which I would either like to grammatically change or shorten a bit

XianzheMa:
> Thanks, my only comment is on the naming which I would either like to grammatically change or shorten a bit

thanks! I have renamed them to `aborted_reason`/`AbortedReason`!

MaxiBoether:
> > Thanks, my only comment is on the naming which I would either like to grammatically change or shorten a bit
> 
> thanks! I have renamed them to `aborted_reason`/`AbortedReason`!

We should go for either `AbortReason` or `EvalAbortedReason` - `AbortedReason` is a bit inappropriate. Please fix that 😅

XianzheMa:
> > > Thanks, my only comment is on the naming which I would either like to grammatically change or shorten a bit
> > 
> > 
> > thanks! I have renamed them to `aborted_reason`/`AbortedReason`!
> 
> We should go for either `AbortReason` or `EvalAbortedReason` - `AbortedReason` is a bit inappropriate. Please fix that 😅

Yes sorry for my brevity. It is `EvaluationAbortedReason`

PR #428, opened 2024-05-09
Make `start_timestamp` and `end_timestamp` in evaluation gRPC requests optional

XianzheMa:
This PR makes two parameters I introduced in a previous PR https://github.com/eth-easl/modyn/pull/388 optional. Previously, I made it so that the default value "0" means no limit. I.e. a request with both `start_timestamp` and `end_timestamp` as `0` means no time restriction on the dataset.

I realized that this semantic is clumsy as sometimes we need "0" to mean the real timestamp "0". And now there is conflict. This PR makes `start_timestamp` and `end_timestamp` optional, and when they are optional, it means there is no limit. When they are set with a value, it means a limit with the numerical value.

### DIFF
--- a/modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
+++ b/modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
@@ -459,19 +459,16 @@ TEST_F(StorageServiceImplTest, TestGetNumberOfSamplesInDatasetWithRange) {
   ASSERT_NO_THROW(result = StorageServiceImpl::get_number_of_samples_in_dataset_with_range(1, session));
   ASSERT_EQ(result, 2);
 
-  ASSERT_NO_THROW(result = StorageServiceImpl::get_number_of_samples_in_dataset_with_range(1, session, 0, 0));
-  ASSERT_EQ(result, 2);
-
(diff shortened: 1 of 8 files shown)


modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
```diff
@@ -459,19 +459,16 @@ TEST_F(StorageServiceImplTest, TestGetNumberOfSamplesInDatasetWithRange) {
   ASSERT_NO_THROW(result = StorageServiceImpl::get_number_of_samples_in_dataset_with_range(1, session));
   ASSERT_EQ(result, 2);
 
-  ASSERT_NO_THROW(result = StorageServiceImpl::get_number_of_samples_in_dataset_with_range(1, session, 0, 0));
```
MaxiBoether, line 462: Do we have a test now for having the default vs having explicit 0? If not, could you extend this test to test not set vs 0 as well?

MaxiBoether (approved):
Thanks. I think we might need to extend the tests to test for not set vs 0, but I might have missed it (didnt sleep a lot haha)

modyn/tests/storage/internal/grpc/storage_service_impl_test.cpp
XianzheMa: sure!

PR #429, opened 2024-05-09
Record the time span of a trigger

XianzheMa:
This PR makes a way to record the first and last timestamp of the data within a trigger.

This information is crucial for flexible evaluation: the first and last timestamps of the data in a trigger will be passed to the evaluation strategy introduced in https://github.com/eth-easl/modyn/pull/426, to decide what time slices of the evaluation dataset we want to evaluate against.

This PR is part of https://github.com/eth-easl/modyn/pull/422.

### DIFF
--- a/modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
+++ b/modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
@@ -79,8 +79,7 @@ class PipelineExecutor:
         self.num_triggers = 0
         self.current_training_id: Optional[int] = None
         self.triggers: list[int] = []
-        self.last_sample_in_trigger_timestamp: Optional[int] = None
-        self.first_sample_in_trigger_timestamp: Optional[int] = None
+        self.remaining_data_range: Optional[tuple[int, int]] = None
 
     def _update_pipeline_stage_and_enqueue_msg(
         self, stage: PipelineStage, msg_type: MsgType, submsg: Optional[dict[str, Any]] = None, log: bool = False
@@ -154,7 +153,7 @@ class PipelineExecutor:
         )
 
         for i in range(0, new_data_len, self._selector_batch_size):
-            batch = new_data[i : i + self._selector_batch_size]
+            batch = new_data[i: i + self._selector_batch_size]
             batch_size = self._selector_batch_size if i + self._selector_batch_size < new_data_len else new_data_len - i
             self._update_pipeline_stage_and_enqueue_msg(
                 PipelineStage.HANDLE_NEW_DATA,
@@ -262,7 +261,7 @@ class PipelineExecutor:
 
         for i, triggering_idx in enumerate(triggering_indices):
             self._update_pipeline_stage_and_enqueue_msg(PipelineStage.INFORM_SELECTOR_AND_TRIGGER, MsgType.GENERAL)
-            triggering_data = batch[previous_trigger_idx : triggering_idx + 1]
+            triggering_data = batch[previous_trigger_idx: triggering_idx + 1]
             previous_trigger_idx = triggering_idx + 1
 
             # This call informs the selector about the data until (and including)
@@ -279,19 +278,24 @@ class PipelineExecutor:
 
             num_samples_in_trigger = self.grpc.get_number_of_samples(self.pipeline_id, trigger_id)
             if num_samples_in_trigger > 0:
-                if len(triggering_data) > 0:
+                if trigger_id > 0:
+                    # since num_samples_in_trigger is not 0, we can be sure that triggering_data is not empty
+                    first_timestamp = triggering_data[0][1]
                     last_timestamp = triggering_data[-1][1]
                 else:
-                    assert self.last_sample_in_trigger_timestamp is not None
-                    last_timestamp = self.last_sample_in_trigger_timestamp
+                    # now it's the first trigger in this batch. Triggering_data can be empty.
+                    # when it is indeed empty, then there is remaining data in the last batch.
+                    assert len(triggering_data) > 0 or self.remaining_data_range is not None
+                    if len(triggering_data) > 0:
+                        last_timestamp = triggering_data[-1][1]
+                    else:
+                        last_timestamp = self.remaining_data_range[1]
 
-                if previous_trigger_idx == triggering_idx + 1 and self.first_sample_in_trigger_timestamp is not None:
-                    # the first trigger in this batch; whether the first sample in the trigger is
-                    # the first sample in triggering_data depends on if during the last call
-                    # there is non-empty remaining data
-                    first_timestamp = self.first_sample_in_trigger_timestamp
-                else:
-                    first_timestamp = triggering_data[0][1]
+                    if self.remaining_data_range is not None:
+                        first_timestamp = self.remaining_data_range[0]
+                    else:
+                        first_timestamp = triggering_data[0][1]
+
                 self.pipeline_log["supervisor"]["triggers"][trigger_id]["first_timestamp"] = first_timestamp
                 self.pipeline_log["supervisor"]["triggers"][trigger_id]["last_timestamp"] = last_timestamp
                 self._run_training(trigger_id)  # Blocks until training is done.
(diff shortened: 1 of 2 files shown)


modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -277,13 +279,33 @@ def _handle_triggers_within_batch(self, batch: list[tuple[int, int, int]], trigg
 
             num_samples_in_trigger = self.grpc.get_number_of_samples(self.pipeline_id, trigger_id)
             if num_samples_in_trigger > 0:
+                if i > 0:
+                    # since num_samples_in_trigger is not 0, we are sure that triggering_data is not empty
+                    first_timestamp = triggering_data[0][1]
+                    last_timestamp = triggering_data[-1][1]
+                else:
+                    # now it is the first trigger in this batch. Triggering_data can be empty.
```
MaxiBoether, line 287: How can `triggering_data` be empty if `num_samples_in_trigger > 0`? That confuses me a bit. And why can this only happen in the first trigger?

MaxiBoether: so basically i am a bit confused why it's not always the code path for i > 0 and why this here can happen

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -277,13 +279,33 @@ def _handle_triggers_within_batch(self, batch: list[tuple[int, int, int]], trigg
...
+                    # when it is indeed empty, then there is remaining data in the last batch
+                    # because num_samples_in_trigger is not 0.
+                    assert len(triggering_data) > 0 or self.remaining_data_range is not None
+
+                    if self.remaining_data_range is not None:
+                        first_timestamp = self.remaining_data_range[0]
```
MaxiBoether, line 293: What if `len(triggering_data)>0` in this case? Why not just use `triggering_data[0][1]`? I am a bit unsure why we need the remaining data range.

Assuming that `len(triggering_data)==0`, how can there be a first and last timestamp of the samples in the trigger?

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -79,6 +79,8 @@ def __init__(
         self.num_triggers = 0
         self.current_training_id: Optional[int] = None
         self.triggers: list[int] = []
+        # this is to store the first and last timestamp of remaining data in _handle_triggers_within_batch
+        self.remaining_data_range: Optional[tuple[int, int]] = None
```
MaxiBoether, line 83: So if I understand correctly, this stores the range of the leftover data from the last trigger, no?

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
```diff
@@ -356,30 +356,96 @@ def test__handle_triggers_within_batch_empty_triggers(
...
+    pe._handle_triggers_within_batch(first_batch, triggering_indices)
+    assert pe.pipeline_log["supervisor"]["triggers"][0]["first_timestamp"] == 1
+    assert pe.pipeline_log["supervisor"]["triggers"][0]["last_timestamp"] == 2
+
+    assert "first_timestamp" not in pe.pipeline_log["supervisor"]["triggers"][1]
+    assert "last_timestamp" not in pe.pipeline_log["supervisor"]["triggers"][1]
+
+    assert pe.pipeline_log["supervisor"]["triggers"][2]["first_timestamp"] == 3
+    assert pe.pipeline_log["supervisor"]["triggers"][2]["last_timestamp"] == 4
+    assert pe.remaining_data_range == (5, 7)
```
MaxiBoether, line 423: This looks good, I am just confused a bit with the logic it seems haha

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
MaxiBoether: I am wondering if we can make this clear in the name somehow. Otherwise, could you expand the comment to say instead of just "remaining data" to be a bit more precise and say "This stores the first and last timestamp in case we have remaining data after handling all triggers in `_handle_triggers_within_batch `"

MaxiBoether (approved):
Looks good, especially the result in the test is what I expect, I am just a bit confused with the logic. Approving, but it would be great to adjust the one comment first. Thank you

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa: yes you understand it correctly. I don't have a better name but will change the comment!

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa: I answer everything in the unit test Q & A

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
XianzheMa: I answer it in the unit test Q & A!

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
XianzheMa: > How can triggering_data be empty if num_samples_in_trigger > 0? That confuses me a bit. And why can this only happen in the first trigger?

The first `triggering_data` in a batch is not the entire data within the corresponding trigger, but only its subset. The entire data can span over the boundary of two consecutive batches. In that case, it consists of the `remaining_data` from the previous batch and the first `triggering_data` in the current batch. An extreme case is that the part in the current batch (that is, `triggering_data`) is empty.

See the second time `_handle_triggers_within_batch` is called. The first trigger index is -1. So `triggering_data` is indeed `[]`, but the entire data also includes remaining data from the first batch, which is `[(14, 5, 5), (15, 6, 5), (16, 7, 5)]`.

As another example, trigger 5 consists of the first `triggering_data`, i.e. `[(20, 11, 5), (21, 12, 5), (22, 13, 5)]` in the third batch, and the `remaining data`, i.e. `[(19, 10, 5)]` from the second batch.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -181,18 +184,9 @@ def _handle_new_data_batch(self, batch: list[tuple[int, int, int]]) -> bool:
...
         )
 
-        if num_triggers > 0:
-            logger.info(f"There are {num_triggers} triggers in this batch.")
-            self._handle_triggers_within_batch(batch, triggering_indices)
-            return True
-
-        self._sw.start("selector_inform", overwrite=True)
-        selector_log = self.grpc.inform_selector(self.pipeline_id, batch)
-        self.pipeline_log["supervisor"]["selector_informs"].append(
```
XianzheMa, line 191: we call `inform_selector` in two cases:
1. Here, when within a batch there is no trigger.
2. Inside `_handle_triggers_within_batch`, on the remaining data after processing the last trigger.

The logic in two cases is exactly the same. Also we need to update `remaining_data_range` in two cases with the same logic. So I move this part to inside `_handle_triggers_within_batch`.

MaxiBoether: Thank you for simplifying the logic here.

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
MaxiBoether: The first paragraph completely cleared it up for me, thank you for the explanation. Makes total sense. Feel free to merge!

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
MaxiBoether: Thank you, I finally got it with this and the other comment!

PR #432, opened 2024-05-13
fix: Make integration tests runnable locally

robinholzi:
# Motivation

Integration tests configs are currently hard coded. It would be could to be able to centrally configure the config dir for the integration config files to be able to test new versions locally. Being able to place them outside of Git VCS would make debugging easier. 

C++/Cmake config enabling linking with precompiled grpc is taken from #411


cmake/storage_dependencies.cmake
```diff
@@ -66,6 +66,7 @@ endforeach()
 
 
 ################### gRPC ####################
+set(MODYN_GRPC_CPP_PLUGIN "grpc_cpp_plugin")  # default, without gRPC:: prefix
```
MaxiBoether, line 69: I don't fully understand the changes here yet. Could you elaborate a little bit on the `gRPC::` prefix? By default, we use without prefix, and if we find local gRPC, we use it with prefix? In the old implementation, we alias the two. IIRC, the problem was mostly some dependencies inconsistently asking with and without prefix, but I might have missed something.

modyn/storage/src/CMakeLists.txt
```diff
@@ -68,51 +68,10 @@ message(STATUS "Proto generated files in ${PROTO_BINARY_DIR}: ${PROTO_GENERATED_
 
 target_include_directories(modyn-storage-proto PUBLIC "$<BUILD_INTERFACE:${PROTO_BINARY_DIR}>")
 
-target_link_libraries(modyn-storage-proto PUBLIC grpc++ grpc++_reflection)
-
 if (MODYN_USES_LOCAL_GRPC)
-    message(STATUS "Since we are using local gRPC, we need to fix linking issues. If you encounter errors, consider building gRPC from source.")
```
MaxiBoether, line 74: Hmh, this was necessary at the time (at least on my mac using gRPC installed via brew), but apparently, all tests run through without it.

MaxiBoether (approved):
Generally LGTM. I guess I am still a bit confused on the gRPC plugin thing. Maybe you could give some more context on why you changed this. Did you just play around until it worked, or is there some reasoning behind this? (Because at least for me, the current version used to work and allowed me to run tests locally :D)

modyn/storage/src/CMakeLists.txt
```diff
@@ -68,51 +68,10 @@ message(STATUS "Proto generated files in ${PROTO_BINARY_DIR}: ${PROTO_GENERATED_
 
 target_include_directories(modyn-storage-proto PUBLIC "$<BUILD_INTERFACE:${PROTO_BINARY_DIR}>")
 
-target_link_libraries(modyn-storage-proto PUBLIC grpc++ grpc++_reflection)
```
robinholzi, line 71: @MaxiBoether Regarding your comment on the aliasing: I just removed the alias again and it seems to work out fine :) Prior to now I just haven't tested the full matrix of options

MaxiBoether: Ok sure, as long as it works haha

MaxiBoether (approved):
LGTM

PR #435, opened 2024-05-16
feat: Integrate pydantic validation into supervisor

robinholzi:
# Motivation

The supervisor logic can be streamlined by moving the validation logic into the pydantic models.

# Changes

- Add additional validation to pydantic models
- Adjust tests to work on actual "minimal configs"; the dicts previously wouldn't have passed both the old jsonschemas as well as the new pydantic schemas. Adding global fixtures for there models.


cmake/storage_dependencies.cmake
```diff
@@ -66,6 +66,7 @@ endforeach()
 
 
 ################### gRPC ####################
+set(MODYN_GRPC_CPP_PLUGIN "grpc_cpp_plugin")  # default, without gRPC:: prefix
```
robinholzi, line 69: I'll remove this, auto merge didn't remove this properly.

modyn/storage/src/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ protobuf_generate(
     OUT_VAR PROTO_GENERATED_FILES
     LANGUAGE grpc
     GENERATE_EXTENSIONS .grpc.pb.h .grpc.pb.cc
-    PLUGIN "protoc-gen-grpc=\$<TARGET_FILE:grpc_cpp_plugin>"
```
robinholzi, line 62: same here, not needed

modyn/config/examples/example-pipeline.yaml
```diff
@@ -50,10 +50,9 @@ training:
   selection_strategy:
     name: NewDataStrategy
     maximum_keys_in_memory: 500000
-    config:
-      limit: -1
-      reset_after_trigger: True
-      storage_backend: database
+    limit: -1
```
robinholzi, line 53: fyi. when running other pipeline configs from the repo, this also has to be adjusted. I didn't want to update all of these (mostly) duplicated configuration files in this PR.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -48,12 +48,12 @@ def __init__(
         self.reset_after_trigger: bool = config["reset_after_trigger"]
 
         # weighted optimization (with weights supplied by the selector) is quite unusual, so the default us false
-        if "uses_weights" in config:
```
robinholzi, line 51: These updates are needed, also the pydantic model dict dumps contain those fields with None values e.g. `{ "uses_weights": None }`. The current checks with the `in` keyword cannot handle this.

modyn/supervisor/internal/grpc/supervisor_grpc_server.py
```diff
@@ -18,7 +19,7 @@ def callback(modyn_config: dict, server: Any, supervisor: Supervisor) -> None:
 
     def __init__(self, modyn_config: dict) -> None:
         self.modyn_config = modyn_config
-        self.supervisor = Supervisor(modyn_config)
+        self.supervisor = Supervisor(ModynConfig(**modyn_config))
```
robinholzi, line 22: Internally we want to use the pydantic model, however externally we cannot easily offer the ModynConfig as an argument as many tests depend on this. Refactoring the tests in this regard isn't worth it.

modyn/tests/conftest.py
```diff
@@ -0,0 +1,186 @@
+# pylint: disable=redefined-outer-name
```
robinholzi, line 1: Previously we handed over technically invalid configuration dictionaries all the time. Fields that were marked as required in the jsonschema weren't added in the tests configuration just assuming that the implementation wouldn't need them. Now with pydantic we enforce the schema all the time.

modyn/tests/supervisor/internal/test_supervisor.py
```diff
@@ -25,71 +27,11 @@
 PIPELINE_ID = 42
 
 
-def get_minimal_training_config() -> dict:
-    return {
```
robinholzi, line 29: Moved to `conftest.py`

modyn/tests/supervisor/internal/test_supervisor.py
```diff
@@ -172,163 +114,84 @@ def get(self, some_type, pipeline_id):  # pylint: disable=unused-argument
...
-def test__validate_training_options():
-    sup = Supervisor(get_minimal_system_config())
-
-    # Check that our minimal training config gets accepted
-    training_config = get_minimal_training_config()
-    assert sup._validate_training_options(training_config)
-
-    # Check that training without GPUs gets rejected
-    training_config = get_minimal_training_config()
-    training_config["gpus"] = 0
```
robinholzi, line 201: These things are validated in pydantic. No need to test them, as the tests would be more complex than the actual pydantic model.

modyn/supervisor/internal/supervisor.py
```diff
@@ -78,116 +79,63 @@ def init_cluster_connection(self) -> None:
         if "seed" in self.modyn_config:
             self.grpc.seed_selector(self.modyn_config["seed"])
 
-    def _validate_evaluation_options(self, evaluation_config: dict) -> bool:
+    @classmethod
+    def validate_pipeline_config_content(cls, pipeline_config: ModynPipelineConfig) -> bool:
         is_valid = True
-
-        dataset_ids = [dataset["dataset_id"] for dataset in evaluation_config["datasets"]]
```
robinholzi, line 84: Already handled by pydantic model

modyn/supervisor/internal/grpc/supervisor_grpc_server.py
MaxiBoether: Sounds good, let's keep the cost-benefit tradeoff here :) Thank you so much

modyn/storage/src/CMakeLists.txt
MaxiBoether: what do you mean by not needed?

cmake/storage_dependencies.cmake
MaxiBoether: These are conflicts with the integrationtest PR you merged? I am a little confused, I remember seeing these gRPC changes before

modyn/tests/conftest.py
MaxiBoether: Well, let's say I often have been lazy when writing the tests... Thank you for cleaning this up.

MaxiBoether (approved):
This looks good, thank you so much. I am a little confused with the gRPC changes but since you mentioned merging issues I am going to trust that this will be fixed or makes sense. I remember seeing these gRPC changes before in the integrationtest PR. But the other changes all look

modyn/tests/conftest.py
robinholzi: > Well, let's say I often have been lazy when writing the tests... Thank you for cleaning this up.

I mean for a research environment that's a fair approach at the beginning to evaluate things fast. It makes it harder though to adjust previous design decisions. And it can cause more debugging efforts when you don't have fields available that your typing suggests you have.

cmake/storage_dependencies.cmake
robinholzi: I have merged #432  into this PR before Removing those things and merging #432 to main. I guess git just got confused when I merged main into this PR after #432 was merged into main. I didn't want to touch those files in this PR at all.

XianzheMa:
Thank you so much for this effort! Saves a lot of pain

modyn/selector/internal/selector_manager.py
```diff
@@ -188,7 +188,7 @@ def uses_weights(self, pipeline_id: int) -> bool:
     def _instantiate_strategy(self, selection_strategy: dict, pipeline_id: int) -> AbstractSelectionStrategy:
         strategy_name = selection_strategy["name"]
         maximum_keys_in_memory = selection_strategy["maximum_keys_in_memory"]
-        config = selection_strategy["config"] if "config" in selection_strategy else {}
```
XianzheMa, line 191: I don't think this is a semantically equivalent change. 
image

We should be more careful making changes in PRs.

XianzheMa: Ah I guess I get it. The inner class `config` seems to be flattened!

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
```diff
@@ -14,14 +14,14 @@ def __init__(self, downsampling_config: dict, maximum_keys_in_memory: int):
     def _build_downsampling_params(self) -> dict:
         config = super()._build_downsampling_params()
 
-        if "submodular_function" not in self.downsampling_config:
```
XianzheMa, line 17: This is a change to the contrary side.

XianzheMa: will fix in https://github.com/eth-easl/modyn/pull/438

PR #436, opened 2024-05-17
Reject unknown properties in the pipeline config

XianzheMa:
Some configs in the pipeline config are optional, for example `num_samples_to_pass`. If it is not specified, training is limited by another config `epochs_per_trigger`.

It is common for users to make a typo. I once typed `num_sample_to_pass`. The current validation allows additional properties unspecified in the Pydantic model. This leads to the consequence that the pipeline with typo `num_sample_to_pass` passed through, and is executed as if `num_samples_to_pass` is not specified and `epochs_per_trigger` is used. It wastes me one day to find it out.

### DIFF
--- a/modyn/config/schema/pipeline.py
+++ b/modyn/config/schema/pipeline.py
@@ -5,10 +5,16 @@ from typing import Annotated, Any, Dict, List, Literal, Optional, Union
 
 from modyn.supervisor.internal.eval_strategies import OffsetEvalStrategy
 from modyn.utils import validate_timestr
-from pydantic import BaseModel, Field, NonNegativeInt, field_validator, model_validator
+from pydantic import BaseModel as PydanticBaseModel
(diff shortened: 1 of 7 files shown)


modyn/config/schema/pipeline.py
```diff
@@ -312,7 +318,6 @@ class TrainingConfig(BaseModel):
         description="The number of data loader workers on the trainer node that fetch data from storage.", ge=1
     )
     batch_size: float = Field(description="The batch size to be used during training.", ge=1)
-    learning_rate: float = Field(0.1, description="The learning rate to be used during training.")
```
XianzheMa, line 315: `learning_rate` is not controlled by this parameter. They are specified per para groups in `param_groups`.

MaxiBoether (approved):
Thanks Xianzhe! This is indeed helpful to find typos in the config...

PR #437, opened 2024-05-18
refactor: Adjust pipeline orchestration in `PipelineExecutor`

robinholzi:
# Motivation

The current state of `pipeline_executor.py` doesn't follow clear sequential or logical order. Also, the setup doesn't allow to automatically profile and log pipeline stage execution. Restructuring the orchestration of a pipeline in a PipelineExecutor here.

# Changes

- Split all logical pipeline stages into separate `PipelineExecutor` member functions and merge unused PipelineStages (like `RUN_TRAINING` and `TRAINING_DONE`). This allows to have a 1:1 mapping between logical pipeline stages and member functions.
- Sort the pipeline stage functions sequentially according to their position in the pipeline
- Add `@pipeline_stage(...)` decorator which wraps every pipeline stage function. When calling the decorated function we record meta information for `logging` (post-experiment analysis) and `tracking` (for online feedback) purposes. This includes `start` & `end` timestamps as well as the current "dataset/sample time". This is especially useful in experiments to be able to sort/plot different measurements by the time of the processed samples. 
- Move all state (that is not enclosed in controller objects like `grpc`) into a central `ExecutionState` dataclass.
- Adjusted tests accordingly, removing code duplication in tests by parameterizing tests with `pytest.mark.parameterize(...)`
- Some renaming and other minor adjustments

### DIFF
--- a/modyn/supervisor/internal/evaluation_result_writer/json_result_writer.py
+++ b/modyn/supervisor/internal/evaluation_result_writer/json_result_writer.py
@@ -1,12 +1,13 @@
 import json
 import pathlib
+from typing import Any
 
 # pylint: disable=no-name-in-module
 from modyn.evaluator.internal.grpc.generated.evaluator_pb2 import EvaluationData
 from modyn.supervisor.internal.evaluation_result_writer import AbstractEvaluationResultWriter
 
 
-class JsonResultWriter(AbstractEvaluationResultWriter):
+class _JsonResultWriter(AbstractEvaluationResultWriter):
     def __init__(self, pipeline_id: int, trigger_id: int, eval_directory: pathlib.Path):
         super().__init__(pipeline_id, trigger_id, eval_directory)
         self.results: dict = {"datasets": []}
@@ -17,7 +18,19 @@ class JsonResultWriter(AbstractEvaluationResultWriter):
             dataset_results["metrics"].append({"name": metric.metric, "result": metric.result})
         self.results["datasets"].append({dataset_id: dataset_results})
 
-    def store_results(self) -> None:
+
+class JsonResultWriter(_JsonResultWriter):
+    """Does not dump into files itself but only returns the result."""
+
+    def store_results(self) -> dict[str, Any]:
+        return self.results
+
+
+class DedicatedJsonResultWriter(_JsonResultWriter):
+    """Dumps every log result into a dedicated file."""
+
+    def store_results(self) -> pathlib.Path:
         file_name = f"{self.pipeline_id}_{self.trigger_id}.eval"
         with open(self.eval_directory / file_name, "w+", encoding="utf-8") as output_file:
             json.dump(self.results, output_file)
+        return self.eval_directory / file_name
--- a/modyn/supervisor/internal/grpc/enums.py
+++ b/modyn/supervisor/internal/grpc/enums.py
@@ -18,27 +18,68 @@ class PipelineStatus(StrEnum):
     NOTFOUND = "not found"
 
 
+class PipelineType(StrEnum):
+    MAIN = "main"
+
+    REPLAY_DATA = "replay_data"
+    SERVE_ONLINE = "wait_for_new_data"
+
+    NEW_DATA = "new_data"
+    NEW_BATCH = "new_data"
+
+    TRIGGER = "trigger"
+    TRAINING = "training"
+    EVALUATION = "evaluation"
+
+
 class PipelineStage(StrEnum):
+    """For a state transition graph checkout the `PIPELINE.md` file."""
+
+    # Setup
     INIT = "Initialize pipeline executor"
     INIT_CLUSTER_CONNECTION = "Initialize cluster connection"
-    GET_SELECTOR_BATCH_SIZE = "Get selector batch size"
-    HANDLE_NEW_DATA = "Handle new data"
-    NEW_DATA_HANDLED = "New data handled"
-    RUN_TRAINING = "Run training"
-    WAIT_FOR_TRAINING_COMPLETION = "Wait for training completion"
-    TRAINING_COMPLETED = "Training completed 🚀"
-    WAIT_FOR_EVALUATION_COMPLETION = "Wait for evaluation completion"
-    EVALUATION_COMPLETED = "Evaluation completed 🚀"
-    STORE_TRAINED_MODEL = "Store trained model"
-    EVALUATE = "Run evaluation"
-    STORE_EVALUATION_RESULTS = "Store evaluation results"
-    HANDLE_TRIGGERS_WITHIN_BATCH = "Handle triggers within batch"
-    INFORM_SELECTOR_AND_TRIGGER = "Inform selector and trigger"
-    INFORM_SELECTOR_REMAINING_DATA = "Inform selector about remaining data"
+
+    # Replay Data
     REPLAY_DATA = "Replay data"
     REPLAY_DATA_DONE = "Replay data done"
+
+    # Wait for new data
+    SERVE_ONLINE_DATA = "Serve online data"
     FETCH_NEW_DATA = "Fetch new data"
     WAIT_FOR_NEW_DATA = "Wait for new data"
+
+    # Process new data
+    PROCESS_NEW_DATA = "Process new data"
+
+    PROCESS_NEW_DATA_BATCH = "Process new data batch"
+    EVALUATE_TRIGGER_POLICIES = "Evaluate trigger on batch"
+    INFORM_SELECTOR_NO_TRIGGER = "Inform selector about no trigger"
+
+    # Trigger
+    EXECUTE_TRIGGERS = "Execute triggers within batch"
+    EXECUTE_SINGLE_TRIGGER = "Execute single trigger"
+    INFORM_SELECTOR_AND_TRIGGER = "Inform selector and trigger"
+    INFORM_SELECTOR_REMAINING_DATA = "Inform selector about remaining data"
+
+    NEW_DATA_HANDLED = "New data handled"
+
+    # Training
+    TRAIN_AND_STORE_MODEL = "Train and store model"
+    TRAIN = "Run training"
+
+    WAIT_FOR_TRAINING_COMPLETION = "Wait for training completion"
+    """Implements busy waiting for training completion."""
+
+    TRAINING_COMPLETED = "Training completed 🚀"
+    STORE_TRAINED_MODEL = "Store trained model"
+
+    # Evaluation
+    EVALUATE = "Run evaluation"
+    EVALUATE_SINGLE = "Run single evaluation"
+    WAIT_FOR_EVALUATION_COMPLETION = "Wait for evaluation completion"
+    STORE_EVALUATION_RESULTS = "Store evaluation results"
+
+    # Teardown
     DONE = "Pipeline done"
     EXIT = "Exit"
 
--- a/modyn/supervisor/internal/grpc/enums.py
+++ b/modyn/supervisor/internal/grpc/enums.py
@@ -18,22 +18,10 @@ class PipelineStatus(StrEnum):
     NOTFOUND = "not found"
 
 
-class PipelineType(StrEnum):
-    MAIN = "main"
-
-    REPLAY_DATA = "replay_data"
-    SERVE_ONLINE = "wait_for_new_data"
-
-    NEW_DATA = "new_data"
-    NEW_BATCH = "new_data"
-
-    TRIGGER = "trigger"
-    TRAINING = "training"
-    EVALUATION = "evaluation"
-
-
 class PipelineStage(StrEnum):
     """For a state transition graph checkout the `PIPELINE.md` file."""
+
+    MAIN = "Pipeline entrypoint"
 
     # Setup
     INIT = "Initialize pipeline executor"
@@ -52,13 +40,13 @@ class PipelineStage(StrEnum):
     PROCESS_NEW_DATA = "Process new data"
 
     PROCESS_NEW_DATA_BATCH = "Process new data batch"
-    EVALUATE_TRIGGER_POLICIES = "Evaluate trigger on batch"
+    EVALUATE_TRIGGER_POLICY = "Evaluate trigger on batch"
     INFORM_SELECTOR_NO_TRIGGER = "Inform selector about no trigger"
 
     # Trigger
-    EXECUTE_TRIGGERS = "Execute triggers within batch"
-    EXECUTE_SINGLE_TRIGGER = "Execute single trigger"
-    INFORM_SELECTOR_AND_TRIGGER = "Inform selector and trigger"
+    HANDLE_TRIGGERS = "Handle triggers within batch"
+    HANDLE_SINGLE_TRIGGER = "Handle single trigger"
+    INFORM_SELECTOR_ABOUT_TRIGGER = "Inform selector and trigger"
     INFORM_SELECTOR_REMAINING_DATA = "Inform selector about remaining data"
 
     NEW_DATA_HANDLED = "New data handled"
--- a/modyn/supervisor/internal/grpc_handler.py
+++ b/modyn/supervisor/internal/grpc_handler.py
@@ -285,7 +285,7 @@ class GRPCHandler:
         lr_scheduler_configs = {}
         if pipeline_config["training"].get("lr_scheduler"):
             lr_scheduler_configs = pipeline_config["training"]["lr_scheduler"]
-            lr_scheduler_configs["config"] = lr_scheduler_configs.get("config", None)
+            lr_scheduler_configs["config"] = lr_scheduler_configs.get("config") or {}
 
         criterion_config = json.dumps(pipeline_config["training"]["optimization_criterion"].get("config") or {})
 
--- a/modyn/supervisor/internal/grpc_handler.py
+++ b/modyn/supervisor/internal/grpc_handler.py
@@ -44,8 +44,6 @@ from modyn.storage.internal.grpc.generated.storage_pb2 import (
 )
 from modyn.storage.internal.grpc.generated.storage_pb2_grpc import StorageStub
 from modyn.supervisor.internal.evaluation_result_writer import AbstractEvaluationResultWriter
-from modyn.supervisor.internal.grpc.enums import IdType, MsgType, PipelineStage
-from modyn.supervisor.internal.grpc.template_msg import id_submsg, pipeline_stage_msg
 from modyn.supervisor.internal.utils import EvaluationStatusReporter, TrainingStatusReporter
 from modyn.trainer_server.internal.grpc.generated.trainer_server_pb2 import CheckpointInfo, Data
 from modyn.trainer_server.internal.grpc.generated.trainer_server_pb2 import JsonString as TrainerServerJsonString
@@ -159,9 +157,10 @@ class GRPCHandler:
 
         return response.available
 
-    def get_new_data_since(
-        self, dataset_id: str, timestamp: int
-    ) -> Iterable[tuple[list[tuple[int, int, int]], dict[str, Any]]]:
+    def get_new_data_since(self, dataset_id: str, timestamp: int) -> Iterable[tuple[list[tuple[int, int, int]], int]]:
+        """Returns:
+        tuple containing actual data and fetch time in milliseconds
+        """
         assert self.storage is not None
         if not self.connected_to_storage:
             raise ConnectionError("Tried to fetch data from storage, but no connection was made.")
@@ -177,7 +176,10 @@ class GRPCHandler:
 
     def get_data_in_interval(
         self, dataset_id: str, start_timestamp: int, end_timestamp: int
-    ) -> Iterable[tuple[list[tuple[int, int, int]], dict[str, Any]]]:
+    ) -> Iterable[tuple[list[tuple[int, int, int]], int]]:
+        """Returns:
+        tuple containing actual data and fetch time in milliseconds
+        """
         assert self.storage is not None
         if not self.connected_to_storage:
             raise ConnectionError("Tried to fetch data from storage, but no connection was made.")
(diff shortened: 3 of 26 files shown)


robinholzi:
Resolving the merge conflicts was an odyssey here🫨
@XianzheMa @MaxiBoether Let me know in case I dropped any of the recent changes on `main` (I mostly reimplemented your changes in `(test_)pipeline_executor.py`

modyn/supervisor/internal/grpc/enums.py
```diff
@@ -18,27 +18,68 @@ class PipelineStatus(StrEnum):
     NOTFOUND = "not found"
 
 
+class PipelineType(StrEnum):
+    MAIN = "main"
+
+    REPLAY_DATA = "replay_data"
+    SERVE_ONLINE = "wait_for_new_data"
```
XianzheMa, line 25: I thought we could have a strictly identical variable name to the literal name?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
             optimizer_config["source"] = optimizer["source"]
             optimizer_config["param_groups"] = []
             for param_group in optimizer["param_groups"]:
-                config_dict = param_group["config"] if "config" in param_group else {}
+                config_dict = param_group.get("config") or {}
```
XianzheMa, line 281: `get` has the second parameter which is returned if key is absent

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
...
                 optimizer_config["param_groups"].append({"module": param_group["module"], "config": config_dict})
             optimizers_config[optimizer["name"]] = optimizer_config
 
         lr_scheduler_configs = {}
-        if "lr_scheduler" in pipeline_config["training"]:
+        if pipeline_config["training"].get("lr_scheduler"):
```
XianzheMa, line 286: why this change?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
...
             lr_scheduler_configs = pipeline_config["training"]["lr_scheduler"]
-            if "config" not in lr_scheduler_configs:
-                lr_scheduler_configs["config"] = {}
+            lr_scheduler_configs["config"] = lr_scheduler_configs.get("config", None)
```
XianzheMa, line 288: these are not equal. Why the change?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
...
-            criterion_config = json.dumps(pipeline_config["training"]["optimization_criterion"]["config"])
-        else:
-            criterion_config = "{}"
+        criterion_config = json.dumps(pipeline_config["training"]["optimization_criterion"].get("config") or {})
 
-        if "epochs_per_trigger" in pipeline_config["training"]:
-            epochs_per_trigger = pipeline_config["training"]["epochs_per_trigger"]
-        else:
-            epochs_per_trigger = 1
+        epochs_per_trigger = pipeline_config["training"].get("epochs_per_trigger") or 1
```
XianzheMa, line 292: above can both be re-written with the second arg of `get`

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
...
-            transform_list = []
-
-        if "label_transformer_function" in pipeline_config["data"]:
-            label_transformer = pipeline_config["data"]["label_transformer_function"]
-        else:
-            label_transformer = ""
+        seed = pipeline_config["training"].get("seed", None)
+        tokenizer = pipeline_config["data"].get("tokenizer", None)
+        transform_list = pipeline_config["data"].get("transformations") or []
+        label_transformer = pipeline_config["data"].get("label_transformer_function") or ""
```
XianzheMa, line 316: also question on why the second arg of `get` is not used

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -654,11 +599,6 @@ def wait_for_evaluation_completion(
             working_queue.append(current_evaluation_id)
             sleep(1)
 
-        self.pipeline_status_queue.put(
-            pipeline_stage_msg(
-                PipelineStage.EVALUATION_COMPLETED, MsgType.ID, id_submsg(IdType.TRAINING, training_id), True
-            )
-        )
```
XianzheMa, line 661: I think most of changes in this file are unrelated with the purpose of this PR. And `if key not in dict` does not look worse than `if dict.get(key)`. To make sure we didn't accidentally introduce something wrong in the request preparation, I suggest we keep the old code.

XianzheMa (changes requested):
Thanks a lot for the efforts! Yes with the refactored code it looks cleaner, but it's better not to refactor everything in one PR for the reasons:

1. It is difficult to ensure "with naked eyes" the functionality equivalence before and after the changes. I am worried about introducing unforeseen bugs (also missing tests).
2. `pipeline_executor.py` is one of the most important files as it orchestrates the entire lifecycle of a pipeline. We very much rely on its correctness to run experiments. The current version has been used for a long time and thus has a high credibility on correctness. Ensuring its credibility on correctness to run experiments (after all this is our main purpose), should be prioritized over code style refactoring, unless a really needed feature has to be introduced.

In general I really like how it looks after this PR, but I am afraid we miss something during this big refactoring, and then experiments go wrong unnoticed (and we need to spend more time to find it out). Can we split it into multiple PRs that each changes a part of it, so that we have a smooth refactoring?

modyn/supervisor/internal/grpc_handler.py
XianzheMa: In general I advocate the principle "if there is not a really justified reason for a change, then keep it the same".

A pure code style change just adds more work to the reviewer and more risk to uncaught bugs (for example [here](https://github.com/eth-easl/modyn/pull/435/files#r1606098572), hard to catch by the reviewer), because our unit tests certainly do not cover a lot of cases. (Just passing all the tests does not mean it is risk-free). And the old code has less risk to go wrong because we have used it for long.

I hope my thoughts make sense!

XianzheMa: Ah now I see your point in https://github.com/eth-easl/modyn/pull/435/files#r1603030383. Yes in that way this makes sense. Sorry for the misunderstanding!

modyn/supervisor/internal/grpc/enums.py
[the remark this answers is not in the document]
robinholzi: I renamed the enum with my IDE from another place and forgot about the string value! Thanks for the hint!

modyn/supervisor/internal/grpc_handler.py
robinholzi: I know, the two variants are not equivalent though and unfortunately, the key absence isn't enough here:

`{"key": None}.get("key", {}) == None`
`{"key": None}.get("key") or {} == {}`

modyn/supervisor/internal/grpc_handler.py
robinholzi: Using the `in` operator for dicts isn't safe in python as it semantically doesn't always reflect what is intended. Dumping pydantic models can e.g. yield `None` values for dict fields. `in` treats the keys as present while no value is present.

modyn/supervisor/internal/grpc_handler.py
robinholzi: No

MaxiBoether:
hey Robin, thanks for the PR! I will go through it _probably_ on Thursday/Friday. I have to prepare teaching and a NeurIPS resubmission today and tomorrow, and don't think I will have a big enough consecutive timeslot to review before Thursday. Just so you know

robinholzi:
> hey Robin, thanks for the PR! I will go through it _probably_ on Thursday/Friday. I have to prepare teaching and a NeurIPS resubmission today and tomorrow, and don't think I will have a big enough consecutive timeslot to review before Thursday. Just so you know

Thanks Maxi! Thanks totally fine for me :) Let me know if you have any questions, I can always provide you some insights in person :)

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -1,554 +1,855 @@
...
+            replay_data_generator = self.grpc.get_new_data_since(s.dataset_id, s.start_replay_at)
         else:
-            logger.info(f"Evaluation started for model {model_id} on split {interval_start} to {interval_end}.")
-            reporter = EvaluationStatusReporter(
-                self.eval_status_queue, response.evaluation_id, eval_dataset_id, response.dataset_size
+            replay_data_generator = self.grpc.get_data_in_interval(s.dataset_id, s.start_replay_at, s.stop_replay_at)
+
+        for replay_data, request_time in replay_data_generator:
+            # setting sample here to have correct logs in process_new_data
+            s.current_sample_time = min(
```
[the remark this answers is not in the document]
MaxiBoether: also, is this really necessary or not? Because the minimum search might be unnecessary compute for large batches

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -1,554 +1,855 @@
...
-        self._update_pipeline_stage_and_enqueue_msg(
-            PipelineStage.STORE_TRAINED_MODEL, MsgType.ID, id_submsg(IdType.TRIGGER, trigger_id)
-        )
-        # We store the trained model for evaluation in any case.
-        self._sw.start("store_trained_model", overwrite=True)
-        model_id = self.grpc.store_trained_model(self.current_training_id)
-        self.trigger.inform_previous_model(model_id)
-        self.pipeline_log["supervisor"]["triggers"][trigger_id]["store_trained_model_time"] = self._sw.stop()
+    @pipeline_stage(PipelineType.NEW_BATCH, PipelineStage.EVALUATE_TRIGGER_POLICIES, track=True)
+    def _evaluate_trigger_policies(
```
MaxiBoether, line 411: nit: is there any reason this is plural? conceptually, there is just a single policy

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -1,554 +1,855 @@
...
+            The list of the actually processed triggers
+        """
+        logger.info(f"Processing {len(s.triggers)} triggers in this batch.")
+        s.pipeline_status_queue.put(pipeline_stage_msg(PipelineStage.EXECUTE_TRIGGERS, MsgType.GENERAL))
+
+        previous_trigger_index = 0
+        trigger_index = -1
+        trigger_indexes: list[int] = []
+        for i, trigger_index in enumerate(lazy_trigger_indexes):
+            # Run training and evaluation subpipelines
```
MaxiBoether, line 456: needs to be updated without subpipeline concept

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -1,554 +1,855 @@
...
-            else:
-                first_timestamp = triggering_data[0][1]
-                last_timestamp = triggering_data[-1][1]
-        else:
-            assert len(triggering_data) > 0
-            # since num_samples_in_trigger is not 0, we are sure that triggering_data is not empty
-            first_timestamp = triggering_data[0][1]
-            last_timestamp = triggering_data[-1][1]
+    @pipeline_stage(PipelineType.NEW_BATCH, PipelineStage.EXECUTE_TRIGGERS, track=True)
+    def _execute_triggers(
```
MaxiBoether, line 434: I am not sure whether I like the wording `execute` here. Maybe `handle` (because we _execute_ training jobs, while handling triggers), although handle is also quite general...

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
```diff
@@ -116,199 +139,222 @@ def teardown():
...
         get_new_data_mock: MagicMock
         with patch.object(pe.grpc, "get_new_data_since", side_effect=mocked_get_new_data_since) as get_new_data_mock:
-            pe.wait_for_new_data(21)
+            dummy_execution_state.max_timestamp = 21
+            pe._serve_online_data(dummy_execution_state, dummy_logs)
 
-            assert handle_mock.call_count == 6
-            assert get_new_data_mock.call_count == 3
+            assert handle_mock.call_count == 3
+            assert get_new_data_mock.call_count == 3 + 1
```
MaxiBoether, line 304: why the +1?

modyn/tests/supervisor/internal/pipeline_executor/test_pipeline_executor.py
```diff
@@ -116,199 +139,222 @@ def teardown():
...
-                call([(14, 45, 3)]),
-                call([]),
+                call(ANY, ANY, [(10, 42, 0), (11, 43, 0), (12, 43, 1)], 97),
+                call(ANY, ANY, [(13, 43, 2), (14, 45, 3)], 98),
+                call(ANY, ANY, [], 99),
             ]
             assert handle_mock.call_args_list == expected_handle_mock_arg_list
 
-            expected_get_new_data_arg_list = [call("test", 21), call("test", 43), call("test", 45)]
+            expected_get_new_data_arg_list = [call("test", 21), call("test", 43), call("test", 45), call("test", 45)]
```
MaxiBoether, line 313: why the addition here?

MaxiBoether (changes requested):
Thank you Robin! My comments are mostly nitpicky. The biggest thing that needs to change is as discussed today to port the removal of the "multiple pipeline concept" here, and instead have this parent relationship. After addressing the open comments (esp in larger PRs, I prefer to handle reviews such that the author acknowledges everything that has been done with a comment like "addressed" or "done", and then the reviewer solves it when he's happy because otherwise stuff just gets lost) and porting this, I will do another round and then we can merge this :) thank you!

modyn/utils/timer.py
```diff
@@ -0,0 +1,20 @@
+import time
+from typing import Generator, TypeVar
+
+X = TypeVar("X")
+
+
+def timed_generator(generator: Generator[X, None, None]) -> Generator[tuple[X, float], None, None]:
```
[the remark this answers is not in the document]
robinholzi: These TypeVars behave a bit like C++ templates (without the whole compile-time concept though, they only serve type annotation purposes and don't change the runtime behavior).

The same concepts applies to decorators. these type annotations allow us that the returned result of the function/generator has exactly the same types in place of the TypeVars as the input object. With `typing.any` we would lose this information. This is annoying if you e.g. pass a complex data structure as `X` as you wouldn't get any IDE type hints for the returned result.

modyn/tests/utils/test_timer.py
```diff
@@ -0,0 +1,33 @@
+import time
+from typing import Generator
+
+from modyn.utils.timer import timed_generator
+
+
+def test_timed_generator() -> None:
```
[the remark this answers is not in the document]
robinholzi: Yeah, it failed once or twice while my local system was under full load, in the CI it should be fine though as no other heavy processes are executed at the runners.

modyn/supervisor/internal/supervisor.py
```diff
@@ -241,8 +241,9 @@ def start_pipeline(
                 eval_status_queue,
             )
             return pipeline_res_msg(pipeline_id=pipeline_id)
-        except Exception:  # pylint: disable=broad-except
-            return pipeline_res_msg(pipeline_id=pipeline_id, exception="Failed to execute pipeline")
+        except Exception as ex:  # pylint: disable=broad-except
+            raise ex
+            # return pipeline_res_msg(pipeline_id=pipeline_id, exception="Failed to execute pipeline")
```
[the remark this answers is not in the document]
robinholzi: Just a debugging artefact I forgot to remove, thanks for noticing!

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -0,0 +1,406 @@
...
+    `end-start` is not always the actual duration this stage spent in computing."""
+
+    # dataset time of last seen sample
+    sample_idx: int
+    sample_time: int
+
+    # stage specific log info
+    info: StageInfo | None = Field(None)
+
+    def online_df(self, extended: bool = False) -> pd.DataFrame | None:
```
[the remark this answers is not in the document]
robinholzi: The naming of `SelectorInformLog` is off, it should be `SelectorInformInfo`. Like `TrainingInfo`, `StoreEvaluationInfo`, ... it is additional logging info that only occurs at one certain stage and is therefore included in the `info` property of `StageLog` that has type `StageInfo`. `TrainingInfo`, `StoreEvaluationInfo`, etc. therefore all subclass `StageInfo` (directly or transitively).

The `stage_runs_df` property function of `SupervisorLogs` was indeed not used anymore and even relied on non-existent functions, it surprises me that the linting didn't catch that. I removed it completely.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -1,554 +1,855 @@
...
+from modyn.supervisor.internal.grpc.enums import CounterAction, IdType, MsgType, PipelineStage, PipelineType
 from modyn.supervisor.internal.grpc.template_msg import counter_submsg, dataset_submsg, id_submsg, pipeline_stage_msg
 from modyn.supervisor.internal.grpc_handler import GRPCHandler
 from modyn.supervisor.internal.triggers import Trigger
 from modyn.supervisor.internal.utils import EvaluationStatusReporter
 from modyn.utils import dynamic_module_import
+from modyn.utils.timer import timed_generator
+from typing_extensions import Concatenate, ParamSpec
+
+from .models import (
```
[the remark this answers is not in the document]
robinholzi: I'll convert it to an absolute one. From my experience, it is desirable to use relative imports whenever possible (when that doesn't require going back beyond the current package, i.e. the parent). The benefit of that is that you see dependencies more clearly. Isort would e.g. also bundle them together in the last import block so "local" dependencies are visible more clearly.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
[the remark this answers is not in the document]
robinholzi: I think it doesn't make a big difference what sample time we choose for a batch, it should just be one of the times in the batch. Let's use the time of the first sample, wdyt?

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: It would be possible to evaluate multiple policies before deciding. However, the final decision could then be again seen as a single, higher-order (aggregation) policy. So singular makes sense I guess.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: maybe `invoke`?

robinholzi: Okay no, let's just stick with `handle`

robinholzi:
@MaxiBoether I think I've addressed all your comments, thanks for the suggestions. Once the CI is green you can checkout the diff in #439 (includes a couple of other tweaks that make the new logging format usable).

modyn/utils/timer.py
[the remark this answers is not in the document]
MaxiBoether: Ahhh, I see. I thought the IDE would be able to infer the type with typing.Any (so i thought I would get exactly the behavior you mention using typing.Any. But then ok thanks for letting me know

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -0,0 +1,499 @@
...
+
+    config: ConfigLogs
+
+    experiment: bool
+    start_replay_at: int | None = Field(None, description="Epoch to start replaying at")
+    stop_replay_at: int | None = Field(None, description="Epoch to stop replaying at")
+
+    # incremental logs
+
+    supervisor: SupervisorLogs = Field(default_factory=SupervisorLogs)
```
MaxiBoether, line 445: rename to supervisor_logs? was thinking for a moment that this owns a superviosr object now.

modyn/supervisor/internal/evaluation_result_writer/json_result_writer.py
```diff
@@ -17,7 +18,19 @@ def add_evaluation_data(self, dataset_id: str, dataset_size: int, evaluation_dat
             dataset_results["metrics"].append({"name": metric.metric, "result": metric.result})
         self.results["datasets"].append({dataset_id: dataset_results})
 
-    def store_results(self) -> None:
+
+class JsonResultWriter(_JsonResultWriter):
```
MaxiBoether, line 22: I don't get the purpose of this yet

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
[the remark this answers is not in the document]
MaxiBoether: I am still a bit confused. According to the docstring, it's "The unix timestamp of the last sample seen and processed by the pipeline executor". Shouldn't it be the maximum then? The role of this field is not clear to me yet. Is this just for logging/plotting purposes?

MaxiBoether (approved):
Generally, to the best I can overlook this, this looks good. Some minor things are still open.

I am mostly confused about the dfs. We have `df()` on the `SupervisorLogs`, we have the `online_df()` function where I am a little confused by the name, and the relationship of the dfs and logs is not 100% clear yet. Maybe it would be helpful if you could write a paragraph somewhere on how logging works now? Mainly on the dfs, the pipeline log and what their relationship is, when what gets materialized etc. I think that would be helpful. Thank you, we're getting close!

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -0,0 +1,499 @@
...
+    system: ModynConfig
+    pipeline: ModynPipelineConfig
+
+
+class StageInfo(BaseModel):
+    """Base class for stage log info. Has no members, intended to be subclassed"""
+
+    def online_df(self) -> pd.DataFrame | None:
+        """
+        Some stages might want to store data in the online DataFrame.
```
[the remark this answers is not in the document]
robinholzi: I will strip the online from the name and add a global documentation `docs/LOGGING.md` that explains how logging works and how one can use it.

modyn/supervisor/internal/pipeline_executor/models.py
robinholzi: Will do! I thought that would be clear from the typing, but fair point.

modyn/supervisor/internal/evaluation_result_writer/json_result_writer.py
robinholzi: Previously we wrote all evaluation calls to dedicated files. The default should now be not to create one file per evaluation request but use the central json that is built by the `pipeline_executor`. To keep the old functionality i renamed the old JsonResultWriter to `DedicatedJsonResultWriter` while the new one doesn't create files on its own.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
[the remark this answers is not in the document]
robinholzi: Yes, it's only for logging/plotting. It should take the values of a representative sample in the current (training) batch. That allows to sort the pipeline stage execution logs by the dataset time dimension.

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -0,0 +1,499 @@
+# pylint: disable=E1133
```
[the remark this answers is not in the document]
robinholzi: Not sure anymore, I think we can drop it.

robinholzi:
@XianzheMa I'm happy to give you a short intro to the new logging setup in case the `LOGGING.md` documentation isn't sufficient :)

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -276,69 +278,47 @@ def start_training(
...
+        seed = pipeline_config["training"].get("seed", None)
+        tokenizer = pipeline_config["data"].get("tokenizer", None)
+        transform_list = pipeline_config["data"].get("transformations") or []
+        label_transformer = pipeline_config["data"].get("label_transformer_function") or ""
 
         if pipeline_config["training"]["checkpointing"]["activated"]:
             if (
-                "interval" not in pipeline_config["training"]["checkpointing"]
-                or "path" not in pipeline_config["training"]["checkpointing"]
+                pipeline_config["training"]["checkpointing"].get("interval") is None
```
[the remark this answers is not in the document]
robinholzi: Probably I'm not always consistent, sometimes I just want to make it more explicit (e.g. when using the value for more than just `if` blocks).

robinholzi:
@MaxiBoether Huge thanks for the effort reviewing this 💯

PR #438, opened 2024-05-19
Fix selection strategy config

XianzheMa:


modyn/config/schema/pipeline.py
```diff
@@ -101,8 +101,13 @@ class PresamplingConfig(BaseModel):
 class DownsamplingConfig(BaseModel):
     """Config for the downsampling strategy of SelectionStrategy."""
 
-    strategy: Literal["Random", "RandomNoReplacement", "LabelBalanced", "TriggerBalanced", "Loss", "No"] = Field(
-        description="Strategy used to downsample the data. Available strategies: Loss, Gradnorm, No (no downsampling)."
+    class Config:
```
robinholzi, line 104: The most idiomatic solution would probably be to create subclasses for every strategy. I'm fine to leave this for now though.

modyn/selector/internal/selector_strategies/presampling_strategies/utils.py
```diff
@@ -7,29 +7,18 @@ def instantiate_presampler(
...
-        # Try to instantiate the class even if the short name is used
-        presampling_class = instantiate_class(
-            "modyn.selector.internal.selector_strategies.presampling_strategies",
-            presampling_strategy + "PresamplingStrategy",
-            presampling_config,
-            modyn_config,
-            pipeline_id,
-            storage_backend,
-        )
+    presampling_class = instantiate_class(
```
robinholzi, line 16: Nice and clean, thanks! :)

robinholzi (approved):
Looks good, thanks!

modyn/config/schema/pipeline.py
XianzheMa: yes I agree with you. It's just that we are not about to explore the different downsampling strategy now so I didn't make the effort. Thanks for mentioning it!

XianzheMa:
What fails is a linter issue. Fixed in the next PR.

PR #439, opened 2024-05-21
feat: Adjust logging with more meta information for plotting

robinholzi:
integrated suggestions of #437 and other adjustments that are needed to use the new setup for logging and plotting.

No need for dedicated review, I'll just merge this PR into the other one

### DIFF
--- a/modyn/supervisor/internal/grpc/enums.py
+++ b/modyn/supervisor/internal/grpc/enums.py
@@ -18,22 +18,10 @@ class PipelineStatus(StrEnum):
     NOTFOUND = "not found"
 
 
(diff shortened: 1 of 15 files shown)


modyn/supervisor/internal/grpc/enums.py
```diff
@@ -18,23 +18,11 @@ class PipelineStatus(StrEnum):
     NOTFOUND = "not found"
 
 
-class PipelineType(StrEnum):
```
robinholzi, line 21: Decided that a parent relation is more useful for pipeline hierarchy graph/tree plot.

PR #440, opened 2024-05-22
Log the pydantic validation error and add tokenizer to evaluation dataset schema

XianzheMa:
This PR logs the validation error on supervisor side so that once validation fails, user knows what's going wrong and how to fix the pipeline.

Also the configuration for evaluation dataset lacks a `tokenizer`. Considering that the evaluation dataset config resembles the training dataset config, I make an inheritance.

Also the type of `transformations` are corrected.


MaxiBoether (approved):
Thanks, LGTM!

PR #441, opened 2024-05-22
test: Add unit tests enforcing all configs to be up to date

robinholzi:
# Motivation

Let's ensure that our configs never get outdated. It should be the responsibility of the person who changes any configuration option, to update all actively used configs in the repo.

### DIFF
--- a/modyn/tests/config/test_config_integrity.py
+++ b/modyn/tests/config/test_config_integrity.py
@@ -7,10 +7,11 @@ from modynclient.config import read_client_config
 """Directories can either be file paths or directories.
 If they are directories, all yaml files in the directory will be used."""
 
+PROJECT_ROOT = Path(__file__).parents[3]
+
 MODYN_PIPELINE_CONFIG_PATHS: list[str] = [
-    "modynclient/config/examples/dummy.yaml",
-    "modynclient/config/examples/mnist.yaml",
-    "benchmark/",
+    "benchmark/mnist",
+    "benchmark/wildtime_benchmarks",
 ]
 
 MODYN_SYSTEM_CONFIG_PATHS = ["modyn/config/examples/modyn_config.yaml"]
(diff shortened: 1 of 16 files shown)


MaxiBoether:
Nice! Maybe we could also either grep the benchmark/config directories for all yaml/yml files, and use that, or have a test that compares the hardcoded list to a grepped list and fails if they don't contain the same files? To avoid adding new configs and forgetting to add them here.

robinholzi:
> Nice! Maybe we could also either grep the benchmark/config directories for all yaml/yml files, and use that, or have a test that compares the hardcoded list to a grepped list and fails if they don't contain the same files? To avoid adding new configs and forgetting to add them here.

Good idea! We can only do that for directories that only contain one kind of config file though.

modyn/tests/config/test_config_integrity.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def _discover_files(paths: list[str]) -> list[Path]:
+    files = []
+    for path in paths:
+        p = Path(path)
+        if not p.exists():
+            continue
+        if p.is_dir():
+            files.extend(p.rglob("*.yaml"))
```
MaxiBoether, line 31: maybe also yml? I feel like we are not 100% consistent with yaml vs yml as an extension

modyn/tests/config/test_config_integrity.py
```diff
@@ -0,0 +1,55 @@
...
+        else:
+            files.append(p)
```
MaxiBoether, line 33: maybe only if the file is yaml or yml?

modyn/tests/config/test_config_integrity.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pytest
+from modyn.config import read_modyn_config, read_pipeline
+from modynclient.config import read_client_config
+
+"""Directories can either be file paths or directories.
+If they are directories, all yaml files in the directory will be used."""
+
+MODYN_PIPELINE_CONFIG_PATHS: list[str] = [
+    "modynclient/config/examples/dummy.yaml",
```
MaxiBoether, line 11: These paths are probably based on the working directory. Does this also work when I run pytest from some other directory than the project root? I think typically we do something like in the integration tests:

```
SCRIPT_PATH = pathlib.Path(os.path.realpath(__file__))
CONFIG_FILE = SCRIPT_PATH.parent.parent.parent / "modyn" / "config" / "examples" / "modyn_config.yaml"
````

to infer the actual project directory and avoid any relative path fun

**Review by MaxiBoether — APPROVED** [review]
sorry I just now realized that this is still in draft, but after CI is fixed this mostly LGTM :) thanks for making this a test!

File: modyn/tests/config/test_config_integrity.py (hunk 3 above)
Comment (robinholzi) [reply modyn/tests/config/test_config_integrity.py:11]: Ah yes, sure :)
btw. pathlib supports access parents via `path.parents[depth]`

Comment (MaxiBoether) [reply modyn/tests/config/test_config_integrity.py]: ah thanks, did not know this

File: modyn/tests/config/test_config_integrity.py [hunk 4]
```diff
@@ -7,10 +7,11 @@
...
 If they are directories, all yaml files in the directory will be used."""
 
+PROJECT_ROOT = Path(__file__).parents[3]
+
 MODYN_PIPELINE_CONFIG_PATHS: list[str] = [
-    "modynclient/config/examples/dummy.yaml",
-    "modynclient/config/examples/mnist.yaml",
-    "benchmark/",
+    "benchmark/mnist",
+    "benchmark/wildtime_benchmarks",
```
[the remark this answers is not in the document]
robinholzi: Exactly, iirc they didn't not only fail on this one field that I updated but also the training config had several issue. We can update those configurations when starting to run experiments on those datasets again :)

PR #442, opened 2024-05-23
RFC: How to evaluate pipelines and models

robinholzi:
Outlines our evaluation approach.


robinholzi:
@XianzheMa As we discussed this in person already to some extent, feel free to make any further remarks here in case you have opinions on this! :) (I don't want to remove existing functionality here but provide additional evaluation strategies)

MaxiBoether:
I haven't forgotten about this - I will try to read through it this evening!

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+1) The evaluation metric (e.g. F1, accuracy, precision, recall, etc.)
+2) The evaluation data (set of samples out of the evaluation split used for single model evaluation)
+3) The points in time when the evaluations are run
+
+### Use case 1: online evaluation
+
+In this case we have to be extremely cautious to keep evaluation costs low as the are part of the trigger policy evaluation costs (or at least part of the core pipeline that cannot be disregarded like the experiment model evaluations).
+
+During the core pipeline execution (not including experiment mode pipeline steps), we cannot access future data for evaluation. Also we don't have an evaluation split available. Therefore, we need to use the training data up to the current point in time for evaluation. This is not ideal but it is the only way to evaluate the model in an online fashion.
+
```
XianzheMa, line 56: I didn't totally get it. You mean we cannot use evaluation data at all?
I think we can use evaluation data up to the current point (as long as not access future eval data)

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+> The question of how far into the past we want to includes samples remains. More thought is needed before continuing on this.
+
+### Use case 2: offline evaluation
+
+In experiment mode we can potentially utilize the full evaluation split over the whole dataset time range. Being able to "look into the future" is important to determine the performance of a model.
+
+#### [2] Total pipeline performance
+
+We argue that the performance of pipeline w.r.t. a certain evaluation metric is the weighted average of the performance of the most up to date model at each point in time over the whole time range of the pipeline run.
+
```
XianzheMa, line 66: I am not convinced that we need to do a weighted average. It's also just fine (straightforward and intuitive) to plot a curve with each evaluation in time.

In the end we want to publish a paper. I think the paper reader would prefer a straightforward, easy-to-understand evaluation metric (even if it is a curve), than a higher-order average which she/he has to pause a bit to ponder its meaning. What do you think?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        training4 :a41, after t4, 1y
+        model 4          :a42, after a41, 10y
+
+```
+
+With our assumption of `training_time=0` this `training_<i>` would be can be removed so that we can evaluate one model on the data between two triggers (hence the next training batch)
+
+##### State of modyn (at the time of writing)
+
+We currectly evaluate a model directly after training it (e.g. we are now at `Trigger 4`) and have now robust approach of combining the results of the single model evaluations to a pipeline performance. We support a `OfflineEvaluationStrategy` that specifies the interval in which we want to evaluate the models. We also support `MatrixEvaluationStrategy`: We evaluate on predefined grid of evaluation points.
```
XianzheMa, line 112: You mean Offset?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+We face several problems:
+
+1) We don't know when the next trigger occurs.
+    This might be fine if we assume a time based trigger policy with a regular schedule where we can set the OfflineEvaluationStrategy lookahead offset to the trigger interval. However, for a generic trigger policy we cannot assume this. In consequence when setting a fixed lookahead offset we might end up evaluating a model with a lookahead offset that is too large or too small. (E.g. when always evaluate on the next 5 years after a training. That would mean that if a policy doesn't retrain for 20 years we miss a lot of data)
+
```
XianzheMa, line 118: But evaluation is not necessarily coupled to the training: still every once in a while you can do evaluation on the newly arrived data, right?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+> After `Trigger 4` we e.g. could end up only considering the first 5 years of the time where `model 4` is used. That could completely skew the pipeline performance if e.g. as heavy drift occurs after 5 years of using `model 4`.
+
```
XianzheMa, line 120: Still while using `model 4` to serve, we can evaluate on new data once in a while right? The evaluation interval is not necessarily tied to when the model is trained.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+2) As mentioned above the individual performances of every evaluation cannot simple be averaged to get the pipeline performance as number of samples in every time range is important!
+
+3) Without making any assumptions on how & when to trigger `MatrixEvaluationStrategy` is not of any use as the grid splits are arbitrary and don't correspond to the times when triggers occur and models are transitioned.
+
+4) `OfflineEvaluationStrategy` is not flexible enough to handle arbitrary intervals. Even if we would run after the pipeline where we know the exact points in time when triggers occurred, we could still not specify the intervals as `OfflineEvaluationStrategy` only supports 1-sided intervals. It currently assumes being given `first_timestamp` and `last_timestamp` representing the first and last sample timestamp for the data batch that lead to the trigger. These do not correspond to the start and end time of the model usage.
+
```
XianzheMa, line 126: I think this is just a variable naming thing. You can input the start and end time of the model usage to the strategy, it will as well just return the desired intervals right?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        data6 :e6, after e5, 5y
+        data7 :e7, after e6, 5y
+```
+
+##### Proposed solution
+
+We add a new evaluation strategy with an according pipeline stage that runs evaluation after the rest of the core pipeline finished. By doing so we have access to all trigger times and can evaluate the models in the correct time ranges. After running evaluations on every "triggering-interval" (time between two triggers), we can calculate the pipeline performance as the weighted average of the model performances accounting for the number of samples in the time range.
+
+> Hypothesis: There's no real need to run evaluations directly after the training. We can always do that after the core pipeline finished. If required we can also make use of information like `first_timestamp` and `last_timestamp` from the logs for a specific trigger.
+
```
XianzheMa, line 172: Yes this is correct. I placed it there to have *early feedback*, i.e. without waiting for the whole pipeline to finish, I can already access some of the early model's evaluation results.

XianzheMa (commented):
I cannot comment on all the thoughts as I don't have context on triggering research. I will leave this to @MaxiBoether , but I have two core points:

1. We should talk about the flexibility of *the eval strategy itself* and the flexibility of *where evaluation is conducted* separately. For the latter, yes now it is only at the end of each training, cannot be used in the case if we want to continuously monitor the performance of the model over new data. But for the former, I still argue it is useful: for example, as we use one model to serve, we can keep evaluating its performance over the last 1 month's data from the current timestamp. The current strategy can satisfy this need.
2. I am not convinced with the need to average multiple evaluation scores along the time to form "a single metric". I'd favor plotting a curve as it is easy to understand, and easy to find out where the model performs bad (for example, during a high-drift period).

docs/EVALUATION.md
robinholzi: No, we can definitely run evaluations in that mode. However, in production mode with Modyn, we only use one dataset for training and do not have an additional "testing" or evaluation dataset. When conducting evaluations to provide accuracy feedback into the pipeline (e.g., triggers), we can, as you mentioned, allow the evaluation to use past data. I wanted to point this out to keep it in mind, but for now, I don't intend to focus on this aspect.

docs/EVALUATION.md
robinholzi: Yes, sorry `OffsetEvaluationStrategy`

docs/EVALUATION.md
robinholzi: Conceptually, yes, implementing that is probably rather tricky with the current assumptions in the evaluation logic. Overall that describes the approach of pipeline feedback loop for triggering quite well.

docs/EVALUATION.md
robinholzi: That mainly addresses the question where in the `pipeline_executor.py` code we should run evaluations. Currently, we only evaluate after trainings (and directly after a training we have no information about future triggers etc.).

docs/EVALUATION.md
robinholzi: True, that's only possible after the next training is completed so we have access to the interval bounds. Then all of the interval computation is done in the `pipeline_executor` and not EvaluationStrategy is needed.

docs/EVALUATION.md
robinholzi: Do you use the `early feedback` somewhere during the pipeline run?

docs/EVALUATION.md
robinholzi: That's an important point, thanks for mentioning it!

I think you have something like [this plot in mind (p. 12, Fig. 9 & 10)](https://arxiv.org/pdf/2312.06254), where we have a time-based x-axis with equidistant retraining from a time trigger.

In Fig. 9, we see the different models produced by one pipeline run. 

For your research case of evaluating selection strategy that is totally fair, as you can use static triggering policies and adjust the evaluation strategies to this fixed schedules.

There are a few issues with this approach when comparing trigger policies though.

1) When comparing several trigger policies, we cannot rely on equidistant triggering timestamps, as different pipeline runs could have evaluations centered around different points in time. The `MatrixEvaluationStrategy` partially addresses this but does not focus on evaluating models only within the range where they are used for online inference.

2) We don't want to evaluate every model on all data (which is obvious) because it doesn’t matter how a model trained at the beginning of a dataset performs at the end since it will have been replaced by new models. Plotting this information adds noise without providing value when comparing different pipeline runs with different trigger policies. Cleaning up the data and combining it into one accuracy line is possible for a single dataset on one pipeline run, but comparing multiple pipeline runs is essential for Modyn. Thus, we need a method that allows for quick iterations on pipeline settings and is extensible.

3) Essentially, one can consider a pipeline run as a higher-order model that uses different trained models over time, resulting in an accuracy time series where one pipeline has one accuracy at a time. This time series for a fixed pipeline run is essentially an ensemble of the times series for the single models. We could consider the pipeline accuracy at a fixed time `t` as the accuracy of the model that is 'active' at time `t`.

4) Regarding your point on averaging, when comparing multiple pipeline runs with different trigger policies, we have at least three key dimensions to consider in plots: (dataset) time, cost (proxied by compute time), and evaluation metrics/performance (e.g., accuracy). To compare trigger policies based on their cost/accuracy trade-off, plotting 3 dimensionally is confusing therefore we need to aggregate over time. We could use a scatter plot with cost on the x-axis and accuracy on the y-axis, with one point per policy. This approach requires a scalar measure of accuracy for a pipeline.

robinholzi: Overall, it would probably be possible to do everything with the current model evaluation strategies. I'm quite certain though that this would yield really intertwined code that is hard to adjust and extend.

docs/EVALUATION.md
XianzheMa: I see thanks!

docs/EVALUATION.md
XianzheMa: I see. Thanks for providing extra context! I will leave the discussion on how to eval in the triggering setting to you and Maxi

docs/EVALUATION.md
XianzheMa: What's the assumption that lets implementing this tricky? I think we just need to refactor "where to evaluate", and uses an `OffsetEvalStrategy` to evaluate on the past predefined window of data.

docs/EVALUATION.md
XianzheMa: I just retrieve the pipeline log before the pipeline finishes. Then I can look into the data to ensure no bug is going on.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        model 6          :a6, after a5, 5y
+        Trigger7 :milestone, :after a6
+        model 7          :a7, after a6, 5y
+```
+
+## Evaluation methodology
+
+An evaluation methodology consists of three main components:
+
+1) The evaluation metric (e.g. F1, accuracy, precision, recall, etc.)
```
MaxiBoether, line 47: nit: we can be interested in multiple metrics (different types of f1, top-1 accuracy, top-5 accuracy etc all together)

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+2) The evaluation data (set of samples out of the evaluation split used for single model evaluation)
+3) The points in time when the evaluations are run
+
+### Use case 1: online evaluation
+
+In this case we have to be extremely cautious to keep evaluation costs low as the are part of the trigger policy evaluation costs (or at least part of the core pipeline that cannot be disregarded like the experiment model evaluations).
+
+During the core pipeline execution (not including experiment mode pipeline steps), we cannot access future data for evaluation. Also we don't have an evaluation split available. Therefore, we need to use the training data up to the current point in time for evaluation. This is not ideal but it is the only way to evaluate the model in an online fashion.
```
MaxiBoether, line 55: One thing we can do it build up a validation set on the fly instead of using the training data. This means during execution the selector is not informed about everything but let's say 90% of the data, and in some other database we store the information on the validation set. Of course we should only do this in the online evaluation setting (which simulates the real world and we want to find out whether we can use model performance as a triggering indicator) but this would avoid using the training data itself.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+> The question of how far into the past we want to includes samples remains. More thought is needed before continuing on this.
```
MaxiBoether, line 57: From my understanding, this is basically another triggering policy type (besides drift, time, amount) that we could compare to those triggering policies, which you potentially do in your thesis. In the comparison of those policies, we could then again fall into use case 2 again. What I am trying to say is that the purpose of the evaluation itself is two fold, it can be used actively in the pipeline (1) and to compare policies (2). When using it in the pipeline, we can still compare it, so they are not mutually exclusive. Does this make sense?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Use case 2: offline evaluation
+
+In experiment mode we can potentially utilize the full evaluation split over the whole dataset time range. Being able to "look into the future" is important to determine the performance of a model.
```
MaxiBoether, line 61: This part of evaluation is _post hoc_ - i.e., we use it to gain insights on policies in an experimental setting (for a thesis, paper, etc) but we cannot use this in an online fashion as you mention above.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+#### [2] Total pipeline performance
```
MaxiBoether, line 63: sorry very nit but I am a little confused here - is the [2] a leftover?

MaxiBoether: actually there comes a [1] later on. I think I am missing some context on what you want to discuss here

docs/EVALUATION.md
MaxiBoether: I think all of these points are very valuable and should go into the document. In particular, before diving into the argument that we need to average, I think in this document we should explain that for the comparison of triggering policies, we need to reduce dimensionality and need to compress model performance into a scalar metric. In order to do this, we somehow need to reduce multiple accuracies (accuracy being any metric) that we collect into a single metric. I think we might want to start out by describing more background here in order to get a common understanding.

We get `n` model versions over time and evaluate those on `m` evaluation datasets (can be time slices of the dataset). So overall we have `n * m` values and need to reduce those. Basically it would be helpful to structure this section a bit differently, lay out what is happening and how we then need to process this data for your comparison. I think we explicitly want to talk about (what you also mention in your comments here) comparing selection policies with a common time trigger where we can plot the accuracies over time, and talk about comparing pipelines separately. Does this make sense?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+We argue that the performance of pipeline w.r.t. a certain evaluation metric is the weighted average of the performance of the most up to date model at each point in time over the whole time range of the pipeline run.
```
MaxiBoether, line 65: I don't think it's fully clear to me yet what exactly you want to average weightly. At each step in time, we have exactly _one_ latest model for which we can get `m` evaluation metrics (see my comment below). Do you want to aggregate those `m` values per model, do you want to aggregate all models (`n*m` values? if so, in what sense is this weighted)?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+Mind, that we need to weight every model performance at a point in time by the number of samples within that time range.
```
MaxiBoether, line 67: I would frame it a bit differently: One way to reduce dimensionality (as mentioned above) is to evaluate a model on one window of data (so `m=1` with the definition above). If I udnerstand "number of samples within that time range" correctly, you propose to set `m=1` but I might be missing something. Hence I would propose to restate this a bit with the most general case of having `n` models (where at each timestep, one model is the latest model) and each of those model can be evaluated on `m` datasets (e.g., we can have one model for each year and evaluate each model on each year's evaluation set, leading to the evaluation matrix. Or we can have one model for each year, and evaluate it on the past year, the current year, and future year, i.e. having `m=3`)

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+> E.g. in `model 3` from the Gantt diagram above: assuming our dataset sample distribution is so uneven that 90% of the samples are in the time range between `Trigger3` and `Trigger4`, then the performance of `model 3` should have a higher weight in the pipeline performance than the performance of `model 2`.
```
MaxiBoether, line 69: I am not 100% sure yet we need to weight by number of samples used for training. Just because we have more samples to train a model does not necessarily imply for me its performance is more important. I'd argue I am always equally interested that at whatever timestep I use the current model, it performs well.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+It can be argued that the model performance of the newest model in the time span between the ends of two training processes is measured well by the model performance (evaluation result) when evaluating with using all evaluation split samples between the two triggers.
+
+> **Remark/Assumptions**: Instead of considering the time between two model training completions, we will go forward with the time between two triggers (start of training processes). These are not exactly the same, as we would still use the old model while a training is going on (see diagram below). However, we argue that the difference is negligible as (1) we don't expect much drift between the samples directly after the trigger and after the training (close in time). And (2) compared to the time between two triggers, the duration of a training process is rather small and therefore doesn't have a big impact on the evaluation. Note: We do not evaluate on training samples in this model! We can adjust this assumption later on if needed by incorporating the training time into the evaluation strategy.
```
MaxiBoether, line 73: I agree. I like that you often keep the entire pipeline in mind (e.g., triggers might not actually be equidistant in yearbook) but sometimes it helps to make simplifiyng assumptions first. I think it's completely ok for now to e.g. in CLOC/arXiv/... plot time triggers equidistantly on the x-axis, if that helps.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+```mermaid
+gantt
+    title Pipeline run
+    dateFormat YYYY-MM-DD
+    axisFormat %Y
+    section models
+        Trigger1 :milestone, t1, 1930, 0d
+        train_batch_1 :a10, 1928, 2y
```
MaxiBoether, line 87: what exactly do you mean by train batch 1? it is different from training. I thought it means training on the first batch of data but I might not fully get the gantt chart yet

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        training1 :a11, after t1, 1y
+        model 1          :a12, after a11, until a22
```
MaxiBoether, line 89: this means that is the currently "active" model right?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        model 3          :a32, after a31, until a42
+
+        Trigger4 :milestone, t4, 1950, 0d
+        train_batch_4 :a40, after a30, until t4
+        training4 :a41, after t4, 1y
+        model 4          :a42, after a41, 10y
+
+```
+
+With our assumption of `training_time=0` this `training_<i>` would be can be removed so that we can evaluate one model on the data between two triggers (hence the next training batch)
```
MaxiBoether, line 108: Ah now I get even more what you mean. Yes I think that is actually preferred from an "experimental perspective": In the CLOC case in the paper you linked in the discussion with Xianzhe, even though the 2010 model needs some time to train, I'd not use the 2009 model as the most recent model. I'd use the 2010 model for evaluating current performance on all future data.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+We face several problems:
+
+1) We don't know when the next trigger occurs.
+    This might be fine if we assume a time based trigger policy with a regular schedule where we can set the OffsetEvaluationStrategy lookahead offset to the trigger interval. However, for a generic trigger policy we cannot assume this. In consequence when setting a fixed lookahead offset we might end up evaluating a model with a lookahead offset that is too large or too small. (E.g. when always evaluate on the next 5 years after a training. That would mean that if a policy doesn't retrain for 20 years we miss a lot of data)
+
+> After `Trigger 4` we e.g. could end up only considering the first 5 years of the time where `model 4` is used. That could completely skew the pipeline performance if e.g. as heavy drift occurs after 5 years of using `model 4`.
+
+2) As mentioned above the individual performances of every evaluation cannot simple be averaged to get the pipeline performance as number of samples in every time range is important!
```
MaxiBoether, line 121: But this is "post processing" anyways, isn't it? I.e. we collect all the metrics and posthoc for analyzing a pipeline run get the aggregate performance

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+3) Without making any assumptions on how & when to trigger `MatrixEvaluationStrategy` is not of any use as the grid splits are arbitrary and don't correspond to the times when triggers occur and models are transitioned.
```
MaxiBoether, line 123: I'd not necessarily agree here. Let's consider CLOC again. The simple case is that we have the `n*m` score matrix where we trigger each year and evaluate on each year. Now, let's consider e.g. a drift policy where we don't know before when we trigger. Let's say we trigger x times. We can still build up a `x*m` matrix, that tells us for each produced model, how well it performs on each year?

And if this is not helpful, imagine a plot like in the Modyn paper. On the x-axis we take the evaluation year and on the y axis, we plot the accuracy of the _current_ model that is most recent for each year on the x-axis. This can be extracted from the matrix easily. If the drift strategy triggers a lot less frequently than yearly triggering, but has a similar accuracy curve, we'd easily see that it can be deemed "better".

Overall, I think having _fixed evaluation time slices_ actually is wanted and useful, because it gives a common reference point for all types of triggering strategies where we can ask: How well is the current most recent model in the pipeline performing (by whatever metric)?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+4) `OffsetEvaluationStrategy` is not flexible enough to handle arbitrary intervals. Even if we would run after the pipeline where we know the exact points in time when triggers occurred, we could still not specify the intervals as `OffsetEvaluationStrategy` only supports 1-sided intervals. It currently assumes being given `first_timestamp` and `last_timestamp` representing the first and last sample timestamp for the data batch that lead to the trigger. These do not correspond to the start and end time of the model usage.
+
+>>> The ideal solution would therefore evaluate models on their entire usage period with all samples in that period.
```
MaxiBoether, line 127: I think that is _one_ evaluation policy, but I might be wrong. I don't think we should remove e.g. the evaluation on the matrix for the reasons mentioned before. Yes, the performance of a model while it is being used/most recent is probably the most interesting metric, but for experiments, evaluating the model from year 2006 on data from 2010 might still be interesting.

docs/EVALUATION.md
MaxiBoether: yes there is a tradeoff here. In some sense it would be amazing to overlap more things in Modyn's execution model: We could overlap training a new model, while evaluation is still going on (assuming enough GPU resources etc), we could run evaluations in parallel given multiple GPU resources and whatnot. Right now Modyn's execution model is pretty sequential, but I agree nevertheless early evaluation is helpful to investigate the logs early.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        data3 :e3, after e2, 5y
+        data4 :e4, after e3, 25y
+        data5 :e5, after e4, 25y
+        data6 :e6, after e5, 5y
+        data7 :e7, after e6, 5y
+```
+
+##### Proposed solution
+
+We add a new evaluation strategy with an according pipeline stage that runs evaluation after the rest of the core pipeline finished. By doing so we have access to all trigger times and can evaluate the models in the correct time ranges. After running evaluations on every "triggering-interval" (time between two triggers), we can calculate the pipeline performance as the weighted average of the model performances accounting for the number of samples in the time range.
```
MaxiBoether, line 169: I agree that evaluating at the end as some form of post processing is helpful.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+> Hypothesis: There's no real need to run evaluations directly after the training. We can always do that after the core pipeline finished. If required we can also make use of information like `first_timestamp` and `last_timestamp` from the logs for a specific trigger.
+
+#### [1] Model performance at a certain point in time
```
MaxiBoether, line 173: ok i am a bit confused with the structure here. What are we thinking about now? I think I mentioned it above, I think a little restructuring in a second iteration would help super duper much to align :)

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+We assume the performance of a model changes continuously over time and not at discrete points in time (like the trigger times).
+
+As we want to be able to plot the performance (i.e. accuracy) over the time range of the pipeline run, we have to observe the performance at a higher precision than [2] demands.
```
MaxiBoether, line 177: I'd say take it step by step. I am already happy with plotting evaluation performance like in Figure 9 in the Modyn paper (for now). Of course I understand for the retraining policy we need finer grained accuracies - but for analysis, actually I think we have a lot of flexibility by time slicing our evaluation dataset.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+Plotting with evenly spaced triggering indexes as x-axis (worst case is barplot where everything between two triggers is aggregated) is not sufficient as it disregards performance change in the potentially long intervals between triggers.
+
+Otherwise we could only plot a horizontal bar for every model evaluation at a trigger time. This would not allow us to see how the performance changes over time within one triggering interval. This information is important to fairly evaluate a trigger policy though as it could happen that a severe data drift causes rapid performance degradation. If the model wouldn't trigger for some time though we want to see that in the plot! Only plotting the total performance over the time range would hide this information.
+
+We therefore want to evaluate the model at multiple points in time within the triggering interval. We could define a configuration option that would allow us to either set the number of evaluation steps between two triggers or the time interval or number of samples between two evaluations.
```
MaxiBoether, line 183: Can't we just timeslice the evaluation dataset differently? 

Although I see that in yearbook e.g. the timestamps themselves are on a yearly basis. Then again, I wonder what is the benefit of finer grained accuracy evaluation since for all benchmark datasets I am aware of, we have the timestamps on a certain granularity, and our evaluation already runs on that granularity (e.g., years). I am not sure about CLOC (there we could have finegrained timestamps, need to check).  Hence, I am not sure if you want to spend time on this (I might be missing something here though). If we'd have very finegrained timestamps maybe... it really depends on what you focus on I think. As I mention above, what I like about the yearly aggregation is that we have a common reference point for different triggering policies: For each year, given whatever is the most recent model, what is its performance on that timeframe? That allows us to compare all policies on a shared x axis. I am not sure how that would be possible with an non-evenly spaced x-axis.

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+        data4.2 :e4.2, 1951, 6y
+        data4.3 :e4.3, 1955, 6y
+        data4.4 :e4.4, 1959, 6y
+        data4.5 :e4.5, 1963, 6y
+        data4.6 :e4.6, 1967, 6y
+```
+
+### Open Question
+
+1) With the approach of evaluating a model only on future data (where the model is eventually used), it might even be fair not to use a evaluation split as the samples used for evaluation are never part of the training data of a particular model. Agreed?
```
MaxiBoether, line 247: If I understand you correctly, this is one particular evaluation strategy you are talking about (evaluate model trained on data up to 2010 on data from 2011 for example). In this case yes, we could use the data we will train the 2011 model on, but I wonder if this will not complicate things, since we do have the nice timeslicing logic of the evaluation dataset. Why not just evaluate(2010 model, (2011, 2012]) or something?

docs/EVALUATION.md
```diff
@@ -0,0 +1,249 @@
...
+
+2) What behavior do we desire when we train multiple models on the same timestamp (e.g. if we have a lot of data at a certain point in time). One option for plotting would be to weight average the different according to how many samples of this point in time were covered by the respective model. Considering the samples that are used to evaluate the models we cannot different the evaluation data of two training runs on the same timestamps (even if their batches might contain different samples)
```
MaxiBoether, line 249: This can occur when we don't use time triggering, right? TBH a solution does not immediately come to mind, I need to think about it.

MaxiBoether (changes requested):
Thanks Robin, this is very nice to agree on a common understanding. I think on some things we might have a different perspective. I need more time to deeply think about your Gantt charts and read everything again, but I think it would help to address my comments/thoughts that I added first. I was a bit unsure about the structure of the doc (with the [1] vs [2] thing). Also, I think we need to lay out some more common context on how evaluation in the general case can work (`n` models can be evaluated on `m` evaluation slices, as I mention in several comments). Also, overall, I still have this comparison of triggering policies in mind where we have a common x axis, the y axis is performance on that timestamp (so we share a common evaluation set, like each year in CLOC), and then we can plot the performance (however aggregated into a scalar) over time and the overall pipeline cost. I am not sure this is covered in the doc. I am not sure whether you also suggest having a common evaluation reference point (e.g. list of years like in CLOC) or just evaluate between triggers (which would differ between policies!). Let's try to align on those things.

Sorry again this took some time to read. Now that I have read it I can be faster in responding :D

docs/EVALUATION.md
robinholzi: Sure, I think all options we are currently discussing support that :)

docs/EVALUATION.md
robinholzi: Agreed, currently, this is not implemented though

docs/EVALUATION.md
robinholzi: thanks for your thoughts @MaxiBoether! I will refine the document to give more background. Currently, there are, indeed, quite a lot of unstated assumptions.

docs/EVALUATION.md
robinholzi: In the whole document, I only looked at one arbitrary (but fixed) evaluation metric. When evaluating a pipeline with multiple metrics, we can just plot a different time series (and encode this 4th dimension with a color / or do a separate plot if it gets too messy). I'll integrate the idea of multiple evaluation metrics into the doc!

docs/EVALUATION.md
robinholzi: Okay, what I intended to do there is to not assign the same weights to a model that was used for 1 day to one that was used for 1 year (without retrainings). Using weighting by the time a model was actively used, captures this idea better!

docs/EVALUATION.md
robinholzi: That represents the interval between the sample with the smallest and biggest timestamps in the batch that is handed to the trainer sever for a new training. I assumed that retraining is only performed on the new samples since the last training, which does not hold for all sampling strategies. I will rename that.

docs/EVALUATION.md
robinholzi: That's exactly my line of thought! It surely depends how close we want our experiments to resemble the "production" case (in which you obviously cannot run inference with a model unless its training is finished). I think this simplifying assumption is fair here though.

docs/EVALUATION.md
robinholzi: Exactly!

docs/EVALUATION.md
robinholzi: I have considered that already and it makes sense, yes! I wasn't able to judge how computationally expensive it would be to evaluate all models on all slices (obviously that depends on the number of slices).
In addition to that I had a slight modification to this idea in mind. Instead of using partitions of the dataset ("slices"), it might make sense to use some kind of rolling time window (with the evaluation time in the window's center). By doing we could estimate the performance at a point `t` by considering `offset` years into the past and future. The difference to slicing is that those windows can overlap or fill time sparsely.

docs/EVALUATION.md
robinholzi: Agreed, let's keep the matrix evaluation!

docs/EVALUATION.md
robinholzi: Sure, that's possible. For smaller datasets it's beneficial though to have access to as much data as possible when evaluating.

docs/EVALUATION.md
robinholzi: Yes! For sample counting policies that would occur when a dataset has unusually high sample density at one point in time. Drift policies can also produce this. For now, I don't think this is too much of a concern as we can simply plot several accuracy points at one x value. Let's just try it out first and see how often it occurs and if it's worth thinking more thoroughly.

docs/EVALUATION.md
MaxiBoether: Thanks!

docs/EVALUATION.md
MaxiBoether: Yes, I guess we have three time axes here: What do we currently support, what is valuable and should be implemented soon, and what is nice to have but we might do it in the future

docs/EVALUATION.md
MaxiBoether: Actually I was not talking about multiple metrics here. To clarify: With `m` metrics, I meant something like the matrix approach:   we evaluate each model m times (okay, then we can collect multiple metrics, but I was also thinking about a single metric like accuracy).

MaxiBoether: so it was not clear to me -- assuming a single metric -- what to want to average. Great to include the multiple metric stuff, but my main concern is multiple metrics in the sense that we evaluate the 2010 model on 2006, 2007 etc, not that we get F1, Accuracy etc for the 2010 model

docs/EVALUATION.md
MaxiBoether: Agreed. It depends on the dataset and scenario. I don't think there is one best solution for what evaluation data to use (or even how to sample an evaluation dataset - should it be e.g. balanced across time such that for years where we have less data we also have less evaluation data, or sampled uniformly at random, or with fixed size across years?)

docs/EVALUATION.md
MaxiBoether: Yes, for sure we can extend it - as long as we somehow are able to express the original idea, that's fine :D

MaxiBoether:
@robinholzi LMK when you went through this and deem it ready for a final pass of review. Some stuff might've changed due to SIGMOD discussions etc, so let's make sure it's consistent with the submission :)

robinholzi:
> @robinholzi LMK when you went through this and deem it ready for a final pass of review. Some stuff might've changed due to SIGMOD discussions etc, so let's make sure it's consistent with the submission :)

Sure, I'll take a look at it tonight :)

robinholzi:
@MaxiBoether I updated the evaluation markdown. Generally, it was already quite okay, I removed some open questions that we now fully incorporated into the code base. Overall, the doc is not perfectly structured, but imho it serves its purpose of onboarding new people with enough technical details to understand our decisions. For the higher level ideas, the SIGMOD paper a / my thesis should be the first resource to consult.

docs/EVALUATION.md
```diff
@@ -0,0 +1,372 @@
+# Modyn pevaluation framework
```
MaxiBoether, line 1:
```suggestion
# Modyn evaluation framework
```

docs/EVALUATION.md
```diff
@@ -0,0 +1,372 @@
...
+
+##### Dataset slicing (`MatrixEvaluationStrategy`)
+
+Partitioning the dataset(s) w.r.t. to time at sufficiently small intervals will allow us to extract a time series after running evaluations for every (slice, model, metric) combination.
+
+When plotting this there are several cases:
+
+- When our slices already correspond to the minimal resolution of the samples times (e.g. every year), we don't need to think about interpolation and can just connect the measurement points with lines.
+
+- When we have more continuous data where we cannot evaluate at every point in time, can either plot vertical lines (same value for a certain time range) or interpolate. In the latter case it's fair to plot the points for each evaluation interval at the interval center and connect the points with lines (linear interpolation).
```
MaxiBoether, line 102: One thing related to this I believe is the whole patching stuff. I wanted to discuss this with you anyways again. Even for SIGMOD, we still needed to patch the logs for CGLM, but not anymore for yearbook, if I remember correctly. There was this problem with the start timestamp being 0 and having triggers. We should think about the fundamental issue here again, and at least document it (also in the README of the datasets that are affected, but also here). Can we solve this somehow? It has been some weeks, I don't remember the details right now with all the other stuff going on, but in the best case, we could avoid patching at all. Maybe we can discuss this in our 1:1 this week also, if that's easier

docs/EVALUATION.md
```diff
@@ -0,0 +1,372 @@
...
+
+- time
+- evaluation metric value
+- pipeline run
+- cost
+
+For a cost/performance tradeoff comparison there are two useful plotting strategies:
+
+<details>
+  <summary><b>Plot 2 + Plot 3 (stacked barplot, x=time, y=cost | eval metric value (ensemble of all models), color=pipeline)</b></summary>
```
MaxiBoether, line 207: i don't fully understand the Plot numbering here yet. Is this just "notes" in some sense? Did you plan to add screenshots here?

docs/EVALUATION.md
```diff
@@ -0,0 +1,372 @@
...
+When analyzing how a single pipeline run performed over time, we can line-plot the evaluation results over time using time as the $x$ dimension, the evaluation metric value as the $y$ dimension and the model as a color coded dimension. The different evaluation metrics can be plotted in different plots using facets.
+
+Using this plot we expect to see that every model performs best at the interval it was trained in(, and hopefully also reasonably well in the interval the model was served in). Accuracy should decay in both directions from the training interval.
+
+There are two main approaches to acquire those time series of evaluation results:
+</details>
+
+##### Dataset slicing (`MatrixEvaluationStrategy`)
+
+Partitioning the dataset(s) w.r.t. to time at sufficiently small intervals will allow us to extract a time series after running evaluations for every (slice, model, metric) combination.
```
MaxiBoether, line 96:
```suggestion
Partitioning the dataset(s) w.r.t. time at sufficiently small intervals will allow us to extract a time series after running evaluations for every (slice, model, metric) combination.
```

MaxiBoether (changes requested):
Thank you, Robin. My major comment here is that we should discuss the need for patching as well in this docs and in the dataset documentation. We should also try and understand again the root cause of patching in our current model, maybe we can discuss this in our 1:1. Also, for those plots, not sure if you planned to add screenshots or not due to the numbering.

docs/EVALUATION.md
robinholzi: I thought we had those issues solved with the latest revision on main. However, when I rerun some of the experiments for my thesis (with the newer trigger variants), I will see if they are still present and potentially address them.

docs/EVALUATION.md
robinholzi: I didn't plan to add screenshots. Here the numbers were just used as identifiers to refer to certain plots from different sections.

robinholzi: Added some sample plots.

docs/EVALUATION.md
MaxiBoether: @XianzheMa do you remember which pipelines still need to be patched? CGLM right?

robinholzi: Okay, thanks for the reminder and the explanation! It would be good if one of you @MaxiBoether @XianzheMa can send me a recent pipeline log file so I can inspect the issue in detail.
Generally, I don't want to include these patching things in global documentation, but I want to fix the business logic so it doesn't require patching anymore. 
That's why I propose to merge this PR without mentioning patching.

PR #443, opened 2024-05-23
Add missing `model.eval()` during BtS downsampling

XianzheMa:
In BtS downsampling, eval mode is not turned on but in StB downsampling it is turned on. This PR turns on eval mode in BtS downsampling.

### DIFF
--- a/modyn/trainer_server/internal/trainer/pytorch_trainer.py
+++ b/modyn/trainer_server/internal/trainer/pytorch_trainer.py
@@ -650,17 +650,16 @@ class PytorchTrainer:
         assert self._downsampler is not None
         assert self._downsampling_mode == DownsamplingMode.BATCH_THEN_SAMPLE
         self._model.model.eval()
-        with torch.no_grad():
-            self._downsampler.init_downsampler()
-            self.start_embedding_recording_if_needed()
-            big_batch_output = self._model.model(data)
-            embeddings = self.get_embeddings_if_recorded()
-            self._downsampler.inform_samples(sample_ids, big_batch_output, target, embeddings)
-            self.end_embedding_recorder_if_needed()
+        self._downsampler.init_downsampler()
+        self.start_embedding_recording_if_needed()
+        big_batch_output = self._model.model(data)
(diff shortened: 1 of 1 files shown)


MaxiBoether:
I don't think this is correct. Some downsampling methods rely on the gradient (norm), e.g., Craig. If we don't do the forward pass with grad enabled, they won't work anymore

XianzheMa:
> I don't think this is correct. Some downsampling methods rely on the gradient (norm), e.g., Craig. If we don't do the forward pass with grad enabled, they won't work anymore

thanks for pointing it out! I removed `no_grad()`. But `model.eval()` should be added as it's already added in StB mode.

I think later I can still add `no_grad()` depending on if the downsampler requires grad or not (to improve efficiency), but for now I will leave it as is.

XianzheMa:
> I don't think this is correct. Some downsampling methods rely on the gradient (norm), e.g., Craig. If we don't do the forward pass with grad enabled, they won't work anymore

Thanks for point it out! I removed `no_grad` but kept the `model.eval()`

MaxiBoether (approved):
LGTM now! Some downsamplers actually already use no_grad: https://github.com/eth-easl/modyn/blob/e7d9198e2bae4ccad61fc4fc2ec52f4ec96024ec/modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py#L40

So it's the downsamplers task to do this correctly :)

XianzheMa:
> LGTM now! Some downsamplers actually already use no_grad:
> 
> https://github.com/eth-easl/modyn/blob/e7d9198e2bae4ccad61fc4fc2ec52f4ec96024ec/modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py#L40
> 
> So it's the downsamplers task to do this correctly :)

I think that's not related. The most expensive part of not turning on `torch.no_grad` is the forward pass, because we need to store the activations and track the graph. Ideally we can ask if the downsampler needs the grad or not, if it is not needed, then we turn on `torch.no_grad` during forward pass.

But I want to prioritize implementing RHO-LOSS now.

PR #446, opened 2024-05-25
fix: Fix linting, integrate `ruff` and typo check `pre-commit` checks

robinholzi:
# Motivation

After realizing that our current isort setup is misfunctional and doesn't do anything, I started integrating ruff via pre-commit hooks alongside some other useful tools like spell checkers and file line ending sanitizers.

I tried to keep the change as small as possible and therefore disabled some ruff rules (like the ruff `isort` equivalent, ruff format, line ending cleanup, ...). Those can be enabled at a point in time when no PRs as it could yield quite some merge conflicts.

In theory, the `python_comliance` script could be replaced completely by pre-commit hooks, however, in this PR I didn't want to waste too much time on CI adjustments. I just want to have the pre-commit hooks for my sanity.

Avoiding typos also makes sense I guess :)


.clang-tidy
```diff
@@ -82,10 +82,10 @@ Checks: >
 # openMP                                                doesn't apply (we're not using OpenMP)
 # zircon                                                utility checks that would need configuration
 
-## Aliasses
```
robinholzi, line 85: These changes are all made automatically. I will comment on the things I changed manually here.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
+ci:
```
robinholzi, line 1: Manual setup of pre-commit hooks (some still disabled to keep diff small for now)

_typos.toml
```diff
@@ -0,0 +1,7 @@
+
+[default]
+extend-ignore-re = ["(?Rm)^.*# spellchecker:disable-line$"]
+
+[default.extend-words]
+strat = "strat"
```
robinholzi, line 6: Those two words are recognized as typos. As they occur quite often we whitelist them

experiments/criteo_online_dataset/gen_pipelines.py
```diff
@@ -122,30 +122,36 @@ def label_transformer_function(x: torch.Tensor) -> torch.Tensor:
     data_points_for_trigger: 30000000
```
robinholzi, line 122: I think these were changes by black, as we run it on more directories now.

modyn/tests/evaluator/internal/dataset/test_evaluation_dataset.py
```diff
@@ -1,4 +1,7 @@
 # pylint: disable=unused-argument, no-name-in-module, no-value-for-parameter
+# ruff: noqa: N802  # grpc functions are not snake case
```
robinholzi, line 2: I made those annotations manually, we already have equivalent ones for pylint

scripts/python_compliance.sh
```diff
@@ -1,60 +1,60 @@
-echo "Running compilance check"
+echo "Running compliance check"
 
 PARENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd )"
 
 pushd $PARENT_DIR
 
-if ! command -v mamba &> /dev/null
+if ! command -v micromamba &> /dev/null
```
robinholzi, line 7: In case someone is still using mamba and not micro mamba, I guess forcing them to upgrade isn't to harmful :)

setup.cfg
```diff
@@ -26,13 +26,13 @@ extend-ignore = E203
 
 [isort]
 profile = black
-src_paths = ["src"]
+src_paths = ["src"]  # TODO: isort actually does nothing. When enabling it again, let's transition to ruff
```
robinholzi, line 29: Setting that motivated the PR

robinholzi: The underlying issue is not addressed here though, as almost every file would have a diff. Waiting for a time where we wouldn't cause merge conflicts.

setup.cfg
```diff
@@ -26,13 +26,13 @@ extend-ignore = E203
...
 line_length = 120
 skip_gitignore = true
 extend_skip_glob = ["**/generated/*", "**/*.pyi", "**/*_grpc.py", "**/*_pb2.py"]
 
 [mypy]
-python_version = 3.9
+python_version = 3.11
```
robinholzi, line 35: Hasn't been updated for some time now.

benchmark/wildtime_benchmarks/data_generation_huffpost.py
```diff
@@ -22,7 +22,7 @@ class HuffpostDownloader(Dataset):
     time_steps = [i for i in range(2012, 2019)]
     input_dim = 44
     num_classes = 11
-    drive_id = "1jKqbfPx69EPK_fjgU9RLuExToUg7rwIY"
```
MaxiBoether, line 25: That should not change

black.toml
```diff
@@ -1,14 +1,17 @@
...
+extend-exclude = '''
+(
+    .*\.pyi
+  | .*_grpc\.py
+  | .*_pb2\.py
+  | .*/generated/
+  | benchmark/
+  | plotting/
+  | build/
+  | libbuild/
```
MaxiBoether, line 14: Did you change this on purpose? I remember at some point I needed the local wildcards, but I am not sure whether this is from the pre commit hook or done manually

modyn/metadata_processor/processor_strategies/processor_strategy_type.py
```diff
@@ -13,7 +13,7 @@ class ProcessorStrategyType(Enum):
     BasicProcessorStrategy = "basic_processor_strategy"  # pylint: disable=invalid-name
 
 
-class InvalidProcessorStrategyTypeException(Exception):
+class InvalidProcessorStrategyTypeError(Exception):
```
MaxiBoether, line 16: Why this change?

modyn/model_storage/internal/grpc/model_storage_grpc_servicer.py
```diff
@@ -1,5 +1,7 @@
 """Model storage GRPC servicer."""
 
+# ruff: noqa: N802  # grpc functions are not snake case
```
MaxiBoether, line 3: Why is this only added for the model storage grpc servicer and not the other servicers? (Some leftover from removing ruff changes from this commit?)

.pre-commit-config.yaml
MaxiBoether: I am a bit torn on pre commit hooks. The reason we had a script + CI checks for linting is that sometimes, I have had the situation that black's autoformat completely destroyed my file. Typically, I do a commit after doing changes and then another commit (or multiple) for linting, such that there is a clean separation between linting and my stuff. If I understand it correctly, now we would enforce ruff being run before we can commit.  Or is it possible that we get a commit without the autoformatting + then a follow up commit with autolinting applied?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  autoupdate_commit_msg: "chore: Update pre-commit hooks"
+exclude: ^(\.data/|\.scratch/)
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: "v0.4.1"
+    hooks:
+      - id: ruff
```
MaxiBoether, line 8: In what relationship is ruff to the other tools? Which tools will be replaced (black, autopep, mypy etc) and which will stay?

scripts/python_compliance.sh
MaxiBoether: Then we'd need to update some READMEs as well. I'd actually prefer a brief logic that uses micromamba preferably, but if it does not exist checks for mamba to not break everyone's setup (again after the conda switch :D). We can of course emit a warning/hint to get micromamba.

MaxiBoether (changes requested):
Thank you! This mostly looks good, but I had some smaller questions and comments here and there, especially on using pre commit hooks and micromamba.

.pre-commit-config.yaml
robinholzi: I suggest removing isort and flake8 as those tools are almost fully covered. Black can arguably also be removed without too many concerns, but as black is fairly fast, it doesn't hurt too much (they are compatible). Pylint does some things differently, I'd keep it for now. On a greenfield project, I'd rather not use it though.

----

https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black

### Is the Ruff linter compatible with Black?

Yes. The Ruff linter is compatible with [Black](https://github.com/psf/black) out-of-the-box, as long as the [line-length](https://docs.astral.sh/ruff/settings/#line-length) setting is consistent between the two.

### How does Ruff's linter compare to Pylint?

At time of writing, Pylint implements ~409 total rules, while Ruff implements over 800, of which at least 209 overlap with the Pylint rule set (see: [#970](https://github.com/astral-sh/ruff/issues/970)).

### How does Ruff's linter compare to Flake8?

Ruff can be used as a drop-in replacement for Flake8 when used (1) without or with a small number of plugins, (2) alongside Black, and (3) on Python 3 code.

Ruff can also replace Black, isort, yesqa, eradicate, and most of the rules implemented in pyupgrade.

### How does Ruff's formatter compare to Black?
The Ruff formatter is designed to be a drop-in replacement for [Black](https://github.com/psf/black).

Specifically, the formatter is intended to emit near-identical output when run over Black-formatted code

### How does Ruff compare to Mypy, or Pyright, or Pyre?
Ruff is a linter, not a type checker.

.pre-commit-config.yaml
robinholzi: `git commit -n` skips the pre-commit hooks. Is that what you are searching for? I also use it rather often locally when making a temporary snapshot of my code. If you don't like pre-commit hooks you can just not install them them :) We just have to define what the CI does and take that as source of truth.

MaxiBoether: Ahh, I see. If we can easily skip the pre-commit hooks, then it sounds great!

benchmark/wildtime_benchmarks/data_generation_huffpost.py
robinholzi: Oh snap, good catch! Looks like a false positive in the typo checker

black.toml
RobinHolzingerQC: I changed it on purpose, but we can surely restore the local wildcards.

modyn/metadata_processor/processor_strategies/processor_strategy_type.py
robinholzi: pep8 suggests this: see https://docs.astral.sh/ruff/rules/error-suffix-on-exception-name/

modyn/model_storage/internal/grpc/model_storage_grpc_servicer.py
robinholzi: Not sure, upon manual inspection I didn't find other functions that don't comply with pep8 naming conventions.

black.toml
robinholzi: I changed it on purpose, but we can surely restore the local wildcards.

.pre-commit-config.yaml
MaxiBoether: Sounds good. Let's remove them in a follow up PR where we enable ruff.

scripts/python_compliance.sh
MaxiBoether: This is still open I think

MaxiBoether (approved):
Generally LGTM

1) shouldn't we put this into main to keep the diff of the other PR small?

2) i think in the python compliance script we still might want to use mamba in case micromamba is not available. everything else LGTM!

robinholzi:
> Generally LGTM
> 
> 1. shouldn't we put this into main to keep the diff of the other PR small?
> 2. i think in the python compliance script we still might want to use mamba in case micromamba is not available. everything else LGTM!

I merged in #447 so that there wouldn't be any merge conflicts later after merging #447 into main. Switching base branches on GitHub still makes sense to hide the the diff that comes from the other branch. I didn't intend to merge this PR into the other one.

PR #447, opened 2024-05-25
refactor: Use more explicit trigger configurations

robinholzi:
# Motivation

Previously it was rather hard to find out which fields and configuration items need to be set for different trigger policies, especially for drift policies as they have more parameters.

The pipeline configurations within the repository had a quite sparse coverage of the allowed flags. What is more it wasn't documented which of those fields are required, which are optional, and what the defaults are.

# Changes

Now we have dedicated configuration models for every trigger policy and do not need to run the pipeline to determine if a configuration is valid. The typing alone now assures a valid configuration and makes it easier for others to configure their own triggers.

#448 will be based on this

### DIFF
--- a/modyn/config/schema/pipeline.py
+++ b/modyn/config/schema/pipeline.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+from functools import cached_property
 from pathlib import Path
 from typing import Annotated, Any, Dict, List, Literal, Optional, Union
 
 from modyn.supervisor.internal.eval_strategies import OffsetEvalStrategy
 from modyn.utils import validate_timestr
+from modyn.utils.utils import SECONDS_PER_UNIT
 from pydantic import BaseModel as PydanticBaseModel
 from pydantic import Field, NonNegativeInt, field_validator, model_validator
 from typing_extensions import Self
@@ -417,11 +419,21 @@ class DataConfig(BaseModel):
 
 # ------------------------------------------------------ TRIGGER ----------------------------------------------------- #
 
+_REGEX_TIME_UNIT = r"(s|m|h|d|w|y)"
+
 
 class TimeTriggerConfig(BaseModel):
     id: Literal["TimeTrigger"] = Field("TimeTrigger")
-    every: int = Field(description="The interval length for the trigger specified by an integer", ge=1)
-    unit: Literal["s", "m", "h", "d", "w", "mth", "y"] = Field(description="The unit of the interval length.")
+    every: str = Field(
+        description="Interval length for the trigger as an integer followed by a time unit: s, m, h, d, w, y",
+        pattern=rf"^\d+{_REGEX_TIME_UNIT}$",
+    )
+
+    @cached_property
+    def every_seconds(self) -> int:
+        unit = str(self.every)[-1:]
+        num = int(str(self.every)[:-1])
+        return num * SECONDS_PER_UNIT[unit]
 
 
 class DataAmountTriggerConfig(BaseModel):
--- a/modyn/config/schema/pipeline.py
+++ b/modyn/config/schema/pipeline.py
@@ -431,14 +431,17 @@ class DataAmountTriggerConfig(BaseModel):
 
 class DataDriftTriggerConfig(BaseModel):
     id: Literal["DataDriftTrigger"] = Field("DataDriftTrigger")
-    data_points: int = Field(description="The number of data points for one drift detection.", ge=1)
+    detection_interval_data_points: int = Field(
+        1000, description="The number of samples in the interval after which drift detection is performed.", ge=1
+    )
     sample_size: int | None = Field(None, description="The number of samples used for the metric calculation.", ge=1)
-    detection_interval: int = Field(1000, description="The interval in which drift detection is performed.", ge=1)
     metric: str = Field("model", description="The metric used for drift detection.")
     metric_config: dict[str, Any] = Field(default_factory=dict, description="Configuration for the evidently metric.")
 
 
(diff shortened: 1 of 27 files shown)


modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -39,100 +39,14 @@ def __init__(self, trigger_config: dict):
...
 
         self.data_cache: list[tuple[int, int, int]] = []
         self.leftover_data_points = 0
 
-        if len(trigger_config) > 0:
-            self._parse_trigger_config(trigger_config)
-
-        super().__init__(trigger_config)
-
-    def _parse_trigger_config(self, trigger_config: dict) -> None:
```
robinholzi, line 55: Handled automatically by the pydantic models.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -39,100 +39,14 @@ def __init__(self, trigger_config: dict):
...
-            self.detection_interval = trigger_config["data_points_for_detection"]
-        assert self.detection_interval > 0, "data_points_for_detection needs to be at least 1"
-
-        if "sample_size" in trigger_config.keys():
-            self.sample_size = trigger_config["sample_size"]
-        assert self.sample_size is None or self.sample_size > 0, "sample_size needs to be at least 1"
-
-        self.metrics = get_evidently_metrics(self.evidently_column_mapping_name, trigger_config)
-
-    def _init_dataloader_info(self) -> None:
```
robinholzi, line 66: Moved internal helper functions to the file end.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -24,11 +27,8 @@
 class DataDriftTrigger(Trigger):
     """Triggers when a certain number of data points have been used."""
 
-    def __init__(self, trigger_config: dict):
-        self.pipeline_id: Optional[int] = None
```
robinholzi, line 28: Added a `dataclass` wrapper for this set of context variables.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -270,3 +184,84 @@ def inform_previous_trigger_and_data_points(self, previous_trigger_id: int, data
     def inform_previous_model(self, previous_model_id: int) -> None:
         self.previous_model_id = previous_model_id
         self.model_updated = True
+
+    # --------------------------------------------------- INTERNAL --------------------------------------------------- #
+
```
robinholzi, line 189: Only code rearrangements.

modyn/supervisor/internal/triggers/timetrigger.py
```diff
@@ -1,29 +1,25 @@
...
-        if "trigger_every" not in trigger_config.keys():
-            raise ValueError("Trigger config is missing `trigger_every` field")
-
-        timestr = trigger_config["trigger_every"]
-        if not validate_timestr(timestr):
-            raise ValueError(f"Invalid time string: {timestr}\nValid format is <number>[s|m|h|d|w].")
-
-        self.trigger_every_s: int = convert_timestr_to_seconds(trigger_config["trigger_every"])
-        self.next_trigger_at: Optional[int] = None
+    def __init__(self, config: TimeTriggerConfig):
```
robinholzi, line 15: Pydantic allows us a leaner trigger setup

modyn/supervisor/internal/triggers/utils.py
```diff
@@ -4,48 +4,13 @@
...
-from evidently.metrics import EmbeddingsDriftMetric
-from evidently.metrics.data_drift import embedding_drift_methods
 from modyn.supervisor.internal.triggers.embedding_encoder_utils import EmbeddingEncoder
 from modyn.supervisor.internal.triggers.trigger_datasets import DataLoaderInfo, FixedKeysDataset, OnlineTriggerDataset
 from torch.utils.data import DataLoader
 
 logger = logging.getLogger(__name__)
 
 
-def get_evidently_metrics(
```
robinholzi, line 16: moved into the `datadrifttrigger.py` as most functionality is covered automatically by pydantic.

modyn/config/schema/pipeline.py
```diff
@@ -418,16 +418,35 @@ class DataConfig(BaseModel):
 # ------------------------------------------------------ TRIGGER ----------------------------------------------------- #
 
 
-class TriggerConfig(BaseModel):
-    id: str = Field(description="Type of trigger to be used.")
+class TimeTriggerConfig(BaseModel):
+    id: Literal["TimeTrigger"] = Field("TimeTrigger")
+    every: int = Field(description="The interval length for the trigger specified by an integer", ge=1)
+    unit: Literal["s", "m", "h", "d", "w", "mth", "y"] = Field(description="The unit of the interval length.")
```
MaxiBoether, line 424: I'd actually like (in a Pythonic way) to also allow strings like "1m" but I guess that makes Pydantic validation harder. If that is a big problem, we can stick to the updated version.

MaxiBoether (approved):
And another one of those PRs which make everything just much more readable. Thanks so much Robin!

modyn/config/schema/pipeline.py
robinholzi: There are a couple of options to stick with the old config (e.g. "every: 1s") while keeping the splitting logic in the pydantic (and hence away from the business logic):
1) composite string field + two (maybe cached) property functions that splits up the composite string
2) adding a model validator in the before mode. (that can change the model inputs before pydantic sees them). As we would need a type change (str -> int) here, this would be rather nasty.

So let's go for 1, I agree that the old config format is more user-friendly :)

MaxiBoether: Thanks for integrating this!

PR #448, opened 2024-05-25
feat: Support running multiple experiments sequentially using a dynamic config

robinholzi:
# Motivation

We want to avoid having to duplicate pipeline configurations for several flavors of the same base pipeline (e.g. when using different trigger policies).

Additionally, we want a script that runs all configured pipelines sequentially without having to invoke multiple cli commands.

# Note:

I haven't properly tested this logic. Another branch, that is based on this one, is already quite big though. So I don't want to include this logic there to keep the diff small. If there should be any issues with this logic here, I'll resolve that in a follow-up PR. The risk of merging is low as nothing interferes with the logic on main.

### DIFF
--- a/modyn/config/schema/pipeline.py
+++ b/modyn/config/schema/pipeline.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+from functools import cached_property
 from pathlib import Path
 from typing import Annotated, Any, Dict, List, Literal, Optional, Union
 
 from modyn.supervisor.internal.eval_strategies import OffsetEvalStrategy
 from modyn.utils import validate_timestr
+from modyn.utils.utils import SECONDS_PER_UNIT
 from pydantic import BaseModel as PydanticBaseModel
 from pydantic import Field, NonNegativeInt, field_validator, model_validator
 from typing_extensions import Self
@@ -140,8 +142,9 @@ class MultiDownsamplingConfig(BaseModel):
     downsampling_thresholds: List[int] = Field(
         description=(
             "A list of thresholds to switch from a downsampler to another. The i-th threshold is used for the "
-            "transition from the i-th downsampler to the (i+1)-th. This array should have one less item on the list "
-            "of downsamplers."
+            "transition from the i-th downsampler to the (i+1)-th. This array should have one less item than the list "
+            "of downsamplers. For example, if we have 3 downsamplers [A, B, C], and two thresholds [5, 10], the "
+            "downsampler A is used for triggers 0-4, B for triggers 5-9, and C for triggers 10 and above."
         )
     )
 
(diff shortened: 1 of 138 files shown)


modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -492,6 +492,7 @@ def materialize(self, log_dir_path: Path, mode: Literal["initial", "increment",
             pipeline_logdir.rename(backup_dir)
 
         pipeline_logdir.mkdir(parents=True, exist_ok=True)
+        (pipeline_logdir / ".name").write_text(self.config.pipeline.pipeline.name)
```
MaxiBoether, line 495: Why this? We can infer this from the pipeline log, so why have another file?

modyn/config/schema/pipeline.py
```diff
@@ -429,6 +429,9 @@ class TimeTriggerConfig(BaseModel):
         description="Interval length for the trigger as an integer followed by a time unit: s, m, h, d, w, y",
         pattern=rf"^\d+{_REGEX_TIME_UNIT}$",
     )
+    sample_size: int | None = Field(None, description="The number of samples used for the metric calculation.", ge=1)
+    metric: str = Field("model", description="The metric used for drift detection.")
+    metric_config: dict[str, Any] = Field(default_factory=dict, description="Configuration for the evidently metric.")
```
MaxiBoether, line 434: I don't understand these changes - why do they belong to the time trigger?

modynclient/client/client.py
```diff
@@ -169,3 +170,24 @@ def poll_pipeline_status(self) -> None:
             logger.info(f"Pipeline <{self.pipeline_id}> not found.")
         else:
             logger.error(f"unknown pipeline status {json.dumps(res, sort_keys=True, indent=2)}")
+
+
+def run_multiple_pipelines(
```
MaxiBoether, line 175: Not sure this should go into the implementation file of the modyn client (since it uses client) Should this be part of the CLI instead maybe? What was the reasoning for putting it here?

modynclient/client/client.py
```diff
@@ -169,3 +170,24 @@ def poll_pipeline_status(self) -> None:
...
+    client_config: ModynClientConfig,
+    pipeline_configs: list[ModynPipelineConfig],
+    start_replay_at: int = 0,
+    stop_replay_at: int | None = None,
+    maximum_triggers: int | None = None,
+) -> None:
+    logger.info("Start running multiple experiments!")
+    logger.info(f"{client_config.model_dump()}")
```
MaxiBoether, line 183: Debug dump?

modynclient/client/client.py
```diff
@@ -169,3 +170,24 @@ def poll_pipeline_status(self) -> None:
...
+
+    for pipeline_config in pipeline_configs:
+        client = Client(client_config, pipeline_config, start_replay_at, stop_replay_at, maximum_triggers)
+
+    logger.info(f"Starting pipeline: {pipeline_config.pipeline.name}")
+    started = client.start_pipeline()
+    if started:
+        client.poll_pipeline_status()
```
MaxiBoether, line 191: Should this be indented one level further? Because now you instantiate a lot of clients (overriding the variable) and only start one. Maybe it's good to have one test run of this code since you mention you didnt check it yet 😅

modynclient/client/client.py
MaxiBoether: Maybe this even is should an experiments util function?

MaxiBoether (changes requested):
Thanks Robin, had some structural comments here

modyn/supervisor/internal/pipeline_executor/models.py
robinholzi: In the tool that I built for comparing/visualizing pipeline runs I don't want to load the whole file for only showing the user a list of available pipeline runs. Only after the user selection should it be loaded. To make the runs identifiable, I need some identifier that is easily accessible. We could alternatively also just rename the logfile that is currently named `pipeline.log`.

modyn/config/schema/pipeline.py
robinholzi: Oh yeah, that should not be here. It probably chose the wrong variant when resolving a merge conflict.

modynclient/client/client.py
robinholzi: It's not really an entrypoint. I like the experiment util idea!

modynclient/client/client.py
robinholzi: Just discovered that a couple of minutes ago on my local branch when running kick the script  :)

modynclient/client/client.py
robinholzi: It's also done in the entrypoint script, I'll remove it.

MaxiBoether: thx!

MaxiBoether (approved):
LGTM, thanks!

modyn/supervisor/internal/pipeline_executor/models.py
MaxiBoether: Yeah, then let's rename the file to `<pipeline_name>.log` :)

robinholzi: Will do in a followup PR, I don't wont to kick off the CI again here for that

PR #449, opened 2024-05-26
RHO-LOSS step 2: prepare the holdout set

XianzheMa:
This PR implements the logic to produce the holdout set to train an IL model.

### DIFF
--- a/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
+++ b/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
@@ -199,6 +199,19 @@ class AbstractSelectionStrategy(ABC):
         training_set_producer: Callable[[], Iterable[tuple[list[tuple[int, float]], dict[str, Any]]]],
         insertion_threads: int,
     ) -> tuple[int, int, dict[str, Any]]:
+        """
+        Store the training set, produced by the training_set_producer, as TriggerSampleStorage.
+        Relevant metadata for the trigger is also stored in the metadata database.
+
+        :param target_pipeline_id: the pipeline id the training set is associated with.
+        :param target_trigger_id: the trigger id the training set is associated with.
+        :param modyn_config: the modyn configuration.
+        :param training_set_producer: a callable that returns partitioned training samples. The type is the same as
+            the return type of the _on_trigger method.
+        :param insertion_threads: how many threads are used to store. If bigger than 1, multiple threads are used to
+            store the data.
+        :return: total number of keys in the trigger, number of partitions, and a log.
+        """
         # TODO(#276) Unify AbstractSelection Strategy and LocalDatasetWriter
         total_keys_in_trigger = 0
         log: dict[str, Any] = {"trigger_partitions": []}
--- a/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
+++ b/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
@@ -1,6 +1,6 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import Any, Iterable, Optional, Callable
+from typing import Any, Callable, Iterable, Optional
 
 import numpy as np
 from modyn.common.benchmark.stopwatch import Stopwatch
@@ -190,6 +190,7 @@ class AbstractSelectionStrategy(ABC):
             database.session.add(trigger_partition)
             database.session.commit()
 
+    # pylint: disable=too-many-locals
     @staticmethod
     def store_training_set(
         target_pipeline_id: int,
--- a/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
+++ b/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
@@ -1,6 +1,4 @@
 import logging
-import os
-import platform
 from abc import ABC, abstractmethod
 from typing import Any, Iterable, Optional
 
@@ -9,6 +7,7 @@ from modyn.common.benchmark.stopwatch import Stopwatch
 from modyn.common.trigger_sample import ArrayWrapper, TriggerSampleStorage
 from modyn.metadata_database.metadata_database_connection import MetadataDatabaseConnection
 from modyn.metadata_database.models import Trigger, TriggerPartition
+from modyn.selector.internal.storage_backend import AbstractStorageBackend
 from sqlalchemy import func
 
 logger = logging.getLogger(__name__)
@@ -70,14 +69,9 @@ class AbstractSelectionStrategy(ABC):
         self._modyn_config = modyn_config
         self._pipeline_id = pipeline_id
         self._maximum_keys_in_memory = maximum_keys_in_memory
-        self._insertion_threads = modyn_config["selector"]["insertion_threads"]
-
-        self._is_test = "PYTEST_CURRENT_TEST" in os.environ
-        self._is_mac = platform.system() == "Darwin"
-        self._disable_mt = self._insertion_threads <= 0
-
         if self._maximum_keys_in_memory < 1:
             raise ValueError(f"Invalid setting for maximum_keys_in_memory: {self._maximum_keys_in_memory}")
+        self._storage_backend = self._init_storage_backend()
 
         logger.info(f"Initializing selection strategy for pipeline {pipeline_id}.")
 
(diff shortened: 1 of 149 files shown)


modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -106,6 +99,12 @@ def __init__(
             )
 
         self._trigger_sample_directory = self._modyn_config["selector"]["trigger_sample_directory"]
+        self._storage_backend = self._init_storage_backend()
+
+    @abstractmethod
+    def _init_storage_backend(self) -> AbstractStorageBackend:
+        """Initializes the storage backend."""
+        raise NotImplementedError
```
XianzheMa, line 107: I found every selection strategy has a `StorageBackend`, it is just because we didn't implement both backends for all strategies, that the `self_storage_backend` is not able to become a field in the superclass.

This way it will become a field in the superclass. This has the additional advantage that the code to decide if we do multi-threading, i.e.
```
self._insertion_threads = modyn_config["selector"]["insertion_threads"]
self._is_test = "PYTEST_CURRENT_TEST" in os.environ
self._is_mac = platform.system() == "Darwin"
self._disable_mt = self._insertion_threads <= 0
```
previously redundant in both selection strategy and storage backend, can be deduplicated.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -217,59 +215,41 @@ def trigger(self) -> tuple[int, int, int, dict[str, Any]]:
...
 
-        for partition, (training_samples, partition_log) in enumerate(self._on_trigger()):
+        for partition, (training_samples, partition_log) in enumerate(training_set_producer()):
             overall_partition_log = {"partition_log": partition_log, "on_trigger_time": swt.stop("on_trigger")}
 
-            logger.info(
-                f"Strategy for pipeline {self._pipeline_id} returned batch of"
-                + f" {len(training_samples)} samples for new trigger {trigger_id}."
-            )
-
```
XianzheMa, line 227: removed because I think it's not very useful, and if we had to keep it, I have to add another parameter to the function `source_pipeline_id`, because now from which pipeline the data comes from can be different than which pipeline it goes to.

I just don't want to add another parameter simply to keep a logging message.

modyn/selector/internal/storage_backend/abstract_storage_backend.py
```diff
@@ -13,10 +13,15 @@ def __init__(self, pipeline_id: int, modyn_config: dict, maximum_keys_in_memory:
...
-        self._disable_mt = self._insertion_threads <= 0
+        raw_insertion_threads = modyn_config["selector"]["insertion_threads"]
+
+        is_test = "PYTEST_CURRENT_TEST" in os.environ
+        is_mac = platform.system() == "Darwin"
+
+        if raw_insertion_threads <= 0 or (is_test and is_mac):
+            self.insertion_threads = 1
+        else:
+            self.insertion_threads = raw_insertion_threads
```
XianzheMa, line 24: we should keep fields as few as possible

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -18,11 +25,53 @@ def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: Ab
...
+            selector_storage_backend, self._pipeline_id, next_trigger_id, tail_triggers=0
+        )
+
+        holdout_set_size = max(int(current_trigger_dataset_size * self.holdout_set_ratio / 100), 1)
+
+        stmt = self._get_holdout_sampling_query(self._pipeline_id, next_trigger_id, holdout_set_size).execution_options(
+            yield_per=self.maximum_keys_in_memory
+        )
+
+        def training_set_producer() -> Iterable[tuple[list[tuple[int, float]], dict[str, Any]]]:
```
MaxiBoether, line 53: We need to assert with this implementation that the `DatabaseStorageBackend` is used since this implementation is not agnostic to the backend.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -18,11 +25,53 @@ def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: Ab
...
+            with MetadataDatabaseConnection(self._modyn_config) as database:
+                for chunk in database.session.execute(stmt).partitions():
+                    samples = [res[0] for res in chunk]
+                    yield [(sample, 1.0) for sample in samples], {}
+
+        AbstractSelectionStrategy.store_training_set(
```
MaxiBoether, line 59: What I think is somehow missing here is marking the samples as used or so in the table. Otherwise, we will be able to have them in the IL set and the actual training set, and I think they need to be exclusive. Or do you plan to do this in a follow up PR?

MaxiBoether (changes requested):
Some smaller comments, but overall this looks good of course. Thank you!

modyn/selector/internal/selector_strategies/utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def get_trigger_dataset_size(
+    storage_backend: AbstractStorageBackend, pipeline_id: int, trigger_id: int, tail_triggers: Optional[int]
+) -> int:
+    # Count the number of samples that might be sampled during the next trigger. Typically used to compute the
+    # target size for presampling_strategies (target_size = trigger_dataset_size * ratio)
+    assert isinstance(
+        storage_backend, DatabaseStorageBackend
+    ), "CoresetStrategy currently only supports DatabaseBackend"
```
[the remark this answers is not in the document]
XianzheMa: Great catch! Thanks.

Yes I will keep this in mind. Right now let's first have a working RHO-LOSS version with Database backend!

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: There are several ways to determine the holdout set and the training set for the main model in the paper.
One way is to sample a predefined ratio of the training data as hold out set, and still use the entire training data as training set for the main model. This is how they train on Clothing-1M, and this is what we agreed on last time https://docs.google.com/document/d/19dAm9IfVz4tK8-y8tTuo4lIiy5P7jC-juJveyJYB1AI/edit#heading=h.lnhsfaqrxzvk.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: Thanks for pointing it out! `_prepare_holdout_set` is an internal method called by the `inform_next_trigger`, where the assertion is already made.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
MaxiBoether: Ahh, sorry, I did not realize that " The original training set is used for training the main model" includes these 10%. I checked the RHO-LOSS paper and they indeed seem to do this. Well, I think this might work for big datasets, for stuff like yearbook I don't know, but we can always extend it I guess.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
MaxiBoether: Ah, can you move the assertion to the constructor? I think that's where we typically assert this in the implementations

MaxiBoether (approved):
LGTM after the other small comment and CI is resolved :) Thanks!

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: The constructor has to access to this storage backend. The backend is directly passed to `inform_next_trigger`, see
https://github.com/eth-easl/modyn/blob/63be24037c82397a6160ba6d023732e37e683ee8/modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py#L19

MaxiBoether: Ah, I see. Thanks.

MaxiBoether (approved):

PR #452, opened 2024-05-28
feat: Add function fieldvalidators 

robinholzi:
# Motivation

When adjusting my experimental pipeline configuration files I experienced faulty indentations. However, my pipeline only failed in the supervisor during the pipeline execution. It's desirable to fail fast (with nice error message), so let's do the validation already in the pydantic model, then we wouldn't even hand invalid function to the supervisor.

### DIFF
--- a/modyn/config/schema/pipeline.py
+++ b/modyn/config/schema/pipeline.py
@@ -2,14 +2,15 @@ from __future__ import annotations
 
 from functools import cached_property
 from pathlib import Path
-from typing import Annotated, Any, Dict, List, Literal, Optional, Union
+from typing import Annotated, Any, Callable, Dict, List, Literal, Optional, Union
 
-from modyn.supervisor.internal.eval_strategies import OffsetEvalStrategy
-from modyn.utils import validate_timestr
-from modyn.utils.utils import SECONDS_PER_UNIT
 from pydantic import BaseModel as PydanticBaseModel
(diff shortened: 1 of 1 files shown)


modyn/config/schema/pipeline.py
```diff
@@ -417,6 +417,24 @@ class DataConfig(BaseModel):
...
+        try:
+            res = deserialize_function(value, "bytes_parser_function")
+            if not callable(res):
+                raise ValueError("Function 'bytes_parser_function' must be callable!")
+        except AttributeError as exc:
+            raise ValueError("Function 'bytes_parser_function' could not be parsed!") from exc
+        return value
+
+    @property
+    def bytes_parser_function_deserialized(self) -> Callable:
```
MaxiBoether, line 432: If we introduce this property, should we use it in the dataset implementations of the `OnlineDataset` and the `EvaluationDataset`(please don't freak out - they are messy with lots of code duplication haha, see #275)?

MaxiBoether (approved):
Thanks, just wondering if it's a big problem to use the property we introduce here in the codebase

modyn/config/schema/pipeline.py
robinholzi: yes, then the caller does not need to deserialize it themselves. I will leave this decision of whether, when, and whether to update you. I don't intend to touch that ;)

PR #454, opened 2024-05-29
Use explicit pydantic model in selector business logic

XianzheMa:
Sorry for this bulky PR; this is really not my style :(

RHO-LOSS needs more sophisticated parameters to the downsampling config. We currently use `dict` in business code to store the config, which is hard to reason, error-prune, and too cumbersome to further support RHO-LOSS.

Therefore, In this PR I changed the occurrence of all `dict` to the appropriate pydantic models, and also simplified some selection config settings.

After this PR, we will almost have replaced every occurrence of using `dict` as the config with the pydantic model!


benchmark/mnist/mnist.yaml
```diff
@@ -34,7 +34,7 @@ training:
     maximum_keys_in_memory: 1000
     storage_backend: "database"
     limit: -1
-    reset_after_trigger: True
+    tail_triggers: 0
 data:
```
XianzheMa, line 38: The functions of `reset_after_trigger` and `tail_triggers` overlap, that's why previously we have to do validations like when `reset_after_trigger == True` then `tail_trigger` cannot be bigger than 0.

The function of `reset_after_trigger` can completely be replaced by `tail_triggers`. If and only if `tail_triggers == 0`, `reset_after_trigger == True`.

Therefore I remove `reset_after_trigger` for clarity. It becomes a field in the config internally computed based on the value of `tail_triggers`.

experiments/yearbook/compare_trigger_policies/pipeline_config.py
```diff
@@ -41,7 +41,7 @@ def gen_pipeline_config(name: str, trigger: TriggerConfig, eval_strategy: EvalSt
             ],
             optimization_criterion=OptimizationCriterion(name="CrossEntropyLoss"),
             checkpointing=CheckpointingConfig(activated=False),
-            selection_strategy=NewDataSelectionStrategy(
+            selection_strategy=NewDataSelectionConfig(
                 maximum_keys_in_memory=1000, storage_backend="database", limit=-1, reset_after_trigger=True
```
XianzheMa, line 45: Otherwise this config class has a confusing name with the selection strategy class itself.

modyn/selector/internal/selector_manager.py
```diff
@@ -185,24 +186,16 @@ def uses_weights(self, pipeline_id: int) -> bool:
...
+        selection_strategy = TypeAdapter(SelectionStrategyModel).validate_json(raw_selection_strategy)
+        strategy_name = selection_strategy.name
         strategy_module = dynamic_module_import("modyn.selector.internal.selector_strategies")
         if not hasattr(strategy_module, strategy_name):
             raise NotImplementedError(f"Strategy {strategy_name} not available!")
 
         strategy_handler = getattr(strategy_module, strategy_name)
 
-        return strategy_handler(config, self._modyn_config, pipeline_id, maximum_keys_in_memory)
+        return strategy_handler(selection_strategy, self._modyn_config, pipeline_id)
```
XianzheMa, line 199: `maximum_keys_in_memory` is already a field in `config`, we don't need to create a separate argument for it.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -25,51 +26,24 @@ class AbstractSelectionStrategy(ABC):
...
-        modyn_config: dict,
-        pipeline_id: int,
-        maximum_keys_in_memory: int,
-        required_configs: Optional[list[str]] = None,
-    ):
+    def __init__(self, config: SelectionStrategy, modyn_config: dict, pipeline_id: int):
         self._config = config
 
-        if required_configs is None:
-            required_configs = []  # Using [] as default is considered unsafe by pylint
```
XianzheMa, line 39: required_configs will be already validated by pydantic

benchmark/mnist/mnist.yaml
MaxiBoether: Thank you, that seems reasonable! I suppose internally having this field still makes sense because it allows for some optimization when computing the TTS, in comparison to using a general tail triggers integer.

experiments/yearbook/compare_trigger_policies/pipeline_config.py
MaxiBoether: Hmm, I don't know what is the most Pydantic way here: I would have named this `NewDataSelectionStrategyConfig` since it's a config for the selection strategy. But at the same time, this is also the object that _defines_ what selection strategy we use + the according config. 

@robinholzi how would you name this in this case?

modyn/config/schema/modyn_base_model.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+
+class ModynBaseModel(BaseModel):
+    class Config:
+        extra = "forbid"
```
MaxiBoether, line 8: Sorry, I don't understand yet what this is doing and why we need it.

MaxiBoether: Ok now I see it has been there before but any reason you renamed this (and @robinholzi why does it exist in the first place?)

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -25,51 +26,24 @@ class AbstractSelectionStrategy(ABC):
...
-            self.tail_triggers = config["tail_triggers"]
-            if self.tail_triggers < 0 or not isinstance(config["tail_triggers"], int):
-                raise ValueError("Tail trigger must be an integer greater than 0")
-            if (self.reset_after_trigger and self.tail_triggers > 0) or (
-                (not self.reset_after_trigger) and self.tail_triggers == 0
-            ):
-                raise ValueError("Reset after trigger is equivalent to setting tail triggers to 0.")
+        self.uses_weights = config.uses_weights
+
+        self.tail_triggers = config.tail_triggers
```
MaxiBoether, line 38: What if `tail_triggers` is not given at all? Wouldn't this lead to a key error? do we need something like get(tail_triggers, None)?

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -10,13 +11,13 @@ def __init__(
         self,
         modyn_config: dict,
         pipeline_id: int,
-        downsampling_configs: list[dict],
+        downsampling_configs: List[DownsamplingConfig],
```
MaxiBoether, line 14: nit: list[] instead of List[] is the more modern way

modyn/config/schema/pipeline.py
```diff
@@ -101,43 +97,7 @@ class PresamplingConfig(BaseModel):
...
-        description=(
-            "In multi-epoch training and sample_then_batch, how frequently the data is selected. "
-            "`1` selects every epoch. To select once per trigger, set this parameter to 0."
-        ),
-        min=0,
-    )
-
-
-class MultiDownsamplingConfig(BaseModel):
+class MultiDownsamplingConfig(ModynBaseModel):
```
MaxiBoether, line 100: Should this go into downsampling config.py or stay here? It is a bit confusing this follows the same naming scheme as the stuff in the other file.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -78,26 +79,17 @@ def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: Ab
...
 def instantiate_scheduler(
-    config: dict, modyn_config: dict, pipeline_id: int, maximum_keys_in_memory: int
+    config: CoresetSelectionConfig, modyn_config: dict, pipeline_id: int
 ) -> DownsamplingScheduler:
-    if config.get("downsampling_config") is None:
-        # missing downsampler, use NoDownsamplingStrategy
-        list_of_downsamplers = [{"strategy": "No"}]
-        list_of_thresholds: list[int] = []
-    elif config["downsampling_config"].get("downsampling_list") is None:
+    if isinstance(config.downsampling_config, MultiDownsamplingConfig):
```
MaxiBoether, line 84: I think there is some confusing overload here. `MultiDownsamplingConfig` lives (currently) on a different abstraction level than the actual downsampling configs if I understand correctly. It lives on the pipeline config level. Now here we use it to check whether we have a mutlidownsamplingconfig or a single one. I think (even if it is just naming) something is a bit inconsistent here. I think downsampling_config is either the downsampler itself or meta information for downsampling, but right now it can be somewhat both? I am a little confused actually.

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
```diff
@@ -1,32 +1,29 @@
...
+        self.submodular_optimizer = downsampling_config.submodular_optimizer
+        self.selection_batch = downsampling_config.selection_batch
+        self.balance = downsampling_config.balance
         self.remote_downsampling_strategy_name = "RemoteSubmodularDownsamplingStrategy"
 
     def _build_downsampling_params(self) -> dict:
         config = super()._build_downsampling_params()
-
-        if self.downsampling_config.get("submodular_function") is None:
-            raise ValueError(
```
MaxiBoether, line 18: I think this might have gotten lost somehow, at least I don't find this list imported from deepcore anymore

MaxiBoether (changes requested):
Thank you Xianzhe! I had a couple of small comments, but I really appreciate the effort that went into fixing this and updating all of the tests etc... :-)

modyn/config/schema/modyn_base_model.py
XianzheMa: > why it exists
It is created by me in a previous PR, to be default disallow any field from json that is unknown to the schema. https://github.com/eth-easl/modyn/pull/436

> why I rename it
Previously it is called `BaseModel` which is the same name as the one by default provided by Pydantic. To avoid confusion I rename it.

modyn/config/schema/pipeline.py
XianzheMa: Sure let me move it to the downsampling config file. Thanks for catching it

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
XianzheMa: Even previously, `tail_triggers` is only optionally specified by the user, but will be a field in the Pydantic model absolutely (when value `None` if user does not specify it).

I also removed the default value provided by Pydantic if user does not specify it, because I think it's better to always let user explicitly specify it than relying on an unobvious implicit default.

An obvious implicit default is, for example, `NoPresamplingStrategy`

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
XianzheMa: I just moved `MultiDownsamplingConfig` also to the downsampling config file.

I understand this confusion. This exists before this PR (in main branch) https://github.com/eth-easl/modyn/blob/6c3d63f22e98f41d3bcbd0c0cc2fcd1053abdc4f/modyn/config/schema/pipeline.py#L228.

I can rename the pydantic model class `DownsamplingConfig` to `SingleDownsamplingConfig`. How do you like it?

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
XianzheMa: Sorry can you clarify it more? I didn't get it

experiments/yearbook/compare_trigger_policies/pipeline_config.py
robinholzi: I assume, there multiple strategies that allow selection? If so I would leep the `Strategy` infix. If there's no other flavor of selection the Config is just a semantic subunit of `TrainingConfig`,  I'd choose `NewDataSelectionConfig .

modyn/config/schema/downsampling_config.py
```diff
@@ -0,0 +1,145 @@
+from __future__ import annotations
+
+from typing import Annotated, List, Literal, Union
+
+from modyn.config.schema.modyn_base_model import ModynBaseModel
+from pydantic import Field, model_validator
+from typing_extensions import Self
+
+
+class BaseDownsamplingConfig(ModynBaseModel):
```
robinholzi, line 10: Ahh, that looks much cleaner and is way more user-friendly if someone isn't familiar with the configuration for sampling.

modyn/config/schema/downsampling_config.py
```diff
@@ -0,0 +1,145 @@
+from __future__ import annotations
```
robinholzi, line 1: Can we maybe put that into a `sampling` subdirectory?

modyn/config/schema/modyn_base_model.py
robinholzi: I think that's the cleaner solution now, thanks! Forbidding extra fields is useful for detecting misconfigurations on fields with defaults.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
robinholzi: Mind that the type annotation in pydantic only defines what the user of the model can expect. For a missing field with a model specification of `field: int | None` pydantic will raise missing value errors only if no default (e.g. None) is specified.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
robinholzi: I am not fully sure about the semantics behind `MultiDownsamplingConfig`, but why can't we use
` downsampling_config: list[DownsamplingConfig] = Field(default_factory=[] , ...)`

modyn/selector/internal/selector_strategies/downsampling_strategies/kcentergreedy_downsampling_strategy.py
```diff
@@ -1,16 +1,23 @@
...
 from modyn.selector.internal.selector_strategies.downsampling_strategies import AbstractDownsamplingStrategy
 from modyn.utils import DownsamplingMode
 
 
 class KcenterGreedyDownsamplingStrategy(AbstractDownsamplingStrategy):
-    def __init__(self, downsampling_config: dict, modyn_config: dict, pipeline_id: int, maximum_keys_in_memory: int):
+    def __init__(
+        self,
+        downsampling_config: KcenterGreedyDownsamplingConfig,
+        modyn_config: dict,
```
robinholzi, line 10: How much effort would it be to also use the pydantic model here?

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
XianzheMa: Because `MultiDownsamplingConfig` is more than a list of downsamplers :) See the definition in the pipeline.py

modyn/config/schema/downsampling_config.py
XianzheMa: sure

modyn/selector/internal/selector_strategies/downsampling_strategies/kcentergreedy_downsampling_strategy.py
XianzheMa: It would be another PR touching 30~ files :) I will leave it as as for now

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
XianzheMa: These submodular functions are in deepcore https://github.com/eth-easl/modyn/blob/6c3d63f22e98f41d3bcbd0c0cc2fcd1053abdc4f/modyn/trainer_server/internal/trainer/remote_downsamplers/deepcore_utils/submodular_function.py#L6

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
MaxiBoether: ahh, I see. sorry i might have been a bit confused. thanks

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
MaxiBoether: I do not find any place where you import and check this. I double checked  and see now that you hardcode the literal values in line 81 of downsampling_config.py. I think rather than hardcoding we should load the list of values from this function. Before, we checked explicitly whether they are supported there. Now if we were to add a new function we'd need to add it both and the schema and there, but it would be good to have the schema load it from the file you link as before.

experiments/yearbook/compare_trigger_policies/pipeline_config.py
MaxiBoether: Thanks Xianzhe!

MaxiBoether (approved):
Thank you! I think the `MultiDownsampling` thing needs some refactoring, but not important for this PR. My only small remaining comment is on the submodular functions, everything else LGTM!

modyn/selector/internal/selector_strategies/downsampling_strategies/submodular_downsampling_strategy.py
XianzheMa: I think it is the right practice to hardcode in the config file. The config file should be the single source of truth on what the system supports, as the config is the user interface and it is definitely difficult for a newbie to dig out where in the actual business code what function is supported.

I can actually delete the check on `RemoteDownsampler` side https://github.com/eth-easl/modyn/blob/6c3d63f22e98f41d3bcbd0c0cc2fcd1053abdc4f/modyn/trainer_server/internal/trainer/remote_downsamplers/remote_submodular_downsampling_strategy.py#L57 together with the `SUBMODULAR_FUNCTIONS` list there, because we know that once the config is validated, the value must be one of supported functions.

But I anyway have to have another PR that replaces the `dict` on the `RemoteDownsampler` side to Pydantic model, so I do not make the effort now. After this next PR (to replace dict on `RemoteDownsampler` side), the single occurence of supported function list is in the config.

MaxiBoether: Mhm, okay, since also in this file the validation is happening on a literal level and not an object/module level, I think you're right. If we would check what functions are available in an object sense I would disagree but let's not overcomplicate this and keep it only in the config file. Thank you!

PR #457, opened 2024-05-31
RHO-LOSS Step 1: fetch or create rho pipeline id

XianzheMa:
As the title, this PR implements the logic to fetch or create rho pipeline id.


modyn/config/schema/sampling/downsampling_config.py
```diff
@@ -99,6 +99,15 @@ class NoDownsamplingConfig(BaseDownsamplingConfig):
...
 
 
+class ILTrainingConfig(ModynBaseModel):
+    num_workers: int = Field("Number of workers to use for training.", min=1)
+    il_model_id: str = Field("The model class name to use as the IL model.")
+    il_model_config: dict = Field(
+        default_factory=dict, description="Configuration dictionary that will be passed to the model on initialization."
+    )
+    amp: bool = Field(False, description="Whether to use automatic mixed precision.")
+
```
XianzheMa, line 109: It's not a full list of needed training configs but that's what needed to finish this step.

modyn/metadata_database/models/auxiliary_pipelines.py
```diff
@@ -0,0 +1,20 @@
...
+from sqlalchemy import Column, ForeignKey, Integer
+
+
+class AuxiliaryPipeline(MetadataBase):
+    """
+    AuxiliaryPipeline model.
+    """
+
+    __tablename__ = "auxiliary_pipelines"
+    pipeline_id = Column("pipeline_id", Integer, ForeignKey(Pipeline.pipeline_id), primary_key=True)
```
MaxiBoether, line 12: Can't we add an optional `auxilary_pipeline_id` field to the Pipeline model? I think that would be the cleaner solution here. This here looks like the table implementing this relationship, but in the ORM sense, if a pipeline can have exactly one auxiliary pipeline, let's move it to the pipeline itself. The ORM framework takes care of creating those tables.

MaxiBoether (changes requested):
Thanks Xianzhe - I would suggest we do not introduce a new model but instead just add a new field on the `Pipeline` model to match more the object oriented ORM idea

modyn/metadata_database/models/auxiliary_pipelines.py
XianzheMa: We have to create a full dummy pipeline in the `pipeline` table, because otherwise I can't create entries in the trigger related tables because of foreign key constraint, e.g. usually they have a `pipeline_id` field which referencing a row in the `pipeline` table.

For example https://github.com/eth-easl/modyn/blob/028a646364a3360bdfd8355b986d8d1c9ed76cd1/modyn/metadata_database/models/triggers.py#L14

MaxiBoether: yeah, I am mainly suggesting to get rid of the `AuxiliaryPipeline` model and instead just add the field to the pipeline model with a foreign key to itself (but nullable=True)

XianzheMa: Added. Since I am adding a field anyway, I also created a field to store the data config https://github.com/eth-easl/modyn/blob/8f53ab8e2e1b95b02b15c851153ef051f4cbc72d/modyn/config/schema/pipeline.py#L347 in the `pipelines` table, because `RHOLossDownsamplingConfig` needs access to it to issue the training request (used in next PR RHO-LOSS step 3).

MaxiBoether (approved):
Looks perfect, thanks Xianzhe!

PR #459, opened 2024-06-02
RHO-Loss Step 3: Train IL Model

XianzheMa:
This PR is the last PR to implement the RHO-LOSS downsampler on selector side. After this PR, the whole process of training an IL model is finished. Only a PR to use the IL model to train the main model, i.e. `RemoteRHOLossDownsamplingStrategy` is missing.

The changes in this PR can be summarized as:
1. Extract the logic in `GRPCHandler` to talk with `trainer_server`  as a superclass `TrainerServerGRPCHandlerMixin`.
2. Extract the Pydantic Model `TrainingConfig` and place it in a separate file, so that `RHOLossDownsamplingConfig` can reference it (avoid circular import).
3. Use the capabilities of `TrainerServerGRPCHandlerMixin` in `RHOLossDownsamplingStrategy` to train IL model.


modyn/config/schema/pipeline.py
```diff
@@ -522,10 +432,6 @@ class EvalDataConfig(DataConfig):
         description="All metrics used to evaluate the model on the given dataset.",
         min_length=1,
     )
-    tokenizer: Optional[str] = Field(
-        None,
-        description="Function to tokenize the input. Must be a class in modyn.models.tokenizers.",
-    )
```
XianzheMa, line 528: redundant field: this field is already present in the base class `DataConfig`.

modyn/tests/supervisor/internal/test_grpc_handler.py
```diff
@@ -98,8 +99,9 @@ def get_minimal_pipeline_config() -> dict:
 @patch.object(StorageStub, "__init__", noop_constructor_mock)
 @patch.object(EvaluatorStub, "__init__", noop_constructor_mock)
 @patch("modyn.supervisor.internal.grpc_handler.grpc_connection_established", return_value=True)
+@patch("modyn.common.grpc.grpc_helpers.grpc_connection_established", return_value=True)
 @patch.object(grpc, "insecure_channel", return_value=None)
-def get_non_connecting_handler(insecure_channel, init) -> GRPCHandler:
+def get_non_connecting_handler(*args) -> GRPCHandler:
```
XianzheMa, line 104: many Mocks but none of them are used in the function body. `*args` is to save some efforts.

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -44,53 +44,27 @@
...
-        pipeline_status_queue: Optional[mp.Queue] = None,
-        training_status_queue: Optional[mp.Queue] = None,
-        eval_status_queue: Optional[mp.Queue] = None,
-    ) -> None:
+    def __init__(self, modyn_config: dict) -> None:
+        super().__init__(modyn_config=modyn_config)
         self.config = modyn_config
-        self.pipeline_status_queue = pipeline_status_queue
-        self.training_status_queue = training_status_queue
-        self.eval_status_queue = eval_status_queue
```
XianzheMa, line 81: it's interesting: these fields are actually not used anywhere

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
```diff
@@ -42,7 +42,6 @@ def __init__(
         self.requires_remote_computation = True
         self.maximum_keys_in_memory = maximum_keys_in_memory
         self.downsampling_config = downsampling_config
-        self.downsampling_params = self._build_downsampling_params()
```
XianzheMa, line 45: In the case of `RHOLossDownsampler`, the `downsampling_params` contains `il_model_id`, which is dynamic (contingent on trigger id). So we cannot attach a static field.

modyn/supervisor/internal/grpc_handler.py
MaxiBoether: hmm to be honest, idk why they are here. as you say, probably some leftovers...

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -101,7 +75,6 @@ def init_cluster_connection(self) -> None:
         self.init_evaluator()
 
     def init_storage(self) -> None:
-        assert self.config is not None
```
MaxiBoether, line 104: I think these assertions played a role in some unit tests, but if it still works and they don't throw i guess let's remoev them

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -651,11 +646,19 @@ def _train(self, s: ExecutionState, log: StageLog, trigger_id: int) -> int:
...
             s.previous_model_id,
             num_samples_to_pass,
         )
-        trainer_log = self.grpc.wait_for_training_completion(s.current_training_id, s.pipeline_id, trigger_id)
+
+        total_samples = self.grpc.get_number_of_samples(s.pipeline_id, trigger_id)
+        status_bar_scale = self.grpc.get_status_bar_scale(s.pipeline_id)
+        training_reporter = TrainingStatusReporter(
+            self.state.training_status_queue, trigger_id, s.current_training_id, total_samples, status_bar_scale
+        )
```
MaxiBoether, line 659: Why did you move the training status reporter initialization out of `wait_for_training_completion`? If I see correctly, it used to be in there (including status bar scale etc)

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
MaxiBoether: I wonder if we can cache the value using @cached_property by default. However, the problem is we'd need to invalidate this in the RHO-LOSS subclass, and I could not find out if that's possible. Probably calling the function on other classes is not expensive at all, so it should be fine

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
```diff
@@ -57,6 +56,10 @@ def _compute_status_bar_scale(self) -> int:
             return 100
         return self.downsampling_ratio
 
+    @property
+    def downsampling_params(self) -> dict:
+        return self._build_downsampling_params()
```
MaxiBoether, line 61: Actually should we get rid of `_build_downsampling_params()` and just let subclasses override `downsampling_params`? Not sure we need the extra level of indirection. Then subclasses can just use `@cached_property` and RHO LOSS uses @property

MaxiBoether (approved):
To be honest, I did not have many comments, mostly questions. I am mainly wondering about the status tracker in the trainer server grpc mixin and had a minor comment on the levels of indirection in `downsampling_params`. Thank you Xianzhe!

modyn/supervisor/internal/grpc_handler.py
XianzheMa: the type if `self.config` is already non-optional. And it is directly initialized in constructor with a non-optional modyn_config. So I think we are safe

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
XianzheMa: good idea!

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
XianzheMa: Yeah I wouldn't worry too much about an additional function call per trigger. The returned data is a small dict which takes no overhead to compute (if the returned data is computed & fetched from disk, then definitely we need to cache it)

PR #461, opened 2024-06-03
Move `SelectionStrategy` to root level of pipeline configuration

XianzheMa:
This PR lifts the position of Selection Config, i.e. making `SelectionStrategy` no longer a subfield in `Training` but a top-level field in `ModynPipelineConfig`.

The reasons are two-fold:
1. Conceptually, the two core questions Modyn tries to answer is when to train and what to train. So `SelectionStrategy` should be at the same level as `Trigger`.
2. To avoid circular import when implementing `RHOLossDownsamplingConfig`. Otherwise we have `PipelineConfig` -> `TrainingConfig` -> `SelectionStrategy` -> `RHODownsamplingConfig` -> `ILTrainingConfig` -> `TrainingConfig`.


MaxiBoether (approved):
Thank you Xianzhe, I was actually wanting to do this for quite some time now haha

PR #462, opened 2024-06-03
feat: Enable `torch.inference_mode` if downsampler supports it

MaxiBoether:
Not all downsamplers need information from autograd. With this change, we make the forward pass cheaper for those downsampling strategies.

### DIFF
--- a/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_abstract_matrix_downsampling_strategy.py
+++ b/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_abstract_matrix_downsampling_strategy.py
@@ -34,30 +34,30 @@ def test_init():
 @patch.multiple(AbstractMatrixDownsamplingStrategy, __abstractmethods__=set())
 def test_collect_embeddings():
     amds = AbstractMatrixDownsamplingStrategy(*get_sampler_config())
+    with torch.inference_mode(mode=(not amds.requires_grad)):
+        amds.matrix_content = MatrixContent.EMBEDDINGS
 
-    amds.matrix_content = MatrixContent.EMBEDDINGS
+        assert amds.requires_coreset_supporting_module
+        assert not amds.matrix_elements  # thank you pylint! amds.matrix_elements == []
 
-    assert amds.requires_coreset_supporting_module
-    assert not amds.matrix_elements  # thank you pylint! amds.matrix_elements == []
+        first_embedding = torch.randn((4, 5))
+        second_embedding = torch.randn((3, 5))
+        amds.inform_samples([1, 2, 3, 4], None, None, first_embedding)
+        amds.inform_samples([21, 31, 41], None, None, second_embedding)
 
-    first_embedding = torch.randn((4, 5))
-    second_embedding = torch.randn((3, 5))
-    amds.inform_samples([1, 2, 3, 4], None, None, first_embedding)
-    amds.inform_samples([21, 31, 41], None, None, second_embedding)
+        assert np.concatenate(amds.matrix_elements).shape == (7, 5)
+        assert all(torch.equal(el1, el2) for el1, el2 in zip(amds.matrix_elements, [first_embedding, second_embedding]))
+        assert amds.index_sampleid_map == [1, 2, 3, 4, 21, 31, 41]
 
-    assert np.concatenate(amds.matrix_elements).shape == (7, 5)
-    assert all(torch.equal(el1, el2) for el1, el2 in zip(amds.matrix_elements, [first_embedding, second_embedding]))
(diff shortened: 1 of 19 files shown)


XianzheMa:
Thanks. This is in my radar as well. I also wanted to improve on this

MaxiBoether:
> Thanks. This is in my radar as well. I also wanted to improve on this

So do you want to improve this PR or do the same? If we want to merge this I need an approval :D

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
```diff
@@ -90,3 +90,10 @@ def select_points(self) -> tuple[list[int], torch.Tensor]:
 
         selected_ids = [self.index_sampleid_map[sample] for sample in downsampled_idxs]
         return selected_ids, weights
+
+    @property
+    def requires_grad(self) -> bool:
+        if isinstance(self.per_sample_loss_fct, torch.nn.CrossEntropyLoss):
+            return False
```
XianzheMa, line 97: In the `self.get_scores` method, now we can remove the `with torch.inference_mode()`

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -120,3 +120,8 @@ def init_downsampler(self) -> None:
     def _select_indexes_from_matrix(self, matrix: np.ndarray, target_size: int) -> tuple[list[int], torch.Tensor]:
         # Here is where the actual selection happens
         raise NotImplementedError()
+
+    @property
+    def requires_grad(self) -> bool:
+        # Default to true if None
+        return self.matrix_content is None or self.matrix_content == MatrixContent.GRADIENTS
```
XianzheMa, line 127: `self.matrix_content` will be initialized by the subclasses. We can add an assertion `assert self.matrix_content is not None` here. In the actual return statement then we don't bother with `None`. See the `inform_samples` method in this class for example.

Or (actually a better way) we can add another argument `matrix_content` to the constructor of this baseclass, passed by the subclass to initialize the `self.matrix_content`. This way we can completely get rid of the nullity.

XianzheMa (approved):
Thanks. LGTM!

XianzheMa:
> > Thanks. This is in my radar as well. I also wanted to improve on this
> 
> So do you want to improve this PR or did you mean you wanted to do the same and now I did it already? If we want to merge this I need an approval :D We can also merge this PR first and whatever you want to further improve we can change later.

It is just that previously I didn't get time to take a look :D

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
MaxiBoether: the None case is True for tests, though. I needed to added it for unit test cases. Let me take another look

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_abstract_matrix_downsampling_strategy.py
```diff
@@ -34,30 +34,30 @@ def test_init():
 @patch.multiple(AbstractMatrixDownsamplingStrategy, __abstractmethods__=set())
 def test_collect_embeddings():
     amds = AbstractMatrixDownsamplingStrategy(*get_sampler_config())
+    with torch.inference_mode(mode=(not amds.requires_grad)):
+        amds.matrix_content = MatrixContent.EMBEDDINGS
```
XianzheMa, line 38: I think simply moving this line to above the `with` statement would work, i.e. so that `requires_grad()` does not need to consider None. Right?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
MaxiBoether: Done

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
MaxiBoether: moving to constructor

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_abstract_matrix_downsampling_strategy.py
MaxiBoether: probably, but i actually prefer the idea of moving this to the constructor to avoid those none fields!

PR #451, opened 2024-05-28
feat: Add more powerful evaluation methods

robinholzi:
# Motivation

Mostly motivated by #442

Add a more powerful abstraction that allows running multiple evaluation handlers within one pipeline. Also, support different techniques for scheduling and sampling data for evaluations.

# Changes

Mostly production logic changes are in `/supervisor/internal/eval` and `pipeline_executor.py`

Add `EvalTrigger`s:
- `AfterTrainingTrigger`: Run evaluation after every trigger/training
- `StaticEvalTrigger`: Run evaluation at pre-configured timestamps
- `PeriodicEvalTrigger`: Run evaluation in a pre-configured schedule

Add / adapt `EvalStrategy`
- `IntervalEvalStrategy`: from the point of evaluation add some offset in both directions defining the interval of evaluation
- `MatrixEvalStrategy`: same behavior as before (tumbling window)

The rest of the changes come from adjusted configs (moved to subdir), test rewrites, and documentation.

PR #464, opened 2024-06-03
feat: Pipeline run analytics visualisation tool

robinholzi:
dependent on #451 and #463

# Motivation

It is useful to have a tool that allows us to analyze and compare pipeline runs fast. This enables efficient experimentation and discovering of patterns in the log data. The findings from using this tool enable us to create plots for papers with almost no further data processing.

# Changes

Most of the changes are in `/analytics` and are completely greenfield. Some code duplication was needed for the plot components as the UI hooks have to be defined at the server start time (thus not enabling us to generate every component dynamically with all its UI)

# How to Review

It's more important to review the dataframe wrangling and aggregation compared to the UI boilerplate.


modyn/config/schema/pipeline/evaluation/metric.py
```diff
@@ -11,7 +11,7 @@
 class Metric(ModynBaseModel):
     name: str = Field(description="The name of the evaluation metric.")
     config: dict[str, Any] = Field({}, description="Configuration for the evaluation metric.")
-    evaluation_transformer_function: str = Field(
+    evaluation_transformer_function: str | None = Field(
         "",
```
MaxiBoether, line 15: There's a bit of inconsistency with this now I think, at some point we defaulted to None, sometimes to "". Anyways what's important is that we can also instantiate metrics without this (e.g. top n accuracy :D)

modyn/config/schema/pipeline/evaluation/metric.py
```diff
@@ -27,10 +27,15 @@ def validate_evaluation_transformer_function(cls, value: str) -> str | None:
             raise ValueError("Function 'evaluation_transformer_function' could not be parsed!") from exc
         return value
 
+    @field_validator("evaluation_transformer_function", mode="before")
+    @classmethod
+    def clean_evaluation_transformer_function(cls, value: str | None) -> str:
+        return value or ""
```
MaxiBoether, line 33: since the metric refactor PR also touches this, should be sure to sync this. What do you mean by clean btw? That it cannot be None?

dev-requirements.txt
```diff
@@ -13,3 +13,13 @@ grpcio-tools>=1.63
 grpcio>=1.63 # Linux
 mypy-protobuf
 pystack
+
+# Plotting
+plotly
+matplotlib
+seaborn
+pandas
```
MaxiBoether, line 21: pandas is already in the environment :)

analytics/tools/patch_logfile.ipynb
```diff
@@ -0,0 +1,191 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from analytics.app.data.transform import dfs_models_and_evals\n",
+    "\n",
+    "df_logs = logs_dataframe(logs)\n",
+    "# max_timestamp = df_logs[\"sample_time\"].max()\n",
+    "max_timestamp = 1703682949  # TODO: code needs minor change so that logfile contain max timestamp\n",
```
MaxiBoether, line 71: there is an open todo, although I used this notebook to test it and it worked for me :)

analytics/tools/patch_logfile.ipynb
```diff
@@ -0,0 +1,191 @@
+{
+ "cells": [
```
MaxiBoether, line 2: Can we add a brief comment on what this logfile patches (e.g, moves most recent model one forward)?

analytics/app/data/load.py
```diff
@@ -0,0 +1,32 @@
+import os
+from pathlib import Path
+
+from modyn.supervisor.internal.pipeline_executor.models import PipelineLogs
+
+EVAL_DIR = Path(os.environ["MODYN_EVAL_DIR"])
```
MaxiBoether, line 6: Is there any way to make this part of the UI or at least provide it as a command line arg? I switched around between different directories quite a lot (not sure if everything breaks down when the pipelines stem from different datasets, I kept that separate), and always re-exporting an environment variable was a bit annoying

MaxiBoether (approved):
To best honest, I am not sure how to review this properly. I can certainly give proof by example: Things I have plotted on my own looked identical to the plots the dashboard created. However, I am too unfamiliar with dash to even understand where to start looking. I am not sure what is executed after app.run(), how the data pipeline flows etc.

Given the current time constraints, my suggestion is to merge this and pray it is correct. I have high trust that it actually is because it looks identical to the plots I made before. But if we don't completely separate the data parsing and plotting logic, I would probably need to spend 1-2 days to just review this. I might actually go through it when I find the time but right now this is too much to comprehend.  Let's merge this to reduce the number of open PRs for now, and have the plotting ready to go

modyn/config/schema/pipeline/evaluation/metric.py
robinholzi: Good to know, thanks for the pointer!

modyn/config/schema/pipeline/evaluation/metric.py
robinholzi: Not sure what my intention was here, I'll try to rebase it after merging the other PR.

analytics/tools/patch_logfile.ipynb
robinholzi: Thanks for the pointer, the comment is outdated, originally I thought it made sense to a dedicated "max_timestamp" log metric to the logfile. When we update the sample_time in the `pipeline_executor` after executing a `HANDLE_NEW_DATA` loop, we can achieve the same by querying the max sample_time in the log dataframe.

analytics/app/data/load.py
robinholzi: CLI is possible, I'll add that. Adding it to the UI doesn't work as we need the data on the initial render for the ui callbacks to work... Plotly dash oddities I guess...

robinholzi:
@MaxiBoether Thanks for skimming through the PR! I guess skipping the thorough review is reasonable as long as we keep don't completely forget about the fact, that there could be issues! Reviewing would be hard anyways, as the dataframes have no annotations for columns etc. unfortunately.

When building plots for a paper/thesis we have to reinspect the logic anyways. Then it's much easier to validate on a specific case with specific values for fields like 'metric', 'dataset_id', 'eval_handler'!

Issue #408, opened 2024-04-30
Support top-N accuracy


MaxiBoether:
For CLOC, we want to try top 5 accuracy and see whether that helps to get more reasonable results. We need to add support for top N accuracy to the evaluator for this.

The question is a bit if we want to have top N and then top 1 as a special case of top N or not. We might run into a similar issue like #351, where we cannot use multiple scores unless we hack around it. We might want to refactor the evaluation metrics slightly to allow for multiple f1 scores/top N accuracies, if necessary.

Issue #351, opened 2024-01-08
Allow multiple types of F1 Scores


MaxiBoether:
Right now, the evaluator instantiates exactly one metric instance and filters them by name. This means there can be only a single F1 score. However, one might want to e.g., collect weighted and macro F1.

We need to support either a better instantiation function or refactor the F1 score to multiple classes representing the different types of F1 inheriting from a parent F1 class.

Example WeightedF1 hack:

```
from typing import Any

from modyn.evaluator.internal.metrics.f1_score import F1Score


class WeightedF1Score(F1Score):
    """
    Temporary Hack to allow weighted F1 and Macro FW
    - num_classes: the total number of classes.
    - (optional) average: the method used to average f1-score in the multiclass setting (default macro).
    - (optional) pos_label: the positive label used in binary classification (default 1), only its f1-score is returned.
    """

    def __init__(self, evaluation_transform_func: str, config: dict[str, Any]) -> None:
        config["average"] = "weighted"
        super().__init__(evaluation_transform_func, config)

    @staticmethod
    def get_name() -> str:
        return "WeightedF1-score"
```

PR #467, opened 2024-06-04
fix: Allow using multiple F1 scores in evaluation

MaxiBoether:
This changes the evaluator logic to allow different types of the same metric implementation (e.g., macro and micro F1), to collect as much data as possible.

Solves #351. Paves the way for #408.


modyn/selector/selector_entrypoint.py
```diff
@@ -15,13 +14,16 @@
...
+# Especially the import of the SelectorGRPCServer will transivitly cause errors 
 try:
     mp.set_start_method("spawn")
 except RuntimeError as error:
     if mp.get_start_method() != "spawn" and "PYTEST_CURRENT_TEST" not in os.environ:
-        logger.error("Start method is already set to {}", mp.get_start_method())
+        logger.error(f"Start method is already set to {mp.get_start_method()}")
         raise error
 
+from modyn.selector.internal.grpc.selector_server import SelectorGRPCServer  # fmt: skip  # noqa  # isort:skip
```
MaxiBoether, line 25: Don't ask me why, but when validating whether we can instantiate a metric in the pipeline schema, this fails unless the set the spawn start method before importing the gRPC server. I have no idea why this is only a problem with the selector and whatnot, but it is.

MaxiBoether: ok nevermind in the other classes this has been solved already apparently, not sure why it just came up now in the selector.

MaxiBoether:
@XianzheMa Do you think you'll be able to review this before next week? Otherwise I might ping @robinholzi since I want to merge this experimental stuff to run some small scale tests

XianzheMa:
> @XianzheMa Do you think you'll be able to review this before next week? Otherwise I might ping @robinholzi since I want to merge this experimental stuff to run some small scale tests

my productivity in the remaining of this week is not guaranteed :D please ping Robin thanks!

modyn/config/schema/pipeline.py
```diff
@@ -459,6 +459,20 @@ def evaluation_transformer_function_deserialized(self) -> Callable | None:
             return deserialize_function(self.evaluation_transformer_function, "evaluation_transformer_function")
         return None
 
+    @model_validator(mode="after")
+    def can_instantiate_metric(self) -> Self:
```
robinholzi, line 463: Rather than validating the `Metric.config` dictionary via a model validator, I would prefer a `MetricConfig` type that is a pydantic annotated union (like for selection strategy, trigger config, ...) of explicit metric config subtypes for every metric (e.g. `F1ScoreMetricConfig`, `AccuracyMetricConfig`, ...). I think that's fair as the metric classes are also part of our code base (and not of some library). This would make the configuration more explicit and easier to use. Also, we could strip the validation from the business logic. With the dictionaries, I faced the problem of not knowing what fields are available/necessary for certain metrics. Also, it doesn't catch the specifications of not existing options.
The model validator already helps to make things more explicit and easier to catch as validation is already done at config creation time, however, more explicit typing could verify config correctness just by typing compliance.
I understand though if you don't have the time to implement that, so I'm fine merging this as it's already an improvement over the status quo.

modyn/tests/evaluator/internal/grpc/test_evaluator_grpc_servicer.py
```diff
@@ -339,6 +340,35 @@ def test_setup_metrics(test_connect_to_model_storage, test_connect_to_storage):
...
+            config=JsonString(value=json.dumps({"num_classes": 2, "average": "micro"})),
+            evaluation_transformer=PythonString(value=""),
+        )
+        # not double macro, but macro and micro work
+        metrics = evaluator._setup_metrics([macro_f1_config, micro_f1_config, macro_f1_config])
+
+        assert len(metrics) == 2
+        assert isinstance(metrics[0], F1Score)
+        assert isinstance(metrics[1], F1Score)
+        assert metrics[0].average == F1ScoreTypes.MACRO
```
robinholzi, line 366: Mostly personal preference, but I think literal types are better suited here as they are more lightweight and support the same IDE convenience. I only use enum types if I need some way to define extra functionality through member functions.

robinholzi (approved):
Thanks for the improvements, if you want to go one step further and make the usage of metrics even more convenient, here are some ideas :)
If you decide to implement them, feel free to rerequest the review!

modyn/config/schema/pipeline.py
MaxiBoether: i guess you're right and i took the lazy approach. will try to make this better tomorrow. thanks for the suggestion and kicking my lazy ass here :D

modyn/tests/evaluator/internal/grpc/test_evaluator_grpc_servicer.py
MaxiBoether: i guess i did not really think too much here. when i refactor this to using the pydantic stuff, I'll probably choose literals :D

MaxiBoether:
thanks for the swift review!

modyn/config/schema/pipeline.py
MaxiBoether: As mentioned on Element, this has some implications that blow up this PR beyond its scope of allowing multiple F1 scores. Let's update this, but let's do it in a follow up PR. I had some initial implementation in https://github.com/eth-easl/modyn/pull/467/commits/acae267a9a2dcf487034a231d57ec62e58f6781e but there are some questions as mentioned on element on how to do this correctly

Issue #396, opened 2024-04-18
Trainer_server: inaccurate measurement of operations on GPU


XianzheMa:
The current way to measure operations on GPU, e.g. forward pass, backward pass, etc. is not accurate ([here](https://github.com/eth-easl/modyn/blob/ed861c92cc05f7ab99b2694df4f1281234539639/modyn/trainer_server/internal/trainer/pytorch_trainer.py#L410) as an example), because CUDA kernels are executed asynchronously. Instead of
```python
stopw.start("Forward", resume=True)
output = self._model.model(data)
stopw.stop("Forward")
```

we should synchronize cuda operations, i.e. do

```python
torch.cuda.synchronize(device=self._device)
stopw.start("Forward", resume=True)
output = self._model.model(data)
torch.cuda.synchronize(device=self._device)
stopw.stop("Forward")
```
.

Here is a chart to manifest the difference:
![image](https://github.com/eth-easl/modyn/assets/100585941/b1a67c59-dc1a-44ce-b039-5e1c866b6585)

 The left panel is without cuda synchronization and the right is with.

XianzheMa:
However, we probably don't want to directly inject those synchronization lines into the program, as synchronization slows down the training process. We should have a config "measure_gpu_operations", when set to True, we use the stop watch and add cuda synchronization.

robinholzi:
> However, we probably don't want to directly inject those synchronization lines into the program, as synchronization slows down the training process. We should have a config "measure_gpu_operations", when set to True, we use the stop watch and add cuda synchronization.

For my use case where I want to have a more fine granular view on the "Training" pipeline stage a synchronization option `measure_gpu_operations` wouldn't be a fair experiment setup. This is because I don't only want to compare the operations within training, but also set them into relation with other costs like trigger policy evaluation or information the selector. CUDA synchronization would probably affect training time negatively, skewing the comparison on pipeline level.

robinholzi:
Adding synchronization everywhere also looks quite brittle as we have quite some model calls spread across several loops and functions. It's quite easy to miss sth. there.

MaxiBoether:
No, for sure we don't want synchronization because that will decrease throughput. We cannot really measure those individual ops, unfortunately, just overall time spent training.

XianzheMa:
> > However, we probably don't want to directly inject those synchronization lines into the program, as synchronization slows down the training process. We should have a config "measure_gpu_operations", when set to True, we use the stop watch and add cuda synchronization.
[...]

PR #468, opened 2024-06-04
refactor: Reorganize and document `pytorch_trainer`

robinholzi:
# Motivation

I wanted to find out how different performance measurements were implemented in the pytorch trainer to be able to reason about them. I found it hard to follow the control flow though so I restructured the file a bit.

# Changes

What I did:
- Changed function order (moved core training loop function to the top and internal helper methods to the end.
- Added comment dividers
- Created a small control flow mermaid state diagram so new people can get a grasp of the training control flow more quickly.
- adjusted the names of local function names for consistency (made some private, and more important ones public)

# Review notice

The diff looks massive, however, the changes here are not really worth reviewing, as I didn't touch the function bodies except from the few cases where I left a gh-comment.


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -452,7 +287,7 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
                         loss = torch.dot(self._criterion_nored(output, target), weights / weights.sum())
                     else:
                         loss = self._criterion(output, target)
-                stopw.stop("Loss")
```
robinholzi, line 455: I added a level of indentation here, I suppose that's fair?

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -57,7 +57,7 @@
     seed_everything,
 )
 
-AvailableQueues = Enum("AvailableQueues", ["TRAINING", "DOWNSAMPLING"])
+AvailableQueues = Literal["TRAINING", "DOWNSAMPLING"]
```
robinholzi, line 60: I think it's appropriate to use literals here as the are more lightweight.

XianzheMa:
The time measurements there are not reliable. See https://github.com/eth-easl/modyn/issues/396

modyn/trainer_server/internal/trainer/pytorch_trainer.py
MaxiBoether: Yes, probably it will not make a difference, unless exiting the amp scope for some reason is an expensive operation. Good catch!

MaxiBoether (approved):
Thank you! I gave my reasoning for using enums, maybe it makes sense to you, maybe it doesn't :D I like enums for avoiding typos / silent failures but I would also be ok with changing this.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
robinholzi: Fair, probably that's related to flaky linting. In theory, linters / mypy can catch illegal literal values.

PR #469, opened 2024-06-04
feat: Support top-N Accuracy

MaxiBoether:
Instead of top-1 accuracy only, we also support top-n accuracy now. Note that this explicitly does NOT support evaluation transformer functions that reduce the model output to the chosen class.

Solves #408.


robinholzi (approved):
Looks good! I think I found a less invasive way to take advantage of pydantic models for metric configs. I'll try that once you merge this :)

PR #471, opened 2024-06-05
Fix selection strategy on RHO pipeline

XianzheMa:
This short PR fixes the dummy value for the slection strategy on rho pipeline.


integrationtests/run.sh
```diff
@@ -26,6 +26,8 @@ echo "Running evaluator integration tests"
 python $SCRIPT_DIR/evaluator/integrationtest_evaluator.py
 echo "Running supervisor integration tests"
 python $SCRIPT_DIR/supervisor/integrationtest_supervisor.py
+echo "Running metadata processor integration tests"
+python $SCRIPT_DIR/metadata_processor/integrationtest_metadata_processor.py
```
XianzheMa, line 30: missed integraion test somehow

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -18,7 +18,6 @@
 class RHOLossDownsamplingStrategy(AbstractDownsamplingStrategy):
 
     IL_MODEL_STORAGE_STRATEGY = ModelStorageStrategyConfig(name="PyTorchFullModel")
-    IL_MODEL_DUMMY_SELECTION_STRATEGY = "{}"
```
XianzheMa, line 21: Realized I shouldn't purely use a dummy place holder for the selection strategy on the rho pipeline. Because this value decides if we downsample data when training the IL model. So we should fill it with a selection strategy with no downsampling. (Also, it should be a schema-conforming config, because `selector_manager.py` will instantiate the strategy.)

integrationtests/run.sh
MaxiBoether: This ass on purpose. The metadata processor is not used ATM. We'd need something like this when we do something like TracIn embeddings, but right now the metadata processor is somewhat on ice, which is why we don't test it. If this runs through, we can have it but if it ever causes problems and we're not using the processor I would comment it out again :D

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -28,6 +27,11 @@ def __init__(
         maximum_keys_in_memory: int,
     ):
         super().__init__(downsampling_config, modyn_config, pipeline_id, maximum_keys_in_memory)
+        # The choice of the selection strategy does not matter as long as it does not downsample data
+        # at the trainer server. (We don't want to further select on holdout set)
```
MaxiBoether, line 31: It does not matter? I think if you e.g. went for random 50% presampling it would matter, wouldn't it? I agree this is the correct strategy to use here but I don't understand why it does not matter

MaxiBoether (approved):

integrationtests/run.sh
XianzheMa: thanks for the context!

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: The presampling does not matter. Even if we have a random 50% presampling we never put it into use. 

Recall normally how presampling is triggered is by calling `inform_data_and_trigger`, which internally computes and stores the TSS using a SQL query produced by presampler.

In the RHO-LOSS case, the rho pipeline's training data (i.e. holdout set) is hard-code computed by sampling a ratio of trigger data from the main pipeline. The presampling config of the rho pipeline never plays a role.

The downsampling config of the rho pipeline can play a role however (if we set a downsampler), as the training process on the trainer server side for a rho pipeline is identical to a normal pipeline.

integrationtests/run.sh
XianzheMa: I found it's not immediately runnable. Then I will just comment it!

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
MaxiBoether: Ah, that makes sense. I somehow thought you'd inform the selector about the samples you want to use for IL model training again.

MaxiBoether (approved):

Issue #163, opened 2023-02-24
Configure how often to perform lr_scheduler.step()


fotstrt:
Currently, if a `lr_scheduler` is provided, we call `lr_scheduler.step()` at each training step. However, the user might need to update the `lr_scheduler` at a different granularity, for example [per epoch](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate). This should be configurable, for example via the pipeline config file.

MaxiBoether:
Closed :)

PR #472, opened 2024-06-05
feat: Allow to pass number of expected batches or epochs to learning rate scheduler

MaxiBoether:
In this PR,

1) we allow the user to configure whether to step the LR scheduler every batch or epoch
2) introduce two constants `MODYN_NUM_EPOCHS` and `MODYN_NUM_BATCHES` which -- hopefully, guarded by assertions -- describe how many epochs and batches we will train on that can be used in the learning rate scheduler configuration instead of giving a number.

Solves #163.


modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -192,16 +192,26 @@ def get_training_info(
...
             "optimizers": ["default"] if num_optimizers == 1 else ["opt1", "opt2"],
             "config": {},
         }
+    elif lr_scheduler == "torch_cosine":
+        lr_scheduler_config = {
+            "name": "CosineAnnealingLR",
+            "source": "PyTorch",
+            "step_every": "batch",
+            "optimizers": ["default"] if num_optimizers == 1 else ["opt1"],
+            "config": {"T_max": "MODYN_NUM_BATCHES"},
```
robinholzi, line 213: I find this way of encoding that configuration option a bit weird. I don't have a better idea though as I don't now what potential other options the config dictionary could hold here. Let's keep that for now :)

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -537,6 +596,14 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
         self._log["num_batches"] = batch_number + 1
         self._log["total_train"] = total_stopw.measurements.get("TotalTrain", 0)
 
+        assert (
```
robinholzi: Thanks, I think it's good to know for certain how many epochs we eventually trained

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -154,6 +154,29 @@ def __init__(
...
+        num_samples_in_trigger = self.get_num_samples_in_trigger()
+        batches_per_epoch = num_samples_in_trigger // self._batch_size  # We reuse this later
+        self._expected_num_batches = batches_per_epoch
+
+        num_samples_per_epoch = (
+            self._expected_num_batches * self._batch_size
+        )  # scale up again to multiples of batch size
+
+        if downsampling_enabled:
+            num_samples_per_epoch = max(int(self._downsampler.downsampling_ratio * num_samples_per_epoch / 100), 1)
```
robinholzi, line 169: why not use the `//` operator?

robinholzi (approved):
Nice, thanks. I find the configuration format a bit unintuitive, but okay for now :)

MaxiBoether:
haha you've reviewed before it's fully done but it should be done soon :D thank you anyways, will address your comments!

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
MaxiBoether: Yes, I also am not sure if this is the best option, but we somehow need to offer a "runtime" replacement here for a very particular use case, so for now it should be ok...

modyn/trainer_server/internal/trainer/pytorch_trainer.py
MaxiBoether: cool, yeah, also to investigate/log the training samples vs time and investigate the weirdness we discussed today...

modyn/trainer_server/internal/trainer/pytorch_trainer.py
MaxiBoether: addressed

MaxiBoether:
@robinholzi I just re-requested a review since now the implementation is complete and I tested it on some variations of the MNIST pipeline.  I hope it will not break in other configs with some wrong calculations :) Anyways, we only fail in case the number of batches is truly relevant (when using a LR scheduler), in other case we just warn.

image

Issue #385, opened 2024-04-08
C++ Storage: CsvFileWrapper (and FileWatcher if necessary)


jenny011:
# Bug
This bug is specific to csv data files.
There are several corrupted lines in wildtime huffpost and arxiv training datasets. `rapidcsv` cannot parse them correctly, thus `GetCell()` used in `get_label()` throws a vector range error.
The corrupted lines are likely the same ones causing python csv reader to merge multiple samples together.
FileWatcher does not deal with this error gracefully. Whenever the error occurs, samples collected but not yet inserted to db are gone, regardless of whether they are corrupted or not. That results in 0 entry for corresponding files in `samples` table.

# Proposed Fix
It is not yet clear how to fix the samples, so I suggest to remove the corrupted samples either in CsvFileWrapper or before inserting into modyn.

PR #473, opened 2024-06-06
fix: Fix csv quoting that broke some datasets

robinholzi:
# Motivation

Some datasets like huffpost use `rapidcsv`'s default quoting chars in their feature columns. This led to some years of the dataset not being discovered by the storage server. We want to have control over the `rapidcsv` quoting behavior within `modyn_config.yaml` to e.g. disable quoting for some datasets.

image

I'm not fully sure what decision lead to [skipping some samples](https://github.com/eth-easl/modyn/blob/4a9cd98bd9d199e907dc707c2e89149c5dce3f13/benchmark/wildtime_benchmarks/data_generation_arxiv.py#L20)  rather than adding those quoting options. 
As far as I can see, this addresses https://github.com/eth-easl/modyn/issues/385

# Changes

- Propagate relevant `rapidcsv` parsing options to `modyn_config`.

### DIFF
--- a/modyn/config/schema/config.py
+++ b/modyn/config/schema/config.py
@@ -64,6 +64,9 @@ class DatasetCsvFileWrapperConfig(_DatasetBaseFileWrapperConfig):
     """
 
     separator: str = Field(",", description="The separator used in CSV files.")
+    quote: str = Field("\\0", description="The quote character used in CSV files.")
+    quoted_linebreaks: bool = Field(True, description="Whether linebreaks are quoted in CSV files.")
+
     label_index: int = Field(
         description=(
             "Column index of the label. For columns 'width, 'height, 'age', 'label' you should set label_index to 3."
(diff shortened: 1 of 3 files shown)


modyn/config/schema/config.py
```diff
@@ -64,6 +64,9 @@ class DatasetCsvFileWrapperConfig(_DatasetBaseFileWrapperConfig):
     """
 
     separator: str = Field(",", description="The separator used in CSV files.")
+    quote: str = Field("\\0", description="The quote character used in CSV files.")
+    quoted_linebreaks: bool = Field(True, description="Whether linebreaks are quoted in CSV files.")
```
MaxiBoether, line 68: Do these default values already fix the behavior for arXiv and huffpost or should we update the modyn example config for them?

MaxiBoether (approved):
LGTM with some questions, but I would be happy if @XianzheMa could double check this since he dealt with this issue before and I have not checked out the corrupted lines.

modyn/config/schema/config.py
robinholzi: I only tested it for `huffpost`, there the null-char default (disables quoting more or less). There this resolved all my issues. I can check for arxiv before merging though :)

MaxiBoether: Thanks, that would be great!

robinholzi: Training works for `arxiv` :)

PR #474, opened 2024-06-06
Add description and scripts for downloading and generating the CGLM dataset

MaxiBoether:
This adds a data generation script and some documentation on the CGLM dataset. This can be further improved but for now gets the job done. 

I also wanted to persist the scripts for scraping the metadata somewhere, but they are not cleaned at all. I am okay with this, since I do not want to put a lot of energy into that right now and we provide the final scraped data anyways, so it's mostly for documentation purposes (unless you want to scrape wikimedia for 2 weeks ;D).

### DIFF
--- a/benchmark/cglm/README.md
+++ b/benchmark/cglm/README.md
@@ -1,0 +1,89 @@
+# CLOC Data
+
+In this directory, you can find the files necessary to run experiments with the CGLM dataset. 
+This dataset contains pictures of landmarks (e.g., the Montblanc) over time.
+The dataset was introduced [in this paper](https://drimpossible.github.io/documents/ACM.pdf).
+However, their host is down and we have to regenerate the dataset.
+It makes use of the [Google Landmarks V2 dataset](https://github.com/cvdfoundation/google-landmark).
+
+## Generating the dataset for Modyn
+
+As a first step, you will need to download the Google Landmarks v2 Images.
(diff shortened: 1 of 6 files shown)


MaxiBoether:
@robinholzi if you want to you can review this but since this is more documentation and scripts feel also free to just move all responsibility to me and approve this :D

benchmark/cglm/README.md
```diff
@@ -0,0 +1,89 @@
...
+mkdir cglm && cd cglm
+bash ../download-dataset.sh train 499
+```
+
+Next, we need the image metadata.
+We have pregenerated and host the metadata on Github for convenience.
+If you want to check out how we generate this metadata, check out the next file.
+
+```bash
+wget https://github.com/eth-easl/cglm-metadata/raw/main/cglm_labels_timestamps_clean.csv
```
robinholzi, line 25: It might make sense to store the fully scraped dataset (before dataframe transformations) in public nfs share of the sgs cluster.

robinholzi (approved):
Thanks for making this new dataset available :)
I just skimmed through the code, I guess it serves its purpose, code quality doesn't have to be perfect there.

benchmark/cglm/README.md
MaxiBoether: Will do!

PR #476, opened 2024-06-07
feat: Evaluator cleanup endpoint

robinholzi:
# Motivation

We want to release resources that were stored for an evaluation after their results have been collected in the supervisor.


modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -317,3 +319,32 @@ def get_evaluation_result(
...
+                process_handler.terminate()
+                process_handler.join(timeout=30)
+                if process_handler.is_alive():
+                    process_handler.kill()
+
+            self._evaluation_process_dict.pop(evaluation_id)
+
+        for e_id in evaluation_ids:
+            self._evaluation_dict.pop(e_id)
+
```
MaxiBoether, line 349: You might want to run a gc.collect() here.

MaxiBoether (approved):
Just a small comment on using `gc.collect` everything else LGTM!

PR #478, opened 2024-06-07
refactor: Split up pipeline config and fix imports

robinholzi:
# Motivation

To reduce merge conflicts going forward let's split up the pipeline config into separate modules and subpackages.

# Changes

- Splitting up pipeline config
- Fixing imports: Use relative imports to avoid importing the `modyn.config` package within its descendants.


MaxiBoether (approved):
Thanks Robin! Makes handling this a lot easier.

PR #479, opened 2024-06-08
feat: Add eval handler abstraction

robinholzi:
# Motivation

We want to have evaluation handlers and be able to specify multiple handlers per pipeline. Also, we refine the config models for more evaluation strategies and evaluation triggers that will be implemented in a followup PR.


modyn/config/schema/pipeline/evaluation/config.py
```diff
@@ -46,15 +54,22 @@ class EvaluationConfig(ModynBaseModel):
         ),
         min_length=1,
     )
-    datasets: List[EvalDataConfig] = Field(
-        description="An array of all datasets on which the model is evaluated.",
+    datasets: dict[str, EvalDataConfig] = Field(
+        description="An array of all datasets on which the model is evaluated keyed by an arbitrary reference tag.",
```
MaxiBoether, line 58: I guess I don't fully get that yet:

````
    arxiv_test:
      dataset_id: arxiv_test
```

Is there any ways to avoid this redundant specification of arxiv_test? Can't we use the reference tag as the name?

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -0,0 +1,30 @@
...
+
+class EvalHandlerConfig(ModynBaseModel):
+    name: str = Field(
+        "Modyn EvalHandler", description="The name of the evaluation handler used to identify its outputs in the log"
+    )
+    strategy: EvalStrategyConfig = Field(description="Defining the strategy and data range to be evaluated on.")
+    trigger: EvalTriggerConfig = Field(None, description="Configures when the evaluation should be performed.")
+    datasets: List[str] = Field(
+        description="All datasets on which the model is evaluated.",
+        min_length=1,
```
MaxiBoether, line 20: I don't get why we have another datasets list here. We already have a datasets list in the evaluation config. Why do we need to specify the IDs again?

modyn/config/schema/pipeline/evaluation/trigger/base.py
```diff
@@ -0,0 +1,10 @@
+from modyn.config.schema.base_model import ModynBaseModel
+from pydantic import Field
+
+
+class MatrixEvalTriggerConfig(ModynBaseModel):
+    """Base class for EvalTriggerConfigs that allow to evaluate every model to be evaluated at the trigger times"""
+
+    matrix: bool = Field(
+        False, description="Weather to evaluate all models at this these points or always only the most recent one"
```
MaxiBoether, line 9: I don't understand this

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -710,19 +710,21 @@ def _evaluate_and_store_results(
...
         """Evaluate the trained model and store the results."""
+        # pylint: disable=too-many-locals
         assert self.grpc.evaluator is not None, "Evaluator not initialized."
         assert self.state.pipeline_config.evaluation is not None, "Evaluation config not set."
         s.pipeline_status_queue.put(
             pipeline_stage_msg(PipelineStage.EVALUATE, MsgType.ID, id_submsg(IdType.TRIGGER, trigger_id))
         )
 
-        eval_strategy_config = self.state.pipeline_config.evaluation.eval_strategy
+        eval_strategy_config = self.state.pipeline_config.evaluation.handlers[0].strategy
```
MaxiBoether, line 720: I don't get yet why we just take the first handler. Is this PR not using the actual handlers yet?

MaxiBoether (changes requested):
The code itself looks good but conceptually I am unfortunately confused about the relationship of strategies, handlers, and triggers, and timestamps.

From my understanding, one handler comprises a trigger (=> when should the evaluation be executed) and a strategy (=> on what data should we evaluate). Is that correct? I have two main questions:

1) I don't understand evaltriggers really, to be honest. I feel like the level of abstraction is weird. `AfterTrainingTrigger` is a Modyn concept. After we finish a training, we evaluate the trained model on all intervals defined by a eval strategy. But now the `StaticEvalTrigger` and `PeriodicEvalTrigger` live on the data timeline, not on the Modyn timeline. I don't understand how this maps to the same level of abstraction. Let's say we define a static schedule at timestamps 42 and 1337. How does this work? Does this mean as soon as you see a sample with a timestamp >= 42, you'll take the current trained model, and evaluate it? If so, then you could also run all those evaluations directly after training, couldn't you? To me, I would understand if we e.g. had an `AfterTrainingEvalTrigger` or  `AfterPipelineEvalTrigger` that in the experimental case fires when the entire pipeline is over. I don't understand how these different time concepts map to the triggering concept. 

2) Doesn't the strategy (mostly) imply the trigger? If we have a strategy that requires us to know when triggers happen, we need to run the evaluation after the pipeline is done.  For the matrix case and the offset case, we can just evaluate after each training. I don't really understand the levels of abstraction here, unfortunately, it seems.

Maybe some documentation/examples on the different concepts would help me either clarify my points or understand the design decisions you made here. Thanks so much already

modyn/config/schema/pipeline/evaluation/trigger/periodic.py
```diff
@@ -0,0 +1,31 @@
...
+from modyn.utils.utils import SECONDS_PER_UNIT
+from pydantic import Field
+
+from .base import MatrixEvalTriggerConfig
+
+
+class PeriodicEvalTriggerConfig(MatrixEvalTriggerConfig):
+    """Comparable to TimeTriggerConfig and DataAmountTriggerConfig, but for scheduling evaluation runs.
+
+    Note: This strategy will run evaluations after the core pipeline.
```
MaxiBoether, line 14: Here for example. This is a trigger. But is it also a strategy? I dont understand the difference

MaxiBoether (changes requested):

modyn/config/schema/pipeline/evaluation/trigger/base.py
MaxiBoether: how can matrix on the matrix eval startegy be false?

modyn/config/schema/pipeline/evaluation/trigger/base.py
```diff
@@ -0,0 +1,10 @@
+from modyn.config.schema.base_model import ModynBaseModel
+from pydantic import Field
+
+
+class MatrixEvalTriggerConfig(ModynBaseModel):
+    """Base class for EvalTriggerConfigs that allow to evaluate every model to be evaluated at the trigger times"""
```
MaxiBoether, line 6: is this base or matrix? i am confused

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    """ Note: the datasets have to be defined in the root EvaluationConfig model."""
+
+    @model_validator(mode="after")
+    def check_trigger(self) -> "EvalHandlerConfig":
+        if isinstance(self.strategy, UntilNextTriggerEvalStrategyConfig) and not (
+            isinstance(self.trigger, AfterTrainingEvalTriggerConfig)
+        ):
+            raise ValueError("UntilNextTriggerEvalStrategyConfig can only be used with AfterTrainingEvalTriggerConfig.")
```
MaxiBoether, line 29: but after training means directly after training doesn't it? from my understanding, we can run an evaluation either after training or after all trainings (after the pipeline), and if we need to know the next trigger, we need to run it after the pipeline. how can we run it after the training? i think this might be another indication that we should adapt the levels of abstraction

MaxiBoether:
I think what I initially thought  this contains is the implementation of the current supported stuff in the main branch, but with the concept of evalhandlers and triggers and strategies. Anyways, we have to figure out the relationships between those concepts first, but it might make sense to have this PR just the existing eval strategies and triggers (if we in some form keep those concepts) and then the follow up PR implements the actual newe strategies and triggers that you need.

robinholzi:
Thanks @MaxiBoether for raising your concerns about the level of abstraction, there is, indeed, some entanglement between the different dimensions of our evaluation setup.

As you correctly identified, I initially thought that the concepts of “when” to evaluate and “what data” to evaluate are orthogonal. E.g. when triggering at the time of training or at a fixed schedule you could decide to take data some time into future and past of the the trigger time. However, in some cases like the current `MatrixEvaluation` and the `BetweenTwoTriggers` evaluation they are dependent. Also, the “when” question is actually overloaded. There is the “when” in terms of “Modyn pipeline time” and “dataset time”. 

I think the naming isn’t totally clear and those concepts should be divided up in our setup. However, it’s not straightforward as our experiment pipeline setup also mixes those concepts when accessing future data within one pipeline run for evaluating models on all data slices. During implementation I was probably focusing on the “production case” too much where the evaluation schedule should be more independent from triggering.

Regarding the current `MatrixEvaluation` there already is a naming ambiguity. Essentially, it should mean that every model is evaluated on every evaluation interval. This is not a strategy (that selects the point in time and data) itself, but rather an option of every strategy/handler. E.g. we could also define the evaluation intervals as the intervals between two triggers and train every model on every of those. The same applies to intervals generated by a fixed periodic schedule or static points in the datasets.

I gave it another thought and propose the following abstraction (I left it close to the config format)

```yaml
handler: 
  # on modyn pipeline level
  # configures what datasets a certain evaluation series should run on, what intervals and models to combine, ...;
  # this basically reduces the choices in the space `Datasets x Models x Intervals x TimeOfExecutionInPipeline`
  
  execution_time: Literal[“after_training”, “after_pipeline”, “manual”, “async_during_pipeline”]  
  # when the evaluation handlers should be executed; 
  # "manual": will skip the evaluations in the pipeline run; useful if someone wants to start them with a different entry point later on in isolation 
    # (useful for debugging and when evaluation is not part of the performance critical part of the pipeline, can be executed while the system is busy with other things e.g.)
  # "async_during_pipeline": future feature, not needed atm., useful when the pipeline should perform evaluations asynchronously during the pipeline (e.g. for continously generating data for training trigger decisions)

  models (shared across all strategies): Literal[“matrix”, “most_recent”]
  # what models to evaluate the strategy on, matrix will evaluate every model on every interval given by the strategy
  
  strategy: Union[SlicingEvalStrategy, BetweenTwoTriggersEvalStrategy, StaticEvalStrategy, PeriodicStrategy]
  # on dataset level; something that defines start & end timestamp of evaluation requests; 
  # determines both the point in time (dataset) and range of data to start evaluating on;
  # SlicingEvalStrategy: effectively what the old MatrixEvalStrategy did (in terms of interval generation, not in terms of defining the cross product with all models)
  # PeriodicStrategy & StaticEvalStrategy: these strategies have orthogonal "when" (schedule) and "what data" concepts (offsets in potentially both directions of the scheduled times)

  dataset: list[str]
  # list of dataset references defined at top level of evaluation config, to enable running one handler on multiple datasets (e.g. train, test, combined)
  # this makes sense as e.g. the BetweenTwoTriggersEvalStrategy can actually run on the training data.
  # we could default this to use run the handler on every of the datasets in the root setting
```

If you agree with this abstraction I can adjust the implementation of the config models and business logic (just requires some rearranging of the existing code, no new implementation)

modyn/config/schema/pipeline/evaluation/trigger/base.py
robinholzi: Because matrix strategy time slices the dataset and builds tumbling windows - more than it's name promises actually. The matrix concept also applies to other ways of generating intervals. See my long comment below.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: business logic changes were intended in next PR, this should only changes config models.

modyn/config/schema/pipeline/evaluation/strategy/slicing.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import annotations
+
+from typing import Literal
```
robinholzi, line 3: Old Matrix Evaluation

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -0,0 +1,55 @@
...
+    datasets: List[str] = Field(
+        description=(
+            "List of dataset references defined at top level of evaluation config, to enable running one "
+            "handler on multiple datasets (e.g. train, test, combined)"
+        )
+    )
+    """
+    this makes sense as e.g. the BetweenTwoTriggersEvalStrategy can actually run on the training data.
+    we could default this to use run the handler on every of the datasets in the root setting
+    """
```
MaxiBoether, line 55: This feels out of place. Should this be above datasets?

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -0,0 +1,55 @@
...
+from pydantic import Field
+
+from .strategy import EvalStrategyConfig
+
+EvalHandlerExecutionTime = Literal["after_training"]
+"""
+after_training: will evaluate the models on the datasets after the training has finished
+after_pipeline: will evaluate the models on the datasets after the pipeline has finished
+manual: will skip the evaluations in the pipeline run; useful if someone wants to start them with a different
+    entry point later on in isolation (useful for debugging and when evaluation is not part of the performance critical
```
MaxiBoether, line 15: Should we add a note here this is for the future?

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    execution_time: EvalHandlerExecutionTime = Field(
+        description="When evaluations should be performed in terms of where in the code / when during a pipeline run."
+    )
+
+    models: Literal["matrix", "most_recent"] = Field(
+        "most_recent",
+        description="For a evaluation interval given by the strategy, this selects what models to evaluate there.",
+    )
```
MaxiBoether, line 40: We had a bit of discussion on this. Would be good to port the result of this here, either as a comment or into the description

MaxiBoether (approved):
Thanks Robin!

modyn/config/schema/pipeline/evaluation/handler.py
robinholzi: Python docstring are weird as they have to be specified below a variable. But this, indeed, feel out of context, I'll move it to the description.

PR #482, opened 2024-06-08
feat: Support pretrained weights for ResNet models from torchvision

MaxiBoether:
For CLOC/CGLM, @Sipondo suggested using a pretrained model. I somewhat agree this makes sense. Hence, I add support for loading the pretrained weights from torch. We need to call some internal functions here because we somewhat hijacked the implementation to record the embeddings. No unit tests for this because I do not want to download data in the tests, but I tested it locally and it works :)


robinholzi (approved):
👍

PR #483, opened 2024-06-08
feat: Warmup for data selection

MaxiBoether:
One concern I have with evaluating data selection is that for `n` triggers we should train on all the available data, and only then start using data selection, to make the model adjust somewhat towards the new data distribution first. This adds a warmup period to the downsampling strategy, which disables presampling and downsampling during that warmup.


modyn/config/schema/pipeline/sampling/config.py
```diff
@@ -95,6 +95,14 @@ class NewDataStrategyConfig(_BaseSelectionStrategy):
 class CoresetStrategyConfig(_BaseSelectionStrategy):
     name: Literal["CoresetStrategy"] = Field("CoresetStrategy")
 
+    warmup_triggers: int = Field(
+        description=(
+            "Defines for how many epochs no pre- or downsampling is performed. "
```
XianzheMa, line 100: Is this `how many triggers`? i.e. it controls on the trigger level not epoch level

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -73,12 +91,20 @@ def _get_data(self) -> Iterable[list[int]]:
         assert isinstance(
             self._storage_backend, DatabaseStorageBackend
         ), "CoresetStrategy currently only supports DatabaseBackend"
+        # We set is_warmup here as well to support unit tests more easily.
+        # In end-to-end workflows, it should have been set by trigger() already.
+        if (self._next_trigger_id + 1) > self.warmup_triggers:
+            self.is_warmup = False
 
+        presampling_strategy = self.warmup_presampler if self.is_warmup else self.presampling_strategy
+        assert presampling_strategy is not None  # satisfy mypy
```
XianzheMa, line 100: we can make this `self.warmup_presampler` always a `NoPresamplingStrategy`, so that we can get rid of this assert.

We use `self.is_warmup` to decide wether to use `self.warmup_presampler` anyway, no need to make `self.warmup_presampler` optional.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
```diff
@@ -71,7 +73,7 @@ def training_status_bar_scale(self) -> int:
         return self.current_downsampler.status_bar_scale
 
     def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: AbstractStorageBackend) -> None:
-        if self.next_threshold is not None and next_trigger_id >= self.next_threshold:
+        if self.next_threshold is not None and next_trigger_id - self._warmup_triggers >= self.next_threshold:
```
XianzheMa, line 76: can you add a comment in Pydantic config how `warmup_triggers` will affect the calculation of `downsampling_thresholds`?

modyn/selector/internal/selector_strategies/coreset_strategy.py
```diff
@@ -73,12 +91,20 @@ def _get_data(self) -> Iterable[list[int]]:
...
         trigger_dataset_size = None
-        if self.presampling_strategy.requires_trigger_dataset_size:
+        if presampling_strategy.requires_trigger_dataset_size:
             trigger_dataset_size = self._get_trigger_dataset_size()
 
-        stmt = self.presampling_strategy.get_presampling_query(
+        logger.error(trigger_dataset_size)
```
XianzheMa, line 105: maybe logger.info?

modyn/tests/selector/internal/selector_strategies/test_coreset_strategy.py
```diff
@@ -471,3 +471,107 @@ def fake_super_trigger(self):
...
+    strat = CoresetStrategy(conf, get_minimal_modyn_config(), 0)
+
+    data1 = list(range(10))
+    timestamps1 = list(range(10))
+    labels = [0] * 10
+    strat.inform_data(data1, timestamps1, labels)
+    current_data = list(strat._get_data())
+    assert len(current_data) == 10  # trigger 0 data = warmup, so no presampling
+    current_data = flatten(current_data)
+    assert set(current_data) <= set(data1)
```
XianzheMa, line 489: Shouldn't these two sets exactly equal to each other, as we don't select?

modyn/tests/selector/internal/selector_strategies/test_coreset_strategy.py
```diff
@@ -471,3 +471,107 @@ def fake_super_trigger(self):
...
+    # Now trigger id 1 starts
+    strat.trigger()
+
+    data2 = list(range(10, 20))
+    timestamps2 = list(range(10, 20))
+    strat.inform_data(data2, timestamps2, labels)
+    current_data = list(strat._get_data())
+    assert len(current_data) == 20  # presampling with tail 1 and still in warmup
+    current_data = flatten(current_data)
+    assert set(current_data) <= set(data1 + data2)
```
XianzheMa, line 500: same doubt here

integrationtests/selector/integrationtest_selector.py
```diff
@@ -491,6 +495,178 @@ def test_abstract_downsampler(reset_after_trigger: bool) -> None:
...
+
+        worker_1_weights = list(worker1_response.training_samples_weights)
+        worker_2_weights = list(worker2_response.training_samples_weights)
+        assert len(worker_1_samples) == len(worker_1_weights)
+        assert len(worker_2_samples) == len(worker_2_weights)
+
+        total_samples.extend(worker_1_samples + worker_2_samples)
+
+    assert len(total_samples) == len(set(total_samples)), "Received duplicated samples"
+    assert set(total_samples) <= set(
```
XianzheMa, line 643: I thought these two sets should be equal? As we don't select anything

XianzheMa (approved):
LGTM! Only some nits

modyn/config/schema/pipeline/sampling/config.py
MaxiBoether: yes! thanks for catching this

modyn/selector/internal/selector_strategies/coreset_strategy.py
MaxiBoether: true, fixed.

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
MaxiBoether: Will do (but note it does not affect it. This logic takes care that you don't need to think about it)

modyn/selector/internal/selector_strategies/coreset_strategy.py
MaxiBoether: actually this was leftover debugging code. thanks again for catching!

modyn/tests/selector/internal/selector_strategies/test_coreset_strategy.py
MaxiBoether: yes, this was copy pasta leftover

modyn/selector/internal/selector_strategies/downsampling_strategies/downsampling_scheduler.py
XianzheMa: Yes. but how the threshold takes care of now is *after* the warm up period.

MaxiBoether: Yes, exactly, but you don't need to adjust the indices or something. That's what I mean. Anyways, I hope the config comment is sufficient

PR #484, opened 2024-06-09
feat: Add config models for more `EvalStrategies`

robinholzi:
# Motivation

Motivated in https://github.com/eth-easl/modyn/pull/479 we now implement the configuration models for several new strategies.

Namely:
- `BetweenTwoTriggersEvalStrategyConfig`
- `PeriodicEvalStrategyConfig`
- `StaticEvalTriggerConfig`


modyn/config/schema/pipeline/evaluation/strategy/_interval_strategy.py
```diff
@@ -0,0 +1,139 @@
...
+from modyn.const.types import TimeUnit
+from pydantic import Field, field_validator, model_validator
+
+_EVAL_INTERVAL_BOUND_PATTERN = rf"[+-]?\s*((\d+\s*({REGEX_TIME_UNIT}))|(inf))"
+_EVAL_INTERVAL_PATTERN = (
+    rf"^(\[|\()\s*({_EVAL_INTERVAL_BOUND_PATTERN})\s*(,|;)\s*({_EVAL_INTERVAL_BOUND_PATTERN})\s*(\)|\])$"
+)
+
+
+class _IntervalEvalStrategyConfig(ModynBaseModel):
```
robinholzi, line 15: Used for `PeriodicEvalStrategyConfig`. Can potentially replace the boundary selection of `OffsetEvalStrategy`.
However `OffsetEvalStrategy` currently serves two purposes:
- explicit: `the interval selection via a one-sided offset`
- implicit: trigger `time` at every training (with the training data batch start and end as interval center).

I'd like to keep `OffsetEvalStrategy` for now, but generally it could be expressed as a subclass of `_IntervalEvalStrategyConfig` that explicitly decodes the evaluation time at the point of trigger in its name.

MaxiBoether: Could you clarify the implicit effect here? `OffsetEvalStrategy` can be used without ` after_training`, couldn't it? Or can it only be used after_training? I wonder -- based on our discussion yesterday -- how the `OffsetEvalStrategy` has an implicit effect on when we evaluate (at every training)

modyn/config/schema/pipeline/evaluation/strategy/between_two_triggers.py
```diff
@@ -0,0 +1,13 @@
+from typing import Literal
+
+from modyn.config.schema.base_model import ModynBaseModel
+from pydantic import Field
+
+
+class BetweenTwoTriggersEvalStrategyConfig(ModynBaseModel):
+    """This evaluation strategy will evaluate the model on the intervals between two consecutive triggers.
```
MaxiBoether, line 8: probably nit, but you between two consecutive triggers, there is just a single interval, right? and you use the plural of interval because we have multiple triggers?

modyn/config/schema/pipeline/evaluation/strategy/periodic.py
```diff
@@ -0,0 +1,28 @@
...
+
+from modyn.const.regex import REGEX_TIME_UNIT
+from modyn.utils.utils import SECONDS_PER_UNIT
+from pydantic import Field
+
+from ._interval_strategy import _IntervalEvalStrategyConfig
+
+
+class PeriodicEvalStrategyConfig(_IntervalEvalStrategyConfig):
+    """Comparable to TimeTriggerConfig and DataAmountTriggerConfig, but for scheduling evaluation runs."""
```
MaxiBoether, line 12: I don't get that docstring yet. Scheduling runs sounds to me like this defines when the evaluation is executed. Maybe we should update this docstring to describe how the intervals are generated

MaxiBoether (approved):
Mostly LGTM, had some nitpicky comments where the boundaries of trigger and strategy blur a little again. Thanks!

modyn/config/schema/pipeline/evaluation/strategy/_interval_strategy.py
robinholzi: the current version of `OffsetEvalStrategy` just takes `start_timestamp` and `end_timestamp` (the first and last timestamp of the training interval). Then it adds an offset to *ONE* of the intervals bounds of [start_timestamp, end_timestamp]. `_IntervalEvalStrategyConfig` is designed to do the same but is more flexible in its interface: 
- allows offsets at both ends
- allows `+0`, `-0` config syntax to refer to either `start_timestamp` or `end_timestamp`
- allows to specify if bounds are inclusive (through choice parenthesis).

This strategy can be executed after training (obviously we know `start_timestamp` and `end_timestamp` there. The `model=matrix` option does not make sense in this configuration though as we don't know the future `start/end_timestamps` just yet.

`after_training` we don't need to restrict the matrix option anymore.

I will add those restrictions and considerations directly into the config models, thanks for the hint!

PR #485, opened 2024-06-09
feat: Implement new evaluation setup in supervisor (part 1)

robinholzi:
# Motivation

Implement business logic for the new evaluation setup config models from https://github.com/eth-easl/modyn/pull/484

PR #489, opened 2024-06-10
feat: Implement eval strategies and support multiple eval handlers in super…

robinholzi:
# Motivation

part 2 of https://github.com/eth-easl/modyn/pull/485

# Changes

- Adjust EvalStrategy interface slightly
- Implement business logic for eval strategies
- Support multiple EvalHandlers in supervisor (`pipeline_executor.py`)
- Add tests for eval strategies

### DIFF
--- a/modyn/config/schema/pipeline/evaluation/handler.py
+++ b/modyn/config/schema/pipeline/evaluation/handler.py
@@ -33,16 +33,18 @@ class EvalHandlerConfig(ModynBaseModel):
     execution_time: EvalHandlerExecutionTime = Field(
         description="When evaluations should be performed in terms of where in the code / when during a pipeline run."
     )
-    """Caution: execution_time=`after_training` yields slightly different behavior for the `models` option.
+    """Caution: Using `execution_time=after_training` might lead to a slightly different behavior with respect to
+    the `models` option compared to the `execution_time=after_pipeline`.
 
-    Point of difference: decision, if a model is the most recent one for a certain eval request.
-    While `execution_time=after_pipeline` allows to consider all models for a certain eval request interval,
-    and thereby allows to select the most recent one, `execution_time=after_training` doesn't allow this.
-    This is because the models are generated during the pipeline run and the recently trained model might not be the
-    most recent model for the evaluation request that starts after the training interval. (as there might be another
-    training/model that lies between the current training and the next evaluation interval).
-    So for `execution_time=after_training` there might be multiple models that are considered `most_recent` for an
-    interval."""
+    Point of Difference: Determining the 'most recent' model for a given evaluation request.
+
+    When `execution_time` is set to `after_pipeline`, all models generated within a certain evaluation request interval
+    are considered. This allows the selection of the most recent model within that interval.
+    However, when `execution_time` is set to `after_training`, the selection process is different. Models are generated
+    during the pipeline run, and the most recently trained model may not necessarily be the most recent for the
+    subsequent evaluation request. This is due to the possibility of another model being trained between the current
+    training and the next evaluation interval. As a result, with `execution_time=after_training`, there could be
+    multiple models deemed as 'most recent' for a given interval."""
 
     models: Literal["matrix", "most_recent"] = Field(
         "most_recent",
--- a/modyn/config/schema/pipeline/evaluation/handler.py
+++ b/modyn/config/schema/pipeline/evaluation/handler.py
@@ -33,6 +33,16 @@ class EvalHandlerConfig(ModynBaseModel):
     execution_time: EvalHandlerExecutionTime = Field(
         description="When evaluations should be performed in terms of where in the code / when during a pipeline run."
     )
+    """Caution: execution_time=`after_training` yields slightly different behavior for the `models` option.
+
+    Point of difference: decision, if a model is the most recent one for a certain eval request.
+    While `execution_time=after_pipeline` allows to consider all models for a certain eval request interval,
+    and thereby allows to select the most recent one, `execution_time=after_training` doesn't allow this.
+    This is because the models are generated during the pipeline run and the recently trained model might not be the
+    most recent model for the evaluation request that starts after the training interval. (as there might be another
+    training/model that lies between the current training and the next evaluation interval).
+    So for `execution_time=after_training` there might be multiple models that are considered `most_recent` for an
+    interval."""
 
     models: Literal["matrix", "most_recent"] = Field(
         "most_recent",
@@ -67,4 +77,5 @@ class EvalHandlerConfig(ModynBaseModel):
     def validate_datasets(self) -> EvalHandlerConfig:
         if isinstance(self.strategy, OffsetEvalStrategyConfig):
             assert self.execution_time == "after_training", "OffsetEvalStrategy can only be used after training."
+            assert self.models == "most_recent", "OffsetEvalStrategy can only be used with most_recent model currently."
         return self
--- a/modyn/config/schema/pipeline/evaluation/strategy/static.py
+++ b/modyn/config/schema/pipeline/evaluation/strategy/static.py
@@ -4,11 +4,11 @@ from modyn.config.schema.base_model import ModynBaseModel
 from pydantic import Field
 
 
-class StaticEvalTriggerConfig(ModynBaseModel):
+class StaticEvalStrategyConfig(ModynBaseModel):
     """A sequence of intervals at which the evaluation should be performed."""
 
     type: Literal["StaticEvalStrategy"] = Field("StaticEvalStrategy")
-    intervals: set[tuple[int, int | None]] = Field(
+    intervals: list[tuple[int, int | None]] = Field(
         description=(
             "List of intervals in which the evaluation should be performed."
             "Each interval is a tuple of two timestamps. Use open end."
--- a/modyn/supervisor/internal/eval/handler.py
+++ b/modyn/supervisor/internal/eval/handler.py
@@ -1,15 +1,74 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
-
+import pandas as pd
 from modyn.config.schema.pipeline import EvalHandlerConfig
+from modyn.supervisor.internal.eval.strategies.abstract import AbstractEvalStrategy
+from modyn.utils import dynamic_module_import
+from pydantic import BaseModel
+
+eval_strategy_module = dynamic_module_import("modyn.supervisor.internal.eval.strategies")
 
 
-@dataclass
+class EvalRequest(BaseModel):
+    """A complete request for evaluation."""
+
+    trigger_id: int
+    training_id: int
+    model_id: int
+    most_recent_model: bool
+    eval_handler: str
+    dataset_id: str
+    interval_start: int | None
+    interval_end: int | None
+
+
 class EvalHandler:
-    """Handles one series of evaluations configured through an `EvalHandlerConfig`.
+    """Handles one series of evaluations configured through an `EvalHandlerConfig`."""
 
-    We will add more functionality to this class in the future.
-    """
+    def __init__(self, config: EvalHandlerConfig):
+        self.config = config
+        strat_conf = config.strategy
+        self.eval_strategy: AbstractEvalStrategy = getattr(eval_strategy_module, strat_conf.type)(strat_conf)
 
-    config: EvalHandlerConfig
+    def get_eval_requests_after_training(
+        self, trigger_id: int, training_id: int, model_id: int, training_interval: tuple[int, int]
+    ) -> list[EvalRequest]:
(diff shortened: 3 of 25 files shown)


modyn/config/schema/pipeline/evaluation/strategy/static.py
```diff
@@ -4,11 +4,11 @@
 from pydantic import Field
 
 
-class StaticEvalTriggerConfig(ModynBaseModel):
+class StaticEvalStrategyConfig(ModynBaseModel):
     """A sequence of intervals at which the evaluation should be performed."""
 
     type: Literal["StaticEvalStrategy"] = Field("StaticEvalStrategy")
-    intervals: set[tuple[int, int | None]] = Field(
+    intervals: list[tuple[int, int | None]] = Field(
```
robinholzi, line 11: It is more convenient if we can rely on ordering.

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -33,6 +33,16 @@ class EvalHandlerConfig(ModynBaseModel):
     execution_time: EvalHandlerExecutionTime = Field(
         description="When evaluations should be performed in terms of where in the code / when during a pipeline run."
     )
+    """Caution: execution_time=`after_training` yields slightly different behavior for the `models` option.
+
+    Point of difference: decision, if a model is the most recent one for a certain eval request.
```
MaxiBoether, line 38: This sentence reads a bit weird, can we adjust that? The text afterwards i like

modyn/config/schema/pipeline/evaluation/handler.py
```diff
@@ -33,6 +33,16 @@ class EvalHandlerConfig(ModynBaseModel):
...
+    While `execution_time=after_pipeline` allows to consider all models for a certain eval request interval,
+    and thereby allows to select the most recent one, `execution_time=after_training` doesn't allow this.
+    This is because the models are generated during the pipeline run and the recently trained model might not be the
+    most recent model for the evaluation request that starts after the training interval. (as there might be another
+    training/model that lies between the current training and the next evaluation interval).
+    So for `execution_time=after_training` there might be multiple models that are considered `most_recent` for an
+    interval."""
```
MaxiBoether, line 45: Just to confirm my understanding: For each model, we loop over all intervals. They are always the same for all models. However, if we have a new model, it might become the most recent model for an interval where we already evaluated a model on, and hence we evaluate a model on this interval again. So in the worst case, this setting degrades to "matrix" evaluation as well?

modyn/supervisor/internal/eval/handler.py
```diff
@@ -1,15 +1,74 @@
...
+import pandas as pd
 from modyn.config.schema.pipeline import EvalHandlerConfig
+from modyn.supervisor.internal.eval.strategies.abstract import AbstractEvalStrategy
+from modyn.utils import dynamic_module_import
+from pydantic import BaseModel
+
+eval_strategy_module = dynamic_module_import("modyn.supervisor.internal.eval.strategies")
+
+
+class EvalRequest(BaseModel):
```
MaxiBoether, line 12: Any reason for using pydantic for internal state here as well instead of the previously used dataclass?

modyn/supervisor/internal/eval/handler.py
```diff
@@ -1,15 +1,74 @@
...
+                Can be acquired by joining TriggerExecutionInfo and StoreModelInfo dataframes.
 
-    We will add more functionality to this class in the future.
-    """
+        Returns:
+            list of evaluation requests with first and last timestamps being the point of trigger, not the interval.
+        """
+        assert {"trigger_id", "training_id", "id_model", "first_timestamp", "last_timestamp"} - (
+            set(df_trainings.columns)
+        ) == set()
```
MaxiBoether, line 72: Why this? I thought the dataframes are used for internal tracking/logging, but now we use it for business logic?

modyn/supervisor/internal/eval/strategies/_interval.py
```diff
@@ -66,4 +57,4 @@ def get_eval_intervals(
                 + int(self.config.right) * SECONDS_PER_UNIT[self.config.right_unit]
                 - int(not self.config.right_bound_inclusive)
             )
-        yield max(0, left), max(left, right) if right else right
+        return max(0, left), max(left, right) if right else right
```
MaxiBoether, line 60: can this ever return None (according to the type specification it can, but we init left and right with 0 and apply max operations on them)

modyn/supervisor/internal/eval/strategies/between_two_trainings.py
```diff
@@ -1,15 +1,22 @@
-from typing import Iterable, Optional
```
MaxiBoether, line 1: note: the file is named between_two_trainings but the class between two triggers

modyn/supervisor/internal/eval/strategies/abstract.py
```diff
@@ -19,9 +35,7 @@ def __init__(self, config: EvalStrategyConfig):
         self.config = config
 
     @abstractmethod
-    def get_eval_intervals(
-        self, first_timestamp: int, last_timestamp: int
-    ) -> Iterable[tuple[Optional[int], Optional[int]]]:
+    def get_eval_intervals(self, training_intervals: Iterable[tuple[int, int]]) -> Iterable[EvalInterval]:
```
MaxiBoether, line 38: I guess I don't fully get the interface change yet. We used to pass the timestamps of the samples in the trigger. what do we pass now and why?

modyn/supervisor/internal/eval/strategies/abstract.py
```diff
@@ -1,9 +1,25 @@
...
+    """The start timestamp of the evaluation interval handed over to the evaluation server."""
+
+    end: int | None
+    """The end timestamp of the evaluation interval handed over to the evaluation server."""
+
+    training_interval_start: int
+    """The first timestamp in the training request."""
+
+    training_interval_end: int
+    """The last timestamp in the training request."""
```
MaxiBoether, line 20: I do not understand this yet. Why is this part of eval interval? Why can this not be None, but start and end can? From my understanding, any eval interval was some slice of a dataset, but now we have 4 parameters

modyn/supervisor/internal/eval/strategies/between_two_trainings.py
```diff
@@ -1,15 +1,22 @@
...
-    ) -> Iterable[tuple[Optional[int], Optional[int]]]:
-        raise NotImplementedError()
+    def get_eval_intervals(self, training_intervals: Iterable[tuple[int, int]]) -> Iterable[EvalInterval]:
+        return [
+            EvalInterval(
+                start=interval_start,
+                end=interval_end,
+                # This is independent of the training interval, we use the interval centers as training intervals
+                training_interval_start=(interval_end + interval_start) // 2,
+                training_interval_end=(interval_end + interval_start) // 2,
```
MaxiBoether, line 19: i unfortunately do not understand that yet, but that probably lies in the fact that I don't get the new evalinterval class yet

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -717,110 +726,89 @@ def _evaluate_and_store_results(
...
+
+            handler_eval_requests = eval_handler.get_eval_requests_after_training(
+                trigger_id=trigger_id,
+                training_id=training_id,
+                model_id=model_id,
+                training_interval=(first_timestamp, last_timestamp),
+            )
+            eval_requests += handler_eval_requests
+
+        # will be replaced with threadpool in a follow-up PR
```
MaxiBoether, line 743: I am a bit torn whether we need this on main. On the one hand, for now this greatly speeds up the results because we paralleliez eval, on the other hand, when we parallelize eval in the future we want to do it at the evaluator itself, and then we do not need the thread pool anymore but instead can just submit requests in a quick loop

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -203,6 +208,10 @@ def run(self) -> None:
             self._serve_online_data(self.state, self.logs)
 
         self._done(self.state, self.logs)
+
+        self._post_pipeline_evaluation_checkpoint(self.state, self.logs)
```
MaxiBoether, line 212: I don't understand yet what exactly you plan to checkpoint here.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -832,6 +820,18 @@ def _done(self, s: ExecutionState, log: StageLog) -> None:
         self.logs.pipeline_stages = _pipeline_stage_parents  # now includes chronology info
         self.logs.materialize(s.log_directory, mode="final")
 
+    # post core pipeline evaluation
```
MaxiBoether, line 823: What is "core pipeline evaluation"?

MaxiBoether (changes requested):
Thanks Robin! I mostly had questions on EvalInterval and some interface stuff, after that we should hopefully be good to go

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: Technically, the steps after this comment are still part of the pipeline, so I call the steps prior to that `core pipeline` (the main parts that contains trainings etc.). I'll remove this naming convention and comment though.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: The information about the different triggers/trainings/models. E.g. what the first and last timestamp of a training request was, to what model_id it lead etc.
In the next PR that should become much clearer.
The goal behind this is to allow to restart the pipeline at this point. This will be useful when we adjust the evaluation logic and cannot guarantee that it won't fail. With this we don't need to run training again

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: In the next PR i will move the evaluation logic into an `evaluation_executor.py` so it doesn't mess up the `pipeline_executor.py`. There we can simply remove that again in the future, it will only be a couple of lines in a dedicated function, that is responsible for executing a list of evaluation requests.

modyn/config/schema/pipeline/evaluation/handler.py
robinholzi: For the matrix it doesn't make a big difference (just the logged `is_most_recent_model` flag might be a bit off), for non-matrix the `is_most_recent_model` is used for filtering so there might be more evaluations being launched than in the `after_pipeline` case.

modyn/supervisor/internal/eval/handler.py
robinholzi: I originally used a dataclass but then realized that I want to log exactly the content of this structure (previously this was all duplicated in a pydantic logging model). I decided to just make this a pydantic model and use it for logging directly.

modyn/supervisor/internal/eval/strategies/between_two_trainings.py
robinholzi: Thanks!

modyn/supervisor/internal/eval/strategies/abstract.py
robinholzi: For the `after_training` case the same, just wrapped in a list of length 1 of 2-tuples (start, end).
For the `after_pipeline` case, we pass the list of all of those tuple at once (that have been handled in `after_training` via multiple calls the course of the pipeline)

modyn/supervisor/internal/eval/strategies/abstract.py
robinholzi: yeah, this is weird, see my comment below. We might want to discuss this in person.

modyn/config/schema/pipeline/evaluation/handler.py
MaxiBoether: ok then I think I understood it correctly. Thanks!

modyn/supervisor/internal/eval/handler.py
MaxiBoether: Ah ok, I somewhat did not realize we have dataframes in the business logic. i suppose i would have argued against that since I like to have them only for data wrangling, but i suppose we already have them now

MaxiBoether (approved):

modyn/supervisor/internal/eval/handler.py
robinholzi: we don't rely on them that much so adjusting wouldn't be to hard. They just make the implementation (filtering / joining etc.) much more convenient and efficient

PR #490, opened 2024-06-10
feat: Add `EvaluationExecutor` and `after_pipeline` evaluation 

robinholzi:
# Motivation

We want to condense the `PipelineExecutor` and therefore move the evaluation logic to a dedicated `EvaluationExecutor`.

Also, we want to be able to run evaluations after the pipeline and be able to restart the evaluation process without having to retrain.


modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,352 @@
...
+            if eval_result_writer.results["datasets"]
+            else {}
+        )
+
+
+# ------------------------------------------------------------------------------------ #
+#                                       DevTools                                       #
+# ------------------------------------------------------------------------------------ #
+
+if __name__ == "__main__":
```
robinholzi, line 334: Is there a better place to put this?

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -719,98 +714,17 @@ def _evaluate_and_store_results(
         last_timestamp: int,
     ) -> None:
         """Evaluate the trained model and store the results."""
-        # pylint: disable=too-many-locals
```
robinholzi, line 722: `_single_evaluation` and `_store_evaluation_results` are now combined into one function `EvaluationExecutor._single_evaluation`

modyn/supervisor/internal/eval/handler.py
```diff
@@ -71,4 +71,55 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
+        # now build & potentially reduce the Trainings x EvalIntervals space
+        df_cross = df_eval_intervals.merge(df_trainings, how="cross")
+
+        # to check if a combination is the most recent, we first compute if model was trained before the usage starts
+        # last_timestamp (df_trainings) defines the end of the training data;
+        # most_recent_model_interval_end_before (df_eval_intervals): defines center of an eval intervals.
+        df_cross["chronology_ok"] = df_cross["last_timestamp"] <= df_cross["most_recent_model_interval_end_before"]
+
+        # find the maximum model for every EvalCandidate that doesn't violate that constraint
+        max_model_id = (
```
robinholzi, line 95: I think, in this instance dataframes are well suited for business logic as their expressiveness is so much higher for these kinds of operations.

modyn/protos/supervisor.proto
```diff
@@ -33,11 +33,12 @@ message PipelineStageDatasetMsg {
...
 
 message PipelineStageCounterCreateParams {
-    int32 new_data_len = 1;
+    string title = 1;
+    optional int32 new_data_len = 2;
 }
 
 message PipelineStageCounterUpdateParams {
-    int32 batch_size = 1;
+    int32 increment = 1;
```
MaxiBoether, line 41: this might be causing failures in the integration test :)

modyn/supervisor/internal/eval/handler.py
```diff
@@ -71,4 +71,55 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
+        # combine: a model in the cross product is most recent for a certain interval iff
+        #  it has maximum model id for its most_recent_model_interval_end_before
+        final = df_cross.merge(max_model_id, on="most_recent_model_interval_end_before")
+        final["most_recent_model"] = final["id_model"] == final["max_model_id"]
+
+        # convert to list
+        eval_requests: list[EvalRequest] = []
+        for _, eval_candidate in final.iterrows():
+            if self.config.models != "matrix" and not eval_candidate["most_recent_model"]:
+                continue
```
MaxiBoether, line 110: In this case, we can just return before we even iterrows()

modyn/supervisor/internal/eval/handler.py
```diff
@@ -71,4 +71,55 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
+            for dataset_id in self.config.datasets:
+                eval_requests.append(
+                    EvalRequest(
+                        trigger_id=eval_candidate["trigger_id"],
+                        training_id=eval_candidate["training_id"],
+                        model_id=eval_candidate["id_model"],
+                        most_recent_model=eval_candidate["most_recent_model"],
```
MaxiBoether, line 117: I guess I am still a little confused here. Why do we need to send the most recent model here? An Evalrequests says: Please evaluate model x on dataset y in interval (a,b). I am a bit sure why the additional info (training, trigger, most_recent_model) is needed

modyn/supervisor/internal/eval/handler.py
```diff
@@ -71,4 +71,55 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
+
+        # now build & potentially reduce the Trainings x EvalIntervals space
+        df_cross = df_eval_intervals.merge(df_trainings, how="cross")
+
+        # to check if a combination is the most recent, we first compute if model was trained before the usage starts
+        # last_timestamp (df_trainings) defines the end of the training data;
+        # most_recent_model_interval_end_before (df_eval_intervals): defines center of an eval intervals.
+        df_cross["chronology_ok"] = df_cross["last_timestamp"] <= df_cross["most_recent_model_interval_end_before"]
+
+        # find the maximum model for every EvalCandidate that doesn't violate that constraint
```
MaxiBoether, line 94: just for my understanding, we could also find the model with the maximum last_timestamp right?

modyn/tests/supervisor/internal/eval/test_eval_handler.py
```diff
@@ -0,0 +1,94 @@
+from typing import Iterable
+
+import pandas as pd
+from modyn.config.schema.pipeline.evaluation.handler import EvalHandlerConfig
+from modyn.config.schema.pipeline.evaluation.strategy.static import StaticEvalStrategyConfig
+from modyn.supervisor.internal.eval.handler import EvalHandler
+from modyn.supervisor.internal.eval.strategies.abstract import AbstractEvalStrategy, EvalInterval
+
+
+def test_get_eval_requests_after_pipeline() -> None:
```
MaxiBoether, line 10: Can you please add a test with the same setup (matrix/most_recent) for after_training as well? This test helped much in understanding the output, so it would be good to have. Thank you!

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -823,12 +735,22 @@ def _done(self, s: ExecutionState, log: StageLog) -> None:
     @pipeline_stage(PipelineStage.POST_EVALUATION_CHECKPOINT, parent=PipelineStage.MAIN, log=False, track=False)
     def _post_pipeline_evaluation_checkpoint(self, s: ExecutionState, log: StageLog) -> None:
         """Stores evaluation relevant information so that the evaluator can be started on this pipeline run again."""
-        # we will implement this in a follow-up PR
+
+        if not s.pipeline_config.evaluation:
```
MaxiBoether, line 739: can we ignore  s.pipeline_config.evaluation.skip_during_pipeline_execution here?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+            else []
+        )
+
+    def register_tracking_info(self, tracking_dfs: dict[str, pd.DataFrame]) -> None:
+        assert tracking_dfs.get(PipelineStage.HANDLE_SINGLE_TRIGGER.name) is not None
+        assert tracking_dfs.get(PipelineStage.STORE_TRAINED_MODEL.name) is not None
+        self.context = AfterPipelineEvalContext(tracking_dfs=tracking_dfs)
+
+    def create_snapshot(self) -> None:
+        """Create a snapshot of the evaluation state before starting to evaluate."""
```
MaxiBoether, line 79: I guess this is more like training/pipeline metadata? Evaluation state to me sounds  very much like we snapshot the state of the evaluation (partial progress).

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+        # create tempdir if snapshot_dir is None:
+        snapshot_dir.mkdir(parents=True, exist_ok=True)
+
+        # write state: config, pipeline & context
+        eval_state_config = EvalStateConfig(
+            pipeline_id=self.pipeline_id, eval_dir=self.pipeline_logdir, config=self.config, pipeline=self.pipeline
+        )
+        (snapshot_dir / "eval_state.yaml").write_text(eval_state_config.model_dump_json(by_alias=True))
+        (snapshot_dir / "context.pcl").write_bytes(pickle.dumps(self.context))
+
```
MaxiBoether, line 94: If my understanding is correct that this is pipeline metadata, can't we write this into the pipeline logfile and then just restart the evaluation using the pipeline logfile? I'd try avoiding creating many files per pipeline and rather have one condensed summary of the pipeline

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: Good question. I'd say for now it's probably best to have this here...

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+    logs_ = PipelineLogs(
+        pipeline_id=ex.pipeline_id,
+        pipeline_stages={},
+        config=ConfigLogs(system=ex.config, pipeline=ex.pipeline),
+        experiment=True,
+        supervisor_logs=SupervisorLogs(),
+    )
+
+    logs_.supervisor_logs = ex.run_post_pipeline_evaluations(eval_status_queue=Queue())
+    logs_.materialize(snapshot_path, mode="final")
```
MaxiBoether, line 331: will this overwrite existing log files? can we somehow merge that/extend the existing pipeline log file or is that a big effort?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+        reporter.create_tracker()
+        self.grpc.wait_for_evaluation_completion(eval_req.training_id, evaluation)
+
+        eval_result_writer = JsonResultWriter(self.pipeline_id, eval_req.trigger_id, self.pipeline_logdir)
+        self.grpc.store_evaluation_results([eval_result_writer], evaluation)
+        self.grpc.cleanup_evaluations([int(i) for i in evaluation])
+        assert isinstance(eval_result_writer, JsonResultWriter)
+
+        log.info.results = (
+            eval_result_writer.results["datasets"][0][dataset_config.dataset_id]
```
MaxiBoether, line 303: why the [0]?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+            eval_requests += handler_eval_requests
+
+        num_workers = None
+        if self.config.supervisor:
+            num_workers = self.config.supervisor.after_training_evaluation_workers
+        logs = self._launch_evaluations_async(eval_requests, log, eval_status_queue, num_workers)
+        return logs
+
+    def run_post_pipeline_evaluations(self, eval_status_queue: Queue) -> SupervisorLogs:
+        """Evaluate the trained models after the core pipeline and store the results."""
```
MaxiBoether, line 166: I wonder why we need two separate functions. Why can't we iterate over all handlers both post training and post pipeline and skip those according to our current stage? Is there a fundamental problem in executing the handler eiter post training or post pipeline in the same function?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+
+        logs = self._launch_evaluations_async(
+            eval_requests,
+            parent_log=StageLog(
+                id=PipelineStage.EVALUATE_SINGLE.name,
+                start=datetime.datetime.now(),
+                batch_idx=-1,
+                sample_idx=-1,
+                sample_time=-1,
+                trigger_idx=-1,
```
MaxiBoether, line 201: We should add a comment explaining this, it at least looks weird. Why don't we care about all of this post pipeline but care about it post training?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+                trigger_idx=parent_log.trigger_idx,
+            )
+            epoch_nanos_start = current_time_micros()
+            self._single_evaluation(single_log, eval_status_queue, eval_req)
+            single_log.end = datetime.datetime.now()
+            single_log.duration = datetime.timedelta(microseconds=current_time_micros() - epoch_nanos_start)
+            return single_log
+
+        # As we are io bound by the evaluator server, GIL locking isn't a concern, so we can use multithreading.
+        with ThreadPoolExecutor(max_workers=num_workers or 1) as pool:
```
MaxiBoether, line 245: if we have 1 minimum worker anyways, we could also just expect an int and where we call this do `num_workers = config.supervisor.workers if config.supervisor is not None else 1` or sth

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+                sample_time=parent_log.sample_time,
+                trigger_idx=parent_log.trigger_idx,
+            )
+            epoch_nanos_start = current_time_micros()
+            self._single_evaluation(single_log, eval_status_queue, eval_req)
+            single_log.end = datetime.datetime.now()
+            single_log.duration = datetime.timedelta(microseconds=current_time_micros() - epoch_nanos_start)
+            return single_log
+
+        # As we are io bound by the evaluator server, GIL locking isn't a concern, so we can use multithreading.
```
MaxiBoether, line 244: I am actually surprised this works, because using gRPC in a multithreaded fashion at least used to be a problem. But I also did not encounter issues so far, so until we fix the evaluator, let's keep this

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -0,0 +1,331 @@
...
+        def worker_func(eval_req: EvalRequest) -> StageLog:
+            single_log = StageLog(
+                id=PipelineStage.EVALUATE_SINGLE.name,
+                start=datetime.datetime.now(),
+                batch_idx=parent_log.batch_idx,
+                sample_idx=parent_log.sample_idx,
+                sample_time=parent_log.sample_time,
+                trigger_idx=parent_log.trigger_idx,
+            )
+            epoch_nanos_start = current_time_micros()
```
MaxiBoether, line 238: nanos or micros :D?

MaxiBoether (changes requested):
Thanks Robin! I had a couple of comments but nothing major, I think

modyn/protos/supervisor.proto
robinholzi: yes, forgot copy the adjustments in the client from my other branch

modyn/supervisor/internal/eval/handler.py
RobinHolzingerQC: Yes, but we have to do that for every interval under the `last_timestamp <= most_recent_model_interval_end_before` constraint. So we are essentially doing it already.

modyn/supervisor/internal/eval/handler.py
RobinHolzingerQC: why? of course we could filter the most recent models already in the dataframe, but I'm not sure if that would make things simpler.

modyn/supervisor/internal/eval/handler.py
robinholzi: Maybe it's the name that is confusing you. `EvalRequest` is not the structure that is handed over to the evaluator but describes one semantic unit of evaluation (some of the fields are only relevant for logging/plotting)

modyn/tests/supervisor/internal/eval/test_eval_handler.py
robinholzi: ah yes, sure!

robinholzi: the `most_recent_model` determination in the `after_training` case still had conceptual issues, thanks for the pointer, the unit test helped to find those :)

modyn/supervisor/internal/eval/handler.py
MaxiBoether: ah sorry, I was stupid. nevermind.

modyn/supervisor/internal/eval/handler.py
MaxiBoether: Okay, makes sense, I somewhat confused it for a gRPC message since we typically use the suffix request for that

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
robinholzi: that's an artifact from a previous version. The field `skip_during_pipeline_execution` doesn't even exist anymore.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: agreed

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: Ah thanks, that was already wrong in the pipeline_executor xD

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: Hmm, not sure, we have the information available in different formats (dataframe with multiple models vs. single model as primitives), also we have to wrap the logging information differently and have to call different methods of the `EvalHandler` (`get_eval_requests_after_pipeline`, `get_eval_requests_after_training`). So I'd argue it's fair to have two functions for this. There are mostly wrapper functions, the shared logic is in `_launch_evaluations_async` and `_launch_evaluations_async` anyways.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: I think this has been like this before I touched it. One eval_result_writer is designed to handle multiple requests iirc, but we only use it for one.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: So far, I just did that manually. I didn't want this could destroy my main log file. It's enough to simply append the `stage_runs` section from one file to the other. Once this logic has proven itself I am happy to merge the two logfiles automatically.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: Ah i see. the whole result writer thing is a bit shaky anyways because there is duplication here and the pipeline log, but ok thanks for clearing that up!

modyn/supervisor/internal/eval/handler.py
robinholzi: We could call it `EvalBundle` if that makes things more clearer?

modyn/tests/supervisor/internal/eval/test_eval_handler.py
MaxiBoether: perfect!

MaxiBoether (approved):
Thanks Robin, this looks good!

PR #491, opened 2024-06-10
RHO-LOSS Step 4: Implement `RHOLossRemoteDownsampler`

XianzheMa:
This PR implements the remote RHO-LOSS downsampler at trainer server, thus finishing the implementation of RHO-LOSS.

## How to review
The number of touched files is huge but most of them are nominal changes. Please first take a look at [remote_rho_loss_downsampling.py](https://github.com/eth-easl/modyn/pull/491/files#diff-a205a5bf35be8d9152f1f39cfb88a1c4c7f81cb2a1ffb2620cfecf31bca5313b), then at [irreducible_loss_producer.py](https://github.com/eth-easl/modyn/pull/491/files#diff-8a9628f5487d18d2c309c1d1a9061e0adad71fe5e26d27da2a5051bba019c83c), then at their corresponding unit test files, then at rest files.

## What's left
I still need an integration test to connect all pieces together as this selection policy is non-trivial.


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -545,10 +546,10 @@ def downsample_batch(
         self._downsampler.init_downsampler()
         self.start_embedding_recording_if_needed()
 
-        with torch.inference_mode(mode=(not self._downsampler.requires_grad)):
```
XianzheMa, line 548: pylint complains about superfluous parentheses

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
```diff
@@ -22,7 +22,7 @@ def get_tensors_subset(
     # then we extract the data
     if isinstance(data, torch.Tensor):
         sub_data = data[in_batch_index]
-    elif isinstance(data, dict):
+    else:
         sub_data = {key: tensor[in_batch_index] for key, tensor in data.items()}
```
XianzheMa, line 27: data is type-hinted with a `Union` so only two possibilities.

Also pylint complains about potential uninitialized `sub_data` otherwise

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -39,20 +39,37 @@ def __init__(
...
         self.data_config = data_config
-        self.il_model_id: Optional[int] = None
 
     def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: AbstractStorageBackend) -> None:
         if not isinstance(selector_storage_backend, DatabaseStorageBackend):
             raise ValueError("RHOLossDownsamplingStrategy requires a DatabaseStorageBackend")
 
         self._prepare_holdout_set(next_trigger_id, selector_storage_backend)
-        self.il_model_id = self._train_il_model(next_trigger_id)
+        self._train_il_model(next_trigger_id)
```
XianzheMa, line 48: we couldn't simply store the `il_model_id` for use later, because there are multiple selector instances per pipeline, the request to fetch downsampling config can be routed to another unsynced downsampler.

So we have to fetch data from database.

modyn/trainer_server/internal/dataset/key_sources/selector_key_source.py
```diff
@@ -105,16 +105,12 @@ def _connect_to_selector(self) -> SelectorStub:
...
-            # no connection established
-            if attempt < max_retries:
-                logger.info(f"gRPC connection attempt {attempt} failed. Retrying in {retry_delay} seconds...")
-                time.sleep(retry_delay)
-                retry_delay *= 2  # Exponential backoff
-            else:
-                logger.error(f"Failed to establish gRPC connection after {max_retries} attempts.")
-                raise ConnectionError(
-                    f"Could not establish gRPC connection to selector at address {self._selector_address}."
-                )
```
XianzheMa, line 117: don't know how it went through pylint before.
But with the previous code, it's possible for this function to return inconsistent data types.

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
MaxiBoether: i am confused why your pylint complains about different stuff than mine, but ok :D

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
```diff
@@ -0,0 +1,104 @@
+import io
```
MaxiBoether, line 1: Can we restructure the remote_downsamplers directory to have a subdirectory rholoss, where this and the remote_rs2_downsampler files go? in __init__.py of remote_downsamplers we can then import the RS2 Downsampler such that 

from trainer.remote_downsamplers import RS2Downsampler

is still possible. This avoids that this file lives on the same level as the other downsamplers

modyn/trainer_server/internal/dataset/key_sources/selector_key_source.py
MaxiBoether: I don't understand why it would be inconsistent (we either return or raise), but can you also adapt the online_dataset, then? we have the same problem there

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_uncertainty_downsampling_strategy.py
```diff
@@ -15,45 +16,45 @@ def get_sampler_config(balance=False):
...
 
-def test_collect_scores():
-    amds = RemoteUncertaintyDownsamplingStrategy(*get_sampler_config())
+def test_collect_scores(dummy_system_config: ModynConfig):
+    amds = RemoteUncertaintyDownsamplingStrategy(*get_sampler_config(dummy_system_config))
     with torch.inference_mode(mode=(not amds.requires_grad)):
         first_output = torch.randn((4, 5))
         second_output = torch.randn((3, 5))
-        amds.inform_samples([1, 2, 3, 4], first_output, None, None)
+        amds.inform_samples([1, 2, 3, 4], None, first_output, None, None)
```
MaxiBoether, line 36: This affects several files, but the expected type is `Union[dict[str, torch.Tensor], torch.Tensor]`. I don't understand why mypy does not complain about `None`. I whether we should rather test with dummy input instead?

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
MaxiBoether: Good catch!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
```diff
@@ -0,0 +1,69 @@
...
+        sample_ids: list[int],
+        forward_input: Union[dict[str, torch.Tensor], torch.Tensor],
+        forward_output: torch.Tensor,
+        target: torch.Tensor,
+        embedding: Optional[torch.Tensor] = None,
+    ) -> None:
+        training_loss = self.per_sample_loss_fct(forward_output, target).detach()
+        self.index_sampleid_map += sample_ids
+        irreducible_loss = self.irreducible_loss_producer.get_irreducible_loss(sample_ids, forward_input, target)
+        self.rho_loss = torch.cat([self.rho_loss, training_loss - irreducible_loss]).to(training_loss.dtype)
```
MaxiBoether, line 54: Why is the type conversation here necessary? Shouldn't all loss tensors be of the same dtype?

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def select_points(self) -> tuple[list[int], torch.Tensor]:
+        target_size = max(int(self.downsampling_ratio * self.number_of_points_seen / 100), 1)
+        # find the indices of maximal "target_size" elements in the list of rho_loss
+        selected_indices = torch.topk(self.rho_loss, target_size).indices
+        # use sorted() because we keep the relative order of the selected samples
+        selected_sample_ids = [self.index_sampleid_map[i] for i in sorted(selected_indices)]
+        # all selected samples have weight 1.0
+        selected_weights = torch.ones(target_size)
+        return selected_sample_ids, selected_weights
```
MaxiBoether, line 65: Something I just realized is that the RHO-LOSS paper does Batch-then-sample, however it does not accumulate batches which seems odd to me. Let's stick with what we have for now, and then see how it performs.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
```diff
@@ -0,0 +1,69 @@
...
+        training_loss = self.per_sample_loss_fct(forward_output, target).detach()
+        self.index_sampleid_map += sample_ids
+        irreducible_loss = self.irreducible_loss_producer.get_irreducible_loss(sample_ids, forward_input, target)
+        self.rho_loss = torch.cat([self.rho_loss, training_loss - irreducible_loss]).to(training_loss.dtype)
+        self.number_of_points_seen += forward_output.shape[0]
+
+    def select_points(self) -> tuple[list[int], torch.Tensor]:
+        target_size = max(int(self.downsampling_ratio * self.number_of_points_seen / 100), 1)
+        # find the indices of maximal "target_size" elements in the list of rho_loss
+        selected_indices = torch.topk(self.rho_loss, target_size).indices
```
MaxiBoether, line 60: In a follow up PR, we could also try, instead of top k selection as in the paper, to sample based on the loss as weights, so similar to the LossDownsampler where the higher the RHO-LOSS here, the higher the probability of being sampled, but no top k. We could add a config option for this. Let's also add a config option at some point whether we want to finetune the previous IL model. Let's create two issues for this so we don't forget about it :)

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
MaxiBoether: Reading this implementation, I wonder whether we can also allow sample then batch? I know the original paper does bts (without accumulation) but I don't see any reason why this would not work in a StB setting

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def get_irreducible_loss(
+        self, sample_ids: list[int], forward_input: Union[dict[str, torch.Tensor], torch.Tensor], target: torch.Tensor
+    ) -> torch.Tensor:
+        # use sample_ids to index into the precomputed loss values, if they are all in the cache
+        # or use forward_input to compute the loss values
+        fullly_in_cache = all(sample_id in self.loss_cache for sample_id in sample_ids)
+        if fullly_in_cache:
+            # torch.stack creates a new tensor
+            return torch.stack([self.loss_cache[sample_id] for sample_id in sample_ids]).to(self.device)
```
MaxiBoether, line 44: I don't fully get the caching logic yet, I think. My first question is why do we need a cache? Do we ever call `get_irreducbile_loss` multiple times? Because if the loss stays on the GPU, that may eat up VRAM.

My second question is why, if the loss is not fully in cache, we recompute everything, instead of just the data where we have not cached it yet. Anyways, I think I am more confused why we need the cache, if we onlycall get_irreducible_loss in the function inform_samples once.

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
```diff
@@ -0,0 +1,104 @@
...
+
```
MaxiBoether, line 45: third comment on the cache: do we ever clean this up? because it looks like over a whole training run, we would continually grow this, which might be not optimal.

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_rho_loss_downsampling.py
```diff
@@ -0,0 +1,109 @@
...
+@patch.object(IrreducibleLossProducer, "_load_il_model", return_value=dummy_model())
+def test_select_points(mock__load_il_model, dummy_init_params):
+    dummy_init_params["batch_size"] = 5
+    dummy_init_params["params_from_selector"]["downsampling_ratio"] = 60
+    sampler = RemoteRHOLossDownsampling(**dummy_init_params)
+    sampler.rho_loss = torch.tensor([32, 7, 5, 8, 3])
+    sampler.number_of_points_seen = 5
+    sampler.index_sampleid_map = [2, 4, 1, 3, 5]
+    selected_ids, weights = sampler.select_points()
+    assert selected_ids == [2, 4, 3]
```
MaxiBoether, line 108: should it not be [2,3,4] if it is sorted by rho loss? anyways i guess we shuffle, but i was just a little confused

MaxiBoether (changes requested):
Thank you Xianzhe! Most of my comments are minor. My biggest question is around the cache in the producer, but other than that all of this looks very clean and understandable, thanks so much for the work

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
XianzheMa: also confused

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_uncertainty_downsampling_strategy.py
MaxiBoether: oh ok. i guess in this particular case it also does not 100% matter. my thought process was that if we give input we check whether anything goes wrong (it won't) but then again it should also go wrong if None is supplied so we can also keep this

modyn/trainer_server/internal/dataset/key_sources/selector_key_source.py
XianzheMa: I don't understand it now. Why hasn't pylint complained about the other occurrence :( but I will make a consistent change

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
XianzheMa: Implementation-wise it is not a problem at all to have sample-then-batch. It is just that we then select the entire epoch for use, and keep updating the model using this epoch. So the samples selected to use in the second half of this epoch might not be up-to-date.

We previously agreed on this in doc https://docs.google.com/document/d/19dAm9IfVz4tK8-y8tTuo4lIiy5P7jC-juJveyJYB1AI/edit#heading=h.nrwv8u52nq3j.

But sure we can also have it and do experiments, what do you think?

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
XianzheMa: I think I will only move this `irreducible_loss_producer.py` file to the subdirectory, because for consistency we also have the `deepcore_utils` subdirectory which holds all the utilities for deepcore-based downsamplers, which reside out of the subdirectory.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
XianzheMa: Okay the conclusion after our discussion: we will keep it as-is for now. Later when we want to test sample-then-batch, there is no implementation block for us.

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
XianzheMa: Per discussion:
1. we need a cache because we (possibly) will train for multiple epochs, i.e. passing the data multiple times. We will call `get_irreducbile_loss` multiple times for the same sample. The cache does not stay always on GPU: it always stays on CPU and when it is needed, the needed part is moved to gpu.
2. The time to compute partial batch should be the same as compute the full batch. Also, we shouldn't have a partial-hit case (i.e. either fully miss or fully hit), because this transition should happen per-epoch: i.e. during first epoch every sample is missing in the cache, after the first epoch every sample is in the cache. We won't (in the most cases) have duplicate samples within an epoch.

modyn/trainer_server/internal/trainer/remote_downsamplers/irreducible_loss_producer.py
XianzheMa: the cache is cleaned up after we finishing the training, where the entire training process is destroyed

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_rho_loss_downsampling.py
XianzheMa: The current implementation fetches the top k but does not return the subset in the order sorted by rho loss, but in the oder consistent to the input order, i.e. here we have `[2, 4, 1, 3, 5]` as the input sample ids, so we keep the relative order. I think it's natural not to shuffle within the batch here? (As the only function is to select a subset).

But yes, it does not matter much I agree with you.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
XianzheMa: https://github.com/eth-easl/modyn/issues/504

MaxiBoether (approved):
Thanks, looks like everything has been addressed! After you fixed CI and everything runs through, feel free to merge :-)

MaxiBoether:
feel free to reduce the required coverage to 82%, if that's the blocking issue here

codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
```
XianzheMa, line 5: @robinholzi I will only make `codecov/project` action informational: i.e. it shows information but will not block PR. Otherwise currently we cannot proceed.

XianzheMa: Let's hope that I've configured correctly and this time all checks can pass

robinholzi: Makes sense, thanks!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rho_loss_downsampling.py
XianzheMa: Note we initialize `self.rho_loss` as `torch.tensor([])`. This empty tensor has dtype float32 (pytorch default dtype) irregardless of the loss type.

PR #493, opened 2024-06-11
feat: Setup huffpost multi-pipeline experiment

robinholzi:
# Motivation

Configuration for huffpost dataset pipelines using multiple trigger policies and evaluation handlers.


experiments/huffpost/compare_trigger_policies/run.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def construct_pipelines() -> list[ModynPipelineConfig]:
+    pipeline_configs: list[ModynPipelineConfig] = []
+    first_timestamp = datetime.datetime(2012, 1, 28).timestamp()
+    last_timestamp = datetime.datetime(2022, 9, 24).timestamp()
+
+    eval_handlers = [
+        EvalHandlerConfig(
+            name="scheduled-180d",
```
robinholzi, line 24: TODO: insert variable

experiments/huffpost/compare_trigger_policies/run.py
```diff
@@ -0,0 +1,88 @@
...
+            execution_time="after_pipeline",
+            models="matrix",
+            strategy=PeriodicEvalStrategyConfig(
+                every="1d",
```
robinholzi, line 28: TODO: fix

docker/Base/Dockerfile
```diff
@@ -17,3 +17,6 @@ ENV CONDA_DEFAULT_ENV modyn
 ENV MAMBA_DEFAULT_ENV modyn
 RUN /bin/bash -c "mamba init"
 RUN echo "mamba activate modyn" >> /home/appuser/.bashrc
+
+# set environment variable to tell modyn that it is running in a container
+ENV MODYN_DOCKER 1
```
robinholzi, line 22: Would you prefer adding directories like `experiments` into the container? I think it's fair to only run the tests that need those in non-docker mode (hence the environment variable)

MaxiBoether: Nope, I would like to keep experiments outside of the container, if possible. because otherwise we'll rebuild the container if we change an experiment which i execute from outside containers anyways :D

experiments/huffpost/compare_trigger_policies/pipeline_config.py
```diff
@@ -0,0 +1,99 @@
...
+        training=TrainingConfig(
+            gpus=1,
+            device="cuda:0",
+            dataloader_workers=8,
+            use_previous_model=True,
+            initial_model="random",
+            batch_size=64,
+            shuffle=True,
+            optimizers=[
+                OptimizerConfig(
```
MaxiBoether, line 39: just a general note: for some benchmarks, I am still exploring the baseline (basically arXiv right now, but CLOC is also open). however this should not affect what you are doing. you can for sure use a different baseline than we might do for the sigmod submisison and then update later.

MaxiBoether (approved):
Thanks, sorry for the late review. I think this mostly affects your experiments and they might differ from what we do for SIGMOD potentially, but that's totally fine, since its a bit fluid atm

PR #495, opened 2024-06-11
feat: Add original huffpost and arxiv datasets with full time resolution

robinholzi:
# Motivation

Currently we are sourcing arxiv and huffpost datasets from the wilds collection which is preprocessed and samples are binned into years. We want to have finer grained time resolution like in the original datasets. Also, we want to use submission dates rather than `arxiv`'s `update_date` which is rather some arxiv internal field rather than date that has semantic meaning to a paper submission. E.g. there are a lot of papers from < 2000 that have `2007 <= update_date <= 2008` (probably `arxiv` imported them back then)


benchmark/huffpost_kaggle/data_generation.py
```diff
@@ -0,0 +1,129 @@
...
+from tqdm import tqdm
+from modyn.const.types import TimeResolution
+from benchmark.utils.time_resolution_binning import (
+    bin_dataframe_wrt_time_resolution,
+    df_to_csv_with_timestamp,
+    HELP_RESOLUTION,
+)
+
+
+class HuffpostKaggleDownloader(Dataset):
```
robinholzi, line 19: I acknowledge that there is some degree of code duplication originating from the fact that `huffpost` and `arxiv` have similar data formats and are sourced from kaggle. I think that's justified though as the scripts should be independent. (Changing one shouldn't break anything in the other).

benchmark/arxiv_kaggle/data_generation.py
```diff
@@ -0,0 +1,142 @@
...
+    bin_dataframe_wrt_time_resolution,
+    df_to_csv_with_timestamp,
+)
+from modyn.const.types import TimeResolution
+from sklearn.model_selection import train_test_split
+from torch.utils.data import Dataset
+from tqdm import tqdm
+
+
+class ArxivKaggleDownloader(Dataset):
```
MaxiBoether, line 19: Not sure we need to inherit from dataset here. Dataset is a class to be put into a data loader for training. Reading the code, I don't think we need that, do we?

benchmark/arxiv_kaggle/data_generation.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def assert_data_exists(self, data_zip: Path) -> None:
+        assert data_zip.exists(), "Kaggle dataset zip file not found."
+
+    def extract_data(self, data_zip: Path) -> None:
+        shutil.unpack_archive(data_zip, self.data_dir)
+
+    @property
+    def data_json(self) -> Path:
+        return self.data_dir / "arxiv-metadata-oai-snapshot.json"
```
MaxiBoether, line 35: Do you assume this is downloaded by the user? It's not  clear to me where this is coming from. Either we need a README similar to CGLM, or implement some download logic

benchmark/huffpost_kaggle/data_generation.py
MaxiBoether: yeah that's totally fine here

benchmark/huffpost_kaggle/data_generation.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def assert_data_exists(self, data_zip: Path) -> None:
+        assert data_zip.exists(), "Kaggle dataset zip file not found."
+
+    def extract_data(self, data_zip: Path) -> None:
+        shutil.unpack_archive(data_zip, self.data_dir)
+
+    @property
+    def data_json(self) -> Path:
+        return self.data_dir / "News_Category_Dataset_v3.json"
```
MaxiBoether, line 35: similar comment as for arxiv

benchmark/arxiv_kaggle/data_generation.py
MaxiBoether: i am a bit confused since the class (same for huffpost) is named "Downloader" so I was trying to find the download logic somewhere. Not sure it actually exists

scripts/python_compliance.sh
```diff
@@ -46,6 +46,20 @@ else
...
 fi
 
+if $MAMBA_CMD run -n modyn ruff check experiments benchmark/arxiv_kaggle benchmark/huffpost_kaggle ; then
+    echo "No ruff check errors"
+else
+    echo "ruf check errors"
+    exit 1
+fi
+
+if $MAMBA_CMD run -n modyn ruff format experiments benchmark/arxiv_kaggle benchmark/huffpost_kaggle ; then
```
MaxiBoether, line 56: do we need to supply the ruff.toml file or does it get that automatically?

MaxiBoether (changes requested):
Thank you! I like the general binning logic, but my major comment is that we either need a README that explains necessary steps (where to download the data from) similar to CGLM or to implement the download in code (similar to wild time arxiv). If we don't actually download ourselves, we might want to rename the classes :D After that, I think we're good to merge

benchmark/arxiv_kaggle/data_generation.py
robinholzi: No, that's simply an artifact of copying the baseline of another `data_generation.py` script.

benchmark/arxiv_kaggle/data_generation.py
robinholzi: The only reliable way that I found for downloading from kaggle is via the [kaggle API](https://github.com/Kaggle/kaggle-api). It also requires user auth, so I just went for the approach that a user of the scripts downloads the dataset file from kaggle via the browser and the just hands the filepath to the script. The `Download` in the classname is, indeed, a bit off.

scripts/python_compliance.sh
robinholzi: It should be able to pick it up automatically. Unless the script is executed from somewhere outside of the repo. Vut I guess we would notice if that happens.

robinholzi:
> Thank you! I like the general binning logic, but my major comment is that we either need a README that explains necessary steps (where to download the data from) similar to CGLM or to implement the download in code (similar to wild time arxiv). If we don't actually download ourselves, we might want to rename the classes :D After that, I think we're good to merge

Makes sense! I will add a small README :)

MaxiBoether (approved):
Thank you, this looks nice!

robinholzi:
Thanks for the review @MaxiBoether :)

PR #496, opened 2024-06-12
fix: Get `next_trigger_Id` from database

MaxiBoether:
Because there are multiple instances per selector, we need to sync via the db.

### DIFF
--- a/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
+++ b/modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
@@ -1,4 +1,6 @@
 import logging
+import os
+import threading
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Iterable, Optional
 
@@ -46,18 +48,7 @@ class AbstractSelectionStrategy(ABC):
 
         logger.info(f"Initializing selection strategy for pipeline {pipeline_id}.")
 
-        with MetadataDatabaseConnection(self._modyn_config) as database:
-            last_trigger_id = (
-                database.session.query(func.max(Trigger.trigger_id))  # pylint: disable=not-callable
-                .filter(Trigger.pipeline_id == self._pipeline_id)
-                .scalar()
-            )
-            if last_trigger_id is None:
-                logger.info(f"Did not find previous trigger id DB for pipeline {pipeline_id}, next trigger is 0.")
-                self._next_trigger_id = 0
-            else:
-                logger.info(f"Last trigger in DB for pipeline {pipeline_id} was {last_trigger_id}.")
-                self._next_trigger_id = last_trigger_id + 1
+        self._update_next_trigger_id()
 
         if self.has_limit and self.training_set_size_limit > self._maximum_keys_in_memory:
             # TODO(#179) Otherwise, we need to somehow sample over multiple not-in-memory partitions, which is a problem
@@ -72,6 +63,28 @@ class AbstractSelectionStrategy(ABC):
 
         self._trigger_sample_directory = self._modyn_config["selector"]["trigger_sample_directory"]
         self._storage_backend = self._init_storage_backend()
+
+    def _update_next_trigger_id(self) -> None:
+        tid = threading.get_native_id()
+        pid = os.getpid()
+
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            last_trigger_id = (
+                database.session.query(func.max(Trigger.trigger_id))  # pylint: disable=not-callable
+                .filter(Trigger.pipeline_id == self._pipeline_id)
+                .scalar()
+            )
+            if last_trigger_id is None:
+                self._next_trigger_id = 0
+                logger.info(
+                    f"[{pid}][{tid}] Did not find previous trigger id DB for pipeline {self._pipeline_id},"
+                    + " next trigger is 0."
+                )
+            else:
+                self._next_trigger_id = last_trigger_id + 1
+                logger.info(
+                    f"[{pid}][{tid}] Updating next trigger for pipeline {self._pipeline_id} to {self._next_trigger_id}."
+                )
 
     @abstractmethod
     def _init_storage_backend(self) -> AbstractStorageBackend:
@@ -266,6 +279,8 @@ class AbstractSelectionStrategy(ABC):
         Returns:
             tuple[int, int, int]: Trigger ID, how many keys are in the trigger, number of overall partitions
         """
+        self._update_next_trigger_id()
+
         total_keys_in_trigger, num_partitions, log = self.store_training_set(
             self._pipeline_id,
             self._next_trigger_id,
--- a/modyn/selector/internal/grpc/selector_grpc_servicer.py
+++ b/modyn/selector/internal/grpc/selector_grpc_servicer.py
@@ -78,15 +78,19 @@ class SelectorGRPCServicer(SelectorServicer):
 
     def inform_data(self, request: DataInformRequest, context: grpc.ServicerContext) -> DataInformResponse:
         pipeline_id, keys, timestamps, labels = request.pipeline_id, request.keys, request.timestamps, request.labels
-        logger.info(f"[Pipeline {pipeline_id}]: Selector is informed of {len(keys)} new data points")
+        tid = threading.get_native_id()
+        pid = os.getpid()
+        logger.info(f"[{pid}][{tid}][Pipeline {pipeline_id}]: Selector is informed of {len(keys)} new data points")
 
         log = self.selector_manager.inform_data(pipeline_id, keys, timestamps, labels)
         return DataInformResponse(log=JsonString(value=json.dumps(log)))
 
     def inform_data_and_trigger(self, request: DataInformRequest, context: grpc.ServicerContext) -> TriggerResponse:
         pipeline_id, keys, timestamps, labels = request.pipeline_id, request.keys, request.timestamps, request.labels
+        tid = threading.get_native_id()
+        pid = os.getpid()
         logger.info(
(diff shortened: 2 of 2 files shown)


XianzheMa:
Even for the same pipeline (a single pipeline id), do we also have multiple selector instances for it?

MaxiBoether:
> Even for the same pipeline (a single pipeline id), do we also have multiple selector instances for it?

Yes. The reason is that gRPC runs multiple serving processes which - due to Python - don't share memory. We have a process pool, so to say. Hence, multiple processes might each serve the same pipeline. That's the reason why `_populate_pipeline_if_exists` exists

XianzheMa:
Good catch! The `DownsamplingScheduler` also needs to be fixes somehow:

`DownsamplingScheduler` sits inside one selection strategy, since the `_next_trigger_id` might not be in sync, the `DownsamplingScheduler.downsampler_index` might also not be in sync: one instance already advances the scheduler by 10, but another instance is just initialized with the first downsampler.

I couldn't come up with a quick fix, but since we seldom use the `DownsamplingScheduler` to apply multiple downsamplers, we can set this aside for now.

XianzheMa:
> > Even for the same pipeline (a single pipeline id), do we also have multiple selector instances for it?
> 
> Yes. The reason is that gRPC runs multiple serving processes which - due to Python - don't share memory. We have a process pool, so to say. Hence, multiple processes might each serve the same pipeline. That's the reason why `_populate_pipeline_if_exists` exists

I see thanks! But within one process, we should only have one selector per pipeline id right? I see this https://github.com/eth-easl/modyn/blob/607dffffa673b42101ec0977efbc9aef0f172fb4/modyn/selector/internal/selector_manager.py#L26

MaxiBoether:
s

> Good catch! The `DownsamplingScheduler` also needs to be fixes somehow:
> 
> `DownsamplingScheduler` sits inside one selection strategy, since the `_next_trigger_id` might not be in sync, the `DownsamplingScheduler.downsampler_index` might also not be in sync: one instance already advances the downsampler by 10, but another instance is just initialized with the first downsampler.
> 
> I couldn't come up with a quick fix, but since we seldom use the `DownsamplingScheduler` to apply multiple downsamplers, we can set this aside for now.

Oh, really good catch! Can you please open an issue about this? I think a bigger problem might be that the warmup period logic is also off then. We need to fix this :/

MaxiBoether:
> > > Even for the same pipeline (a single pipeline id), do we also have multiple selector instances for it?
> > 
> > 
> > Yes. The reason is that gRPC runs multiple serving processes which - due to Python - don't share memory. We have a process pool, so to say. Hence, multiple processes might each serve the same pipeline. That's the reason why `_populate_pipeline_if_exists` exists
> 
> I see thanks! But within one process, we should only have one selector per pipeline id right? I see this
> 
> https://github.com/eth-easl/modyn/blob/607dffffa673b42101ec0977efbc9aef0f172fb4/modyn/selector/internal/selector_manager.py#L26

yes! the problem is if request a goes to process 1, and request b goes to process 2.

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -266,6 +278,8 @@ def trigger(self) -> tuple[int, int, int, dict[str, Any]]:
         Returns:
             tuple[int, int, int]: Trigger ID, how many keys are in the trigger, number of overall partitions
         """
+        self._update_next_trigger_id()
+
```
XianzheMa, line 282: An sync at this place might not be early enough, as we overwrite `trigger()` in subclasses
https://github.com/eth-easl/modyn/blob/607dffffa673b42101ec0977efbc9aef0f172fb4/modyn/selector/internal/selector_strategies/coreset_strategy.py#L67

MaxiBoether: rip. ok i need to rethink how we do this then

MaxiBoether: tried to fix it at the selector side. actually we also need to update it before `inform_data`

XianzheMa: this should work!

XianzheMa:
> s
> 
> > Good catch! The `DownsamplingScheduler` also needs to be fixes somehow:
> > `DownsamplingScheduler` sits inside one selection strategy, since the `_next_trigger_id` might not be in sync, the `DownsamplingScheduler.downsampler_index` might also not be in sync: one instance already advances the downsampler by 10, but another instance is just initialized with the first downsampler.
> > I couldn't come up with a quick fix, but since we seldom use the `DownsamplingScheduler` to apply multiple downsamplers, we can set this aside for now.
> 
> Oh, really good catch! Can you please open an issue about this? I think a bigger problem might be that the warmup period logic is also off then. We need to fix this :/

will do! Now with the update on selector side, I think warmup should be fine as long as we don't use multiple downsamplers :D

XianzheMa (approved):
LGTM

MaxiBoether:
Thanks for approving, I will wait for the integration tests to run through to make sure this did not affect anything

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
```diff
@@ -73,6 +64,27 @@ def __init__(self, config: SelectionStrategy, modyn_config: dict, pipeline_id: i
         self._trigger_sample_directory = self._modyn_config["selector"]["trigger_sample_directory"]
         self._storage_backend = self._init_storage_backend()
 
+    def _update_next_trigger_id(self) -> None:
+        tid = threading.get_native_id()
+        pid = os.getpid()
+
+        with MetadataDatabaseConnection(self._modyn_config) as database:
+            last_trigger_id = (
```
robinholzi, line 72: I propose to use the atomic transactional capabilities of our database to force different processes to use unique `trigger_id`s. With the current version (`select max query`) we could still run into the problem that two parallel processes request a new trigger id at the same time before writing the trigger back to the db. Then both processes would end up with the same trigger id. 
Doing the `self._next_trigger_id = last_trigger_id + 1` logic  in the database is more reliable I think. We could either use an autoincrement table primary key or sth. more lightweight like a postgres series.

robinholzi (changes requested):

modyn/selector/internal/selector_strategies/abstract_selection_strategy.py
MaxiBoether: generally, i agree. we use a mp.manager.Lock() to prevent what you are saying. However of course it's better to use the db here. There is some historic reason we do not rely on the DB. IIRC we somehow needed the actual ID of the next trigger before we insert and at some point, we did not have a composite primary key of (pid, tid) but just tid(), so there was no obvious +1 logic. We should refactor this, but my proposal would be to merge this anyways as a hotfix to have a somewhat functional main branch (the locks at the grpc servicer level should hopefully make it correct)

robinholzi: Sounds good :) It's already an improvement from the status quo!

robinholzi (approved):

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -125,11 +126,25 @@ def _init_transforms(self) -> None:
...
-    def _init_grpc(self) -> None:
-        self._storage_channel = grpc.insecure_channel(self._storage_address, options=grpc_common_config())
-        if not grpc_connection_established(self._storage_channel):
-            raise ConnectionError(f"Could not establish gRPC connection to storage at address {self._storage_address}.")
-        self._storagestub = StorageStub(self._storage_channel)
+    def _init_grpc(self, worker_id: Optional[int]) -> None:
+        max_retries = 5
+        retry_delay = 1  # seconds
+
+        for attempt in range(1, max_retries + 1):
```
robinholzi, line 133: It's okay this time, generally, this could be achieved through dedicated libraries like [tenacity](https://tenacity.readthedocs.io/en/latest/) that don't obfuscate the business logic that much :)

MaxiBoether: Thanks for yet another nice pointer into the Python ecosystem :D Yeah we have such logic in many places, I guess we should refactor this consistently...

integrationtests/online_dataset/test_online_dataset.py
```diff
@@ -438,7 +441,11 @@ def test_dataset() -> None:
 def main() -> None:
     try:
         test_dataset()
+    except Exception as e:
+        print(f"An exception occured: {e}")
+        raise
```
XianzheMa, line 446: @MaxiBoether just want to ensure it's intended: why do we want to re-raise the exception here? Or is this a leftover from debugging?
If we do intend to re-raise exception, we can simply do
```python
test_dataset()
cleanup_storage_database()
cleanup_dataset_dir()
```
, or if we need the printed message, we can do 
```python
try:
    test_dataset()
    print(...)
else:
    print(...)
    cleanup_storage_database()
    cleanup_dataset_dir()
```

But anyway, we should clean up things no matter the tests go wrong/successful right?

MaxiBoether: My idea here was:

If there is any exception, I want to log and then fail such that we have an exit code -1. in both cases, I want to cleanup at the end. The reason is that i had the feeling at some point the test was exciting with -1 without printing any exception stacktrace, this gives me a custom string to search for in the logs. I dont fully understand yet whether the way i did this is problematic or not

XianzheMa: Sorry I think I misunderstood how `try` block works. It's not problematic at all!

PR #499, opened 2024-06-12
feat: Setup `arxiv` multi-pipeline experiment

robinholzi:
# Motivation

Configuration for arxiv dataset pipelines using multiple trigger policies and evaluation handlers.


MaxiBoether (approved):
I don't have a problem with merging this, although I run similar but different stuff for SIGMOD right now, which I will also merge at some point :)

robinholzi:
> I don't have a problem with merging this, although I run similar but different stuff for SIGMOD right now, which I will also merge at some point :)

Maybe we can merge the shared parts of the configurations at some point to reduce the duplication.

PR #500, opened 2024-06-12
RHO-LOSS integration test

XianzheMa:
should be reviewed after https://github.com/eth-easl/modyn/pull/491


integrationtests/config/rho_loss.yaml
```diff
@@ -0,0 +1,73 @@
+pipeline:
+  name: RHO-Loss Test Pipeline
+  description: Example pipeline
+  version: 1.0.0
+model:
+  id: ResNet50
```
MaxiBoether, line 6: i have not taken a look at this PR, but we cannot train a ResNet50 on a github runner :D if at all, we should use a very very small dummy model, just some linear layer or something, for the integration test

XianzheMa: thanks for spotting this. I changed it to a smaller one

PR #501, opened 2024-06-12
refactor: Add pydantic interface for metrics

robinholzi:
# Motivation

Small quality of live improvement that makes it easier to configure evaluation metrics (as discussed [here](https://github.com/eth-easl/modyn/pull/467#discussion_r1627254655))


MaxiBoether:
is that ready for review?

robinholzi:
Ah yes, I think I was just waiting for the tests. The failing integration tests seems unrelated.

integrationtests/utils.py
```diff
@@ -41,7 +41,7 @@
 MNIST_CONFIG_FILE = MODYNCLIENT_CONFIG_PATH / "mnist.yaml"
 DUMMY_CONFIG_FILE = MODYNCLIENT_CONFIG_PATH / "dummy.yaml"
 CLIENT_ENTRYPOINT = SCRIPT_PATH.parent.parent / "modynclient" / "client" / "modyn-client"
-NEW_DATASET_TIMEOUT = 600
+NEW_DATASET_TIMEOUT = 30
```
MaxiBoether, line 44: Any reason for this change? I guess 600 is indeed a bit high...

modyn/config/schema/pipeline/evaluation/metrics.py
```diff
@@ -0,0 +1,82 @@
...
+        return True
+
+
+class AccuracyMetricConfig(_BaseMetricConfig):
+    name: Literal["Accuracy"] = Field("Accuracy")
+    topn: int = Field(1, description="The top-n accuracy to be computed.", ge=1)
+
+    @property
+    def shape_check(self) -> bool:
+        return self.topn is None
```
MaxiBoether, line 45: Shouldn't it be > 1? I did not check this, but since it defaults to 1, can it ever be None? Or is this done before defaults are applied?

modyn/evaluator/internal/metric_factory.py
```diff
@@ -1,19 +1,17 @@
...
 all_metrics: set[type[AbstractEvaluationMetric]] = {Accuracy, F1Score, RocAuc}
 
 
 class MetricFactory:
     @staticmethod
-    def get_evaluation_metric(
-        name: str, evaluation_transform_func: str, config: dict[str, Any]
-    ) -> AbstractEvaluationMetric:
+    def get_evaluation_metric(config_json: str) -> AbstractEvaluationMetric:
+        parsed_metric_config = validate_metric_config_json(str(config_json))
```
MaxiBoether, line 10: why str()? config_json already is a string

MaxiBoether (approved):
Thanks! I just had some small comments. I am a bit torn on what is cleaner, having this in pydantic or having a more explicit object in the proto file, but i suppose since pydantic is user-facing and provides validation, I rather just dump a json string into grpc instead of not having pydantic. I like the solution, thanks for coming up with this!

integrationtests/utils.py
robinholzi: During debugging, I was annoyed that the tests always take so long even without any particular reason to wait 600s. Even 10s was fine for me.

modyn/evaluator/internal/metric_factory.py
robinholzi: I remember there was some issue with the integration tests where the passed value was a JsonString and not a str. I can check again.

MaxiBoether: I guess then we should rather adjust the integration test :D

robinholzi: Yeah, I was a bit lazy with debugging😅

modyn/config/schema/pipeline/evaluation/metrics.py
MaxiBoether: I am not fully sure I get your comment yet. The logic you posted is indeed the old logic. However, in the pydantic implementation, self.topn is never None, is it? Because self is referring to the accuracy config object, where topn defaults to 1. Hence, right now shape_check is always False, but we want it to be true if topn == 1.

robinholzi: Ah, i see, then I misinterpreted the old logic. If one had set `topn=1` manually then `self. _enable_shape_check` would have been False. Let's use the semantics that you just described!

PR #503, opened 2024-06-13
fix: Uncertainty based downsampler choosing the most certain samples

MaxiBoether:
We had our own version of https://github.com/PatrickZH/DeepCore/issues/11 because our version of their implementation confused where the inversion is placed. I thought it through and think we don't need to do any inversion. I added some comments explaining the thoughts.

Note that this does not address https://github.com/PatrickZH/DeepCore/issues/13!

### DIFF
--- a/modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
+++ b/modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
@@ -17,6 +17,7 @@ class RemoteUncertaintyDownsamplingStrategy(AbstractPerLabelRemoteDownsamplingSt
     This strategy collects a measure of uncertainty (LeastConfidence, Entropy or Margin) for each sample and selects
     the top-k most uncertain samples.
     The user can specify which metric to use with the pipeline parameter score_metric.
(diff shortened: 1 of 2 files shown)


modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
```diff
@@ -64,19 +65,26 @@ def inform_samples(
         # keep the mapping index<->sample_id
         self.index_sampleid_map += sample_ids
 
-    def _compute_score(self, forward_output: torch.Tensor) -> np.ndarray:
+    def _compute_score(self, forward_output: torch.Tensor, disable_softmax: bool = False) -> np.ndarray:
```
XianzheMa, line 68: why do we add this option?

XianzheMa (approved):
LGTM!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
MaxiBoether: For the unit test. Because if there is a softmax applied i have no way of easily testing whats going on. If we just take the model output i can calculate this straightforwardly and check correctness.

PR #506, opened 2024-06-13
fix: Restrict pytest coverage directories

robinholzi:


MaxiBoether (approved):

.coveragerc
```diff
@@ -1,4 +1,6 @@
 [run]
+source=**/modyn/*
 omit = 
     **/*/__init__.py
-    **/generated/*
\ No newline at end of file
+    **/generated/*
+    **/modyn/tests/*
```
XianzheMa, line 6: just a question why isn't this `modyn/tests/*`?

PR #509, opened 2024-06-13
`RHO LOSS` integration test

XianzheMa:
Per the title.


benchmark/mnist/mnist.yaml
```diff
@@ -48,3 +48,40 @@ selection_strategy:
   storage_backend: "database"
   limit: -1
   tail_triggers: 0
+evaluation:
```
XianzheMa, line 51: there are two mnist.yaml files. One here and one in `modynclient` directory.

The one in `modynclient` is not used anywhere in code. So I removed that one but add the evaluation section to this one.

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -46,32 +54,9 @@ def assert_pipeline_exit_without_error(res: dict) -> None:
     assert exit_msg["exit_msg"]["exitcode"] == 0
 
 
-def test_mnist() -> None:
-    pipeline_config = load_config_from_file(MNIST_CONFIG_FILE)
```
XianzheMa, line 50: this func is not used anywhere

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -46,32 +54,9 @@ def assert_pipeline_exit_without_error(res: dict) -> None:
...
-    get_pipeline_status_res = parse_grpc_res(wait_for_pipeline(pipeline_id))
-    print(get_pipeline_status_res)
-    assert_pipeline_exit_without_error(get_pipeline_status_res)
-
-
 def test_one_experiment_pipeline() -> None:
     pipeline_config = load_config_from_file(DUMMY_CONFIG_FILE)
-
+    supervisor = SupervisorStub(connect_to_server("supervisor"))
     res = parse_grpc_res(
```
XianzheMa, line 60: previously `supervisor` variable is used implicitly. It is initialized in `if __name__ == '__main__':` block

integrationtests/utils.py
```diff
@@ -34,12 +34,15 @@
...
+    os.getenv("MODYN_INTEGRATIONTESTS_CONFIG_PATH", SCRIPT_PATH.parent / "config")
+)
 MODYN_DATASET_PATH = pathlib.Path(os.getenv("MODYN_DATASET_PATH", pathlib.Path("/app") / "storage" / "datasets"))
 MODYN_MODELS_PATH = pathlib.Path(os.getenv("MODYN_MODELS_PATH", pathlib.Path("/app") / "model_storage"))
 
 CLIENT_CONFIG_FILE = MODYNCLIENT_CONFIG_PATH / "modyn_client_config_container.yaml"
-MNIST_CONFIG_FILE = MODYNCLIENT_CONFIG_PATH / "mnist.yaml"
-DUMMY_CONFIG_FILE = MODYNCLIENT_CONFIG_PATH / "dummy.yaml"
+DUMMY_CONFIG_FILE = MODYN_INTEGRATIONTESTS_CONFIG_PATH / "dummy.yaml"
+RHO_LOSS_CONFIG_FILE = MODYN_INTEGRATIONTESTS_CONFIG_PATH / "rho_loss.yaml"
```
XianzheMa, line 45: It's a bit unlogical to have pipeline config files used in integration tests in the `modynclient` directory. I moved them to inside `integrationtests` directory.

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -5,23 +5,31 @@
...
 POLL_INTERVAL = 1
+MODYN_CONFIG = get_modyn_config()
+IMAGE_DATASET_ID = "image_test_dataset"
 
 
-def wait_for_pipeline(pipeline_id: int) -> str:
+def wait_for_pipeline(pipeline_id: int) -> GetPipelineStatusResponse:
     req = GetPipelineStatusRequest(pipeline_id=pipeline_id)
-
+    supervisor = SupervisorStub(connect_to_server("supervisor"))
```
XianzheMa, line 32: previously supervisor variable is used implicitly. It is initialized in if __name__ == '__main__': block

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -46,54 +54,9 @@ def assert_pipeline_exit_without_error(res: dict) -> None:
...
-    print(f"start pipeline: {res}")
-    assert pipeline_id >= 0
-
-    get_pipeline_status_res = parse_grpc_res(wait_for_pipeline(pipeline_id))
-    print(get_pipeline_status_res)
-    assert_pipeline_exit_without_error(get_pipeline_status_res)
-
-
-def test_one_experiment_pipeline() -> None:
-    pipeline_config = load_config_from_file(DUMMY_CONFIG_FILE)
```
XianzheMa, line 73: we have a func `test_two_experiment_pipelines` which covers entirely what this function covers.

So I remove this func to marginally save some integration test time.

modynclient/tests/client/test_client_entrypoint.py
```diff
@@ -35,7 +35,7 @@ def noop_poll_pipeline_status(self) -> None:
 @patch.object(Client, "__init__", noop_constructor_mock)
 @patch.object(Client, "start_pipeline", return_value=True)
 @patch.object(Client, "poll_pipeline_status", noop_poll_pipeline_status)
-def test_modyn_client_script_runs(script_runner):
+def test_modyn_client_script_runs(mock_start_pipeline, script_runner):
     ret = script_runner.run("_modyn_client", str(EXAMPLE_PIPELINE), str(EXAMPLE_SYSTEM_CONFIG))
     assert ret.success
```
XianzheMa, line 40: The previous test can succeed no matter what is configured... Because the first arg is a mock (and is mistook as the script runner).

The mock without any spec has arbitrary methods with arbitrary interfaces. And the returned value is still a mock evaluated to True under assertion.

integrationtests/config/dummy.yaml
MaxiBoether, line 1: Unfortunately, we cannot move these dummy config files. They are supposed to serve as a guideline for new users and show configs e.g. for the different benchmark datasets. I also use them for running my own stuff. I think it's actually good that our tests verify that our dummy configs work. Hence, let's revert this change. Thanks!

integrationtests/config/rho_loss.yaml
```diff
@@ -0,0 +1,74 @@
...
+  storage_backend: "database"
+  limit: -1
+  tail_triggers: 0
+  downsampling_config:
+    strategy: RHOLoss
+    sample_then_batch: False
+    holdout_set_ratio: 30
+    ratio: 60
+    il_training_config:
+      il_model_id: ResNet18
```
MaxiBoether, line 57: The github runners are super duper slow and i dont want to overload them. In modyn/models there should be a dummy model with just a linear layer. Can we use this here for both the IL and regular model? That minimizes the used computation

MaxiBoether (changes requested):
Thank you! The test looks good but I would like to use a smaller model and not move the config files :) thank you!

integrationtests/config/rho_loss.yaml
XianzheMa: Make sense I will replace it with Dummy!

integrationtests/config/dummy.yaml
XianzheMa: I don't have a strong opinion on keeping it at the old place, but just want to mention:
We still have a `mnist.yaml` in `modynclient/config/examples`, isn't that enough for showing to users how it works?

For more diverse pipeline configs we have the `benchmark` directory which holds all kinds of more pipeline configs.

For your use case, does it strictly depend on a path under `modynclient`? (I guess the new path under `integrationtests`) should also work.

Anyway if you still think it should be placed back I will do! I just didn't understand the points.

integrationtests/config/rho_loss.yaml
XianzheMa: You are right after I replaced ResNet18 with Dummy the test becomes snappy fast!

This comment was addressed.

integrationtests/config/dummy.yaml
MaxiBoether: Sorry, I did not realize this affects the modyn_client_ config. For some reason I thought we adjust the modyn system config. Nevermind then

MaxiBoether (approved):
Thanks for adjusting the model!

Issue #504, opened 2024-06-13
Have more versions of `RHO-LOSS`


XianzheMa:
1. We can enable `sample-then-batch` for `RHO-LOSS`.
2. When we select points, we can select depending on the RHO Loss as weight, rather than just select the top k.
3. We can finetune the previous IL model instead of training a randomly initialized one

Proposed from https://github.com/eth-easl/modyn/pull/491#discussion_r1637887297

PR #510, opened 2024-06-14
More `RHO LOSS` flavors

XianzheMa:
1. Make it possible to train the il model from last trigger's il model.
2. Enable `sample-then-batch` for `RHO Loss`.

Address https://github.com/eth-easl/modyn/issues/504

### DIFF
--- a/modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
+++ b/modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Iterable, Tuple
+from typing import Any, Iterable, Optional, Tuple
 
 from modyn.common.grpc.grpc_helpers import TrainerServerGRPCHandlerMixin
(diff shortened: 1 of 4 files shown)


modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -51,34 +51,43 @@ def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: Ab
...
                 .scalar()
             )
             assert il_model_id is not None
-
-        config["il_model_id"] = il_model_id
-        return config
+        return il_model_id
 
     def _train_il_model(self, trigger_id: int) -> int:
+        if self.il_training_config.use_previous_model:
```
MaxiBoether, line 81: Nit but we can put this into a single statement like a = b if c else d

MaxiBoether (approved):
Thanks, this looks good!

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: yeah that's my first choice but I feel I will have a really long single line so I switched to this old-school style :D

Issue #508, opened 2024-06-13
Move `after_pipeline|training_evaluation_workers` to pipeline config


MaxiBoether:
Instead of configuring this on a modyn config level, it might be helpful to have this on a pipeline level. For example, on the same Modyn instance, right now I have yearbook and CGLM running. I can evaluate a lot more yearbook models in parallel than CGLM models. Hence it would be great to have this on a pipeline level instead.

MaxiBoether:
Thanks for fixing that!

PR #511, opened 2024-06-14
Move `after_pipeline|training_evaluation_workers` to pipeline config

XianzheMa:
Per the title, this solves https://github.com/eth-easl/modyn/issues/508.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,4 +4,5 @@ coverage:
       default:
         informational: true
     patch:
-      informational: true
+      default:
+        informational: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,3 +3,5 @@ coverage:
     project:
       default:
         informational: true
+    patch:
+      informational: true
--- a/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
+++ b/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
@@ -156,7 +156,9 @@ class EvaluationExecutor:
             )
             eval_requests += handler_eval_requests
 
-        num_workers = self.config.supervisor.after_training_evaluation_workers if self.config.supervisor else 1
+        if len(eval_requests) == 0:
+            return SupervisorLogs()
+        num_workers = self.pipeline.evaluation.after_training_evaluation_workers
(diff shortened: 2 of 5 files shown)


modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -156,7 +156,9 @@ def run_pipeline_evaluations(
             )
             eval_requests += handler_eval_requests
 
-        num_workers = self.config.supervisor.after_training_evaluation_workers if self.config.supervisor else 1
+        if len(eval_requests) == 0:
+            return SupervisorLogs()
+        num_workers = self.pipeline.evaluation.after_training_evaluation_workers
```
XianzheMa, line 161: here we don't need to assert on self.pipeline.evaluation, as it is asserted in the beginning of this func.

MaxiBoether (approved):
Thank you!

codecov.yml
```diff
@@ -3,3 +3,5 @@ coverage:
     project:
       default:
         informational: true
+    patch:
+      informational: true
```
XianzheMa, line 7: We somehow don't have a unit test file for the quite large `evaluation_executor.py`.  https://github.com/eth-easl/modyn/pull/490

@robinholzi Is there a reason? If it is no intentional please add one!

I will for the moment also make `patch` action informational. As it does not make sense for me to create a huge unit test only to cover these trivial changes.

XianzheMa:
I need to resolve the comment to merge the PR, so I copy-paste my question here:
We somehow don't have a unit test file for the quite large evaluation_executor.py. https://github.com/eth-easl/modyn/pull/490

@robinholzi  Is there a reason? If the missing unit test is not intentional please add one!

I will for the moment also make patch action informational. As it does not make sense for me to create a huge unit test only to cover these trivial changes.

robinholzi:
@XianzheMa I guess that depends on what you define as a `unit`. We are testing the logic in `evaluation_executor` with the logic from `test_pipeline_supervisor` just like we did before the introduction of `evaluation_executor`. As there are some uncovered parts though, I will add dedicated tests for them! Thanks for the pointer!

PR #512, opened 2024-06-15
Fix unsynced warmup across selector instances

XianzheMa:
This PR acts more as a patch rather than a fundamental fix to the asynchronism across different selector instances.

A fundamental fix should make the selection strategy *stateless*: by getting rid of the field`self._next_trigger_id` and providing it in the upstream as an argument.

This PR reduces the statefullness by removing the `self.is_warmup`.


MaxiBoether (approved):
Thanks, this LGTM!

PR #513, opened 2024-06-16
test: Add unit tests for `evaluation_executor`

robinholzi:
# Motivation

Let's increase test coverage for `EvaluationExecutor`. Testing this component is a bit nasty as it uses many other components, but okay.


MaxiBoether (approved):
Thank you Robin for increasing the covergae here!

PR #515, opened 2024-06-17
Fix inconsistency on passed number of samples in `batch-then-sample` mode

XianzheMa:


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -278,6 +278,7 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
                         if not batch_accumulator.inform_batch(data, sample_ids, target, weights):
                             stopw.start("FetchBatch", resume=True)
                             stopw.start("IndivFetchBatch", overwrite=True)
+                            self._num_samples += self._batch_size
```
XianzheMa, line 281: When we accumulate batches in `batch-then-sample`, we forget to increment `self._num_samples`.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -628,9 +629,6 @@ def send_model_state_to_server(self) -> None:
         bytes_state = buffer.read()
         self._status_response_queue_training.put(bytes_state)
 
-    def send_status_to_server_training(self, batch_number: int) -> None:
-        self._status_response_queue_training.put({"num_batches": batch_number, "num_samples": self._num_samples})
-
```
XianzheMa, line 633: not used anywhere

MaxiBoether:
Thank you so much for finding this!!

modyn/trainer_server/internal/trainer/pytorch_trainer.py
MaxiBoether: I actually do not think this logic is correct. Num samples tracks how many samples we trained on, not how many samples in total we passed. In StB, it would be incremented just as often as the downsampled set is big. In BtS, we with this change do it as often as we have pre downsampling items. The intention is to track the number of items we actually used for model updates. Hence I dont think they change is correct, but at least not consistent.

MaxiBoether (changes requested):

modyn/trainer_server/internal/trainer/pytorch_trainer.py
robinholzi: Oh, good that we encountered this! I always assumed `num_samples` would be the number of samples **considered** for training in the sense of training interval size. I think, it makes sense to report the number of samples in the training interval, the number of samples we trained on, and potentially also the number of samples we passed. I can be valuable to have all of these metrics available in the supervisor.

XianzheMa: @MaxiBoether I understand your comment. But we *did* use this number `self._num_samples` previously to track how many samples in total we passed. There are two points to prove this:

1. Before having this batch accumulation logic, we incremented `self._num_samples` once per batch, by the pre-downsampled batch size. https://github.com/eth-easl/modyn/blob/988bd013db897e7f15c40865539e732f62c33d84/modyn/trainer_server/internal/trainer/pytorch_trainer.py#L326
2. In client log tracker, we compare "the number of samples passed" with the "number of samples per epoch" to determine the current epoch number. https://github.com/eth-easl/modyn/blob/988bd013db897e7f15c40865539e732f62c33d84/modynclient/client/internal/utils/training_status_tracker.py#L103 I confirmed that the first number stems from the `self._num_samples`, and the second number stems from the data size after (possible) presampling at selector. https://github.com/eth-easl/modyn/blob/988bd013db897e7f15c40865539e732f62c33d84/modyn/supervisor/internal/pipeline_executor/pipeline_executor.py#L661

MaxiBoether: This could for sure explain the epoch number inconsistencies.  I think we should synchronize this somehow. Should we

1) track `_ num_samples_seen` and `_num_samples_trained` in the logs

and

2) report `_num_samples_trained` to the client such that the user is not confused why different sampling strategies return the same number of samples in the progress bar? I think that would be the cleanest fix here. Or do you disagree?

XianzheMa: I get your idea, but if we only report `_num_samples_trained` to the client, then the client has no way to determine the current epoch number. The `_ num_samples_seen` and `_num_samples_trained` only differ by the ratio. I suppose only reporting one of them is enough?

robinholzi: @MaxiBoether Probably there are only two, I was not fully sure what you meant with `how many samples in total we passed` exactly. If this is the total number of samples in the training interval set by the supervisor, then there are only two metrics!

@XianzheMa Generally, I think, the progress bar reporting is not extremely critical and should just give a rough estimation of the progress. I would argue that the exact number isn't too critical there. It's more important to have the different metrics available in the supervisor / the supervisor log file.

MaxiBoether: Hmh, I am a bit confused. Generally, yes, complete logs are the most important thing. With the current implementation (for me at least), the warmup triggers show twice the number of epochs. As long as the epoch count is correct and the relative progression of the progress bar makes sense, I guess reporting either is ok. Having the correct absolute number ( how many samples have we trained on / how many samples will we train on with train in terms of backward pass) would be best but if it's a bigger problem let's go for the easier solution

XianzheMa: Okay. So I will record both `_ num_samples_seen`, `_num_samples_trained` in the logs, and report the former to the client as the current solution!

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -379,7 +380,8 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
         total_stopw.stop("TotalTrain")
 
         self._info(f"Finished training: {self._num_samples} samples, {batch_number + 1} batches.")
-        self._log["num_samples"] = self._num_samples
+        self._log["num_samples_seen"] = self._num_samples
```
MaxiBoether, line 383: I think it still somewhat is inconsistent now. 

When using no downsampling, num_samples_trained == num_samples_seen. 

When using BtS, num_samples_trained is lower than num_samples_seen, and we report num_samples_seen to the client.

When using StB, num_samples_trained == num_samples_seen again and we report num_samples_seen. This is why I proposed to just report the number of samples we'll train on to the client, because in StB we do that already anyways, unless I am missing something. I might also be overlooking something, but I think the three cases Stb/Bts/no sampling should be handled in the same way

MaxiBoether (changes requested):

modyn/trainer_server/internal/trainer/pytorch_trainer.py
XianzheMa: I didn't completely get your points:
When using StB, how do you define `num_samples_seen`? Should it be the same as `num_samples_trained`?

What do you think is inconsistent under the current changes?

modyn/trainer_server/internal/dataset/key_sources/local_key_source.py
```diff
@@ -33,11 +33,11 @@ def get_num_data_partitions(self) -> int:
...
         self._trigger_sample_storage.clean_trigger_data(self._pipeline_id, self._trigger_id)
 
-    def __getstate__(self):
+    def __getstate__(self) -> Dict[str, Any]:
         state = self.__dict__.copy()
         del state["_trigger_sample_storage"]  # not pickable
         return state
 
-    def __setstate__(self, state):
+    def __setstate__(self, state: Dict[str, Any]) -> None:
```
XianzheMa, line 41: @MaxiBoether really curious how the CI passes on your PR :D
any repository founder privilege?

modyn/trainer_server/internal/trainer/pytorch_trainer.py
MaxiBoether: My point is that in BtS and StB, we train and see the same number of samples. As a user, I don't expect different numbers to show up in my client depending on whether I use BtS or StB. My intuition would be to show the number of training samples. If we show the number of samples seen, ok, but currently, in BtS, we show a progress bar with a higher absolute limit than in StB. But whatever, it's just visual. I guess we should focus on sth else

XianzheMa: Currently as we only report the number of samples seen, in BtS and StB the number is the same, so no unexpected inconsistency will appear right?

I agree with you that reporting only number of samples seen to the user is suboptimal, but anyway this PR is an improvement as it fixes an issue (also @robinholzi 's requirement is addressed), can we approve / merge this first? :D

wanna work on some other real problems first

MaxiBoether (approved):

PR #520, opened 2024-06-18
Fix various issues in RS2

MaxiBoether:
We had a couple of issues with RS2 (config not being in the union, when the training data is smaller than the expected training size). 

Part 5/n of porting over SIGMOD changes.

### DIFF
--- a/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_rs2_downsampling.py
+++ b/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_rs2_downsampling.py
@@ -201,3 +201,33 @@ def test_multiple_epochs_without_replacement_leftover_data(dummy_system_config: 
 
             assert all(idx in sample_ids for idx in selected_ids)
             assert len(set(selected_ids)) == len(selected_ids)
+
+
+def test_multiple_epochs_empty_without_replacement_leftover_data(dummy_system_config: ModynConfig):
+    pipeline_id = 0
+    trigger_id = 0
+    batch_size = 32
+    params_from_selector = {"replacement": False, "downsampling_ratio": 40}
+    per_sample_loss = None
+    device = "cpu"
+
+    downsampler = RemoteRS2Downsampling(
+        pipeline_id,
+        trigger_id,
+        batch_size,
+        params_from_selector,
+        dummy_system_config.model_dump(by_alias=True),
+        per_sample_loss,
+        device,
+    )
+    with torch.inference_mode(mode=(not downsampler.requires_grad)):
+        sample_ids = []
+        data = torch.tensor([])
+        target = torch.tensor([])
+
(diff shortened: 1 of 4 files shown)


modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rs2_downsampling.py
```diff
@@ -57,8 +57,13 @@ def _epoch_step_wr(self, target_size: int) -> None:
         self._current_subset = 0
 
     def _epoch_step_no_r(self, target_size: int) -> None:
-        max_subset = len(self._all_sample_ids) // target_size
+        if (max_subset := len(self._all_sample_ids) // target_size) == 0:
+            self._current_subset = 0
+            self._subsets = [[]]
+            return
+
```
XianzheMa, line 64: I am quite confused why this could happen:
This happens when `len(self._all_sample_ids)` < `target_size`. However we have `target_size = max(int(self.downsampling_ratio * len(self._all_sample_ids) / 100), 1)`. As long as `self.downsampling_ratio < 100`, which is naturally the case, the if condition here should never gets evaluated to True?

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_remote_rs2_downsampling.py
```diff
@@ -201,3 +201,33 @@ def test_multiple_epochs_without_replacement_leftover_data(dummy_system_config:
...
+        sample_ids = []
+        data = torch.tensor([])
+        target = torch.tensor([])
+
+        for _ in range(3):
+            downsampler.inform_samples(sample_ids, data, data, target)
+
+            selected_ids, weights = downsampler.select_points()
+            assert len(selected_ids) == 0
+            assert weights.shape == (0,)
```
XianzheMa, line 233: why could this happen?

XianzheMa (approved):
The changes look good, but I am just not sure why they could happen.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_rs2_downsampling.py
MaxiBoether: I think this will only be the case if `len(self._all_sample_ids) == 0`.  This could happen if we have a trigger where we have fewer samples than the batch size, in which case, the data loader does not output a single batch. Then, target size is 1, and sample ids is 0.

XianzheMa: Ah make sense! though I am then further curious what triggering policy leads to a trigger with so few samples ;D

MaxiBoether: it was more CGLM being very small for the first triggers :D

PR #521, opened 2024-06-18
More robust gRPC connections, using `tenacity` where possible

MaxiBoether:
Sometimes, we face outages/random disconnections during training. This fixes it in places where I encountered it last night. I tried to integrate `tenacity` as suggested by @robinholzi, but it's not always possible since the retry logic involves keeping track of already done work, which I don't want to put into class state

Part 6/n of porting over SIGMOD changes.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -24,6 +24,7 @@ dependencies:
     - types-protobuf==5.26.*
     - evidently==0.4.27
     - alibi-detect==0.12.*
+    - tenacity
   - jsonschema
   - psycopg2
   - sqlalchemy>=2.0
--- a/modyn/evaluator/internal/dataset/evaluation_dataset.py
+++ b/modyn/evaluator/internal/dataset/evaluation_dataset.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from typing import Callable, Generator, Iterable, Optional
 
 import grpc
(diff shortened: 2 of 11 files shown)


robinholzi:
Not fully what the problem with partial result states is, but have you considered using [retry contexts](https://tenacity.readthedocs.io/en/latest/#retrying-code-block)?

MaxiBoether:
> @MaxiBoether Not fully what the problem with partial result states is, but have you considered using [retry contexts](https://tenacity.readthedocs.io/en/latest/#retrying-code-block)?

Ah, nope. I guess I can refactor the code to retry contexts then

MaxiBoether:
> @MaxiBoether Not fully what the problem with partial result states is, but have you considered using [retry contexts](https://tenacity.readthedocs.io/en/latest/#retrying-code-block)?

Thanks for suggesting that. I adapted the code. However, I needed to manually catch and reraise exceptions sometimes because I find their callback system to be quite ugly in object-oriented contexts (we need to write a static function that recovers `self` out of args[0] etc). I think this actually looks better than using callbacks, and it allows for easier logging with our logging infra. The retry logic itself is still more hidden now :)

robinholzi (approved):
Thanks for the refactoring! Looks good now :)

PR #522, opened 2024-06-18
Small stability and logging fixes

MaxiBoether:
This concludes the important changes I have on the SIGMOD branch. These changes are too small to open more PRs for them, so I grouped them in a single PR.

### DIFF
--- a/modyn/trainer_server/internal/dataset/key_sources/local_key_source.py
+++ b/modyn/trainer_server/internal/dataset/key_sources/local_key_source.py
@@ -32,3 +32,12 @@ class LocalKeySource(AbstractKeySource):
 
     def end_of_trigger_cleaning(self) -> None:
         self._trigger_sample_storage.clean_trigger_data(self._pipeline_id, self._trigger_id)
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        del state["_trigger_sample_storage"]  # not pickable
+        return state
+
+    def __setstate__(self, state):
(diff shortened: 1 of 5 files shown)


modyn/trainer_server/internal/dataset/key_sources/local_key_source.py
```diff
@@ -32,3 +32,12 @@ def get_num_data_partitions(self) -> int:
...
         self._trigger_sample_storage.clean_trigger_data(self._pipeline_id, self._trigger_id)
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        del state["_trigger_sample_storage"]  # not pickable
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._trigger_sample_storage = TriggerSampleStorage(self.offline_dataset_path)
```
XianzheMa, line 43: when do we pickle/unpickle local_key_source? Didn't find it in any place

XianzheMa (approved):

modyn/trainer_server/internal/dataset/key_sources/local_key_source.py
MaxiBoether: The trainer server uses multiple data loading processes (workers). When they are started, `local_key_source` is pickled as part of the OnlineDataset being pickled. What was a bit weird to me is that it does not always occur (it should from my understanding), but nevertheless after implementing this it never happened again

PR #523, opened 2024-06-18
refactor: Simplify yearbook data generation

robinholzi:
# Motivation

As yearbook is a rather small dataset, I propose to always store the `train`, `test` and `all` (new) variants.

# Changes

- Always output all three splits of the datasets
- Use pathlib instead of os.path for cleaner code


MaxiBoether (approved):
👍  (if CI runs through)

PR #528, opened 2024-06-19
fix: Adjust active model definition

robinholzi:
# Motivation

Previously there were some issues with regard to the active model determination. Also, we want to be able to find the models in intervals that are currently trained.


modyn/supervisor/internal/eval/handler.py
```diff
@@ -15,7 +15,22 @@ class EvalRequest(BaseModel):
...
+            "Whether the model is currently used for inference and can therefore be considered the active. "
+            "This means that the evaluation interval starts after the model was trained. "
+            "Only available for `after_pipeline` evaluations"
+        ),
+    )
+    currently_trained_model: bool | None = Field(
+        None,
+        description=(
+            "Whether the interval is in a range whether the model is currently trained on. "
+            "For a fixed interval this is typically the model after the currently active model. "
```
MaxiBoether, line 30: Isn't it before the currently active model? First we are trained, for the next interval it's active

modyn/supervisor/internal/eval/handler.py
```diff
@@ -83,56 +85,88 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
         assert {"trigger_id", "training_id", "id_model", "first_timestamp", "last_timestamp"} - (
             set(df_trainings.columns)
         ) == set()
+        df_trainings = df_trainings.copy()
```
MaxiBoether, line 88: Any reason for this copy or leftover?

modyn/supervisor/internal/eval/handler.py
```diff
@@ -83,56 +85,88 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
         )
 
         # now build & potentially reduce the Trainings x EvalIntervals space
         df_cross = df_eval_intervals.merge(df_trainings, how="cross")
 
-        # to check if a combination is the most recent, we first compute if model was trained before the usage starts
-        # last_timestamp (df_trainings) defines the end of the training data;
-        # most_recent_model_interval_end_before (df_eval_intervals): defines center of an eval intervals.
-        df_cross["chronology_ok"] = df_cross["last_timestamp"] <= df_cross["most_recent_model_interval_end_before"]
+        # to check if a combination is the most recent or currently, we first compute if model was trained before
```
MaxiBoether, line 114: most recent is told terminology. Is that referring to currently active or trained?

modyn/supervisor/internal/eval/strategies/abstract.py
```diff
@@ -13,9 +13,9 @@ class EvalInterval:
     end: int | None
     """The end timestamp of the evaluation interval handed over to the evaluation server."""
 
-    most_recent_model_interval_end_before: int | None
+    active_model_trained_before: int | None = None
     """When determining the most recent model for this EvalInterval, we choose the newest model (highest model_id)
-        that satisfies the constraint model_training_interval_end <= most_recent_model_interval_end_before."""
+        that satisfies the constraint model_training_interval_end <= active_model_trained_before."""
```
MaxiBoether, line 18: do we keep the equality?

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -273,6 +273,7 @@ def df(self) -> pd.DataFrame:
                     self.eval_request.training_id,
                     self.eval_request.id_model,
                     self.eval_request.most_recent_model,
```
MaxiBoether, line 275: Most recent model is gone from the evaluator. This should crash, shouldn't it?

modyn/supervisor/internal/pipeline_executor/models.py
```diff
@@ -285,6 +286,7 @@ def df(self) -> pd.DataFrame:
                 "training_id",
                 "id_model",
                 "most_recent_model",
```
MaxiBoether, line 288: also here

modyn/supervisor/internal/eval/handler.py
```diff
@@ -83,56 +85,88 @@ def get_eval_requests_after_pipeline(self, df_trainings: pd.DataFrame) -> list[E
...
+        # from all models that have a strictly bigger id than the most recent model. Hence it is the model after the
+        # most recent model.
+        # For that we first build a model -> successor model mapping:
+        model_successor_relation = df_active_models[["id_model"]].drop_duplicates().sort_values(by="id_model")
+        model_successor_relation["next_id_model"] = model_successor_relation["id_model"].shift(-1, fill_value=-1)
+
+        # if there's no active model for the first interval(s), we still need to define the next model as the
+        # trained model
+        model_successor_relation = pd.concat(
+            [model_successor_relation, pd.DataFrame([{"id_model": None, "next_id_model": 21}])]
```
MaxiBoether, line 141: Where is the 21 coming from?

modyn/supervisor/internal/eval/handler.py
robinholzi: Your are looking at a mapping from model -> interval. We should take the interval -> model perspective. For one certain interval we have one active model and then we start training another one (but are not finished yet in the current interval). Therefore the `currently_trained_model` has a higher id than the `currently_active_model`. For one fixed model the interval using it as `currently_trained_model` obviously comes before the interval with the model as `currently_active_model`.

modyn/supervisor/internal/eval/handler.py
robinholzi: because of call by reference. I wanted to avoid weird side effects at the caller's side. the training dataframe is usually < 1k lines, so performance wouldn't be a big concern.

modyn/supervisor/internal/eval/strategies/abstract.py
robinholzi: no docs were outdated, good catch!

modyn/supervisor/internal/pipeline_executor/models.py
robinholzi: I think it's only used in the analytics tool, so I didn't notice so far.

modyn/supervisor/internal/pipeline_executor/models.py
robinholzi: I've replace all occurrences except from the ones in `/analytics` (those I will patch tomorrow)

modyn/supervisor/internal/eval/handler.py
robinholzi: Uff, that shouldn't be there, artifact debugging... I wanted to use the minimum model as it has no predecessor.

modyn/supervisor/internal/eval/handler.py
MaxiBoether: Sorry, brain fart. Makes complete sense...

modyn/supervisor/internal/eval/handler.py
MaxiBoether: Ok perfect!

MaxiBoether (approved):

PR #529, opened 2024-06-19
Fix replacement option in loss and gradnorm downsampling

MaxiBoether:
There was an unused replacement option which defaulted to True, which caused multiple keys to be chosen, which caused the storage to crash since it cannot deal with multiple keys in its current implementation. However, it should not have happened in the first place.

### DIFF
--- a/modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
+++ b/modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
@@ -50,8 +50,6 @@ class AbstractRemoteDownsamplingStrategy(ABC):
         assert "downsampling_ratio" in params_from_selector
         self.downsampling_ratio = params_from_selector["downsampling_ratio"]
 
-        self.replacement = params_from_selector.get("replacement", True)
-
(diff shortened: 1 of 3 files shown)


modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
```diff
@@ -50,8 +50,6 @@ def __init__(
         assert "downsampling_ratio" in params_from_selector
         self.downsampling_ratio = params_from_selector["downsampling_ratio"]
 
-        self.replacement = params_from_selector.get("replacement", True)
-
```
XianzheMa, line 54: sorry I have noticed this issue since a month ago then forgot to correct it :( otherwise everything has gone smoother

XianzheMa, line 54: sorry I have noticed this issue since a month ago then forgot to correct it :( otherwise everything has gone smoother

XianzheMa (approved):

XianzheMa:
let's just merge this I don't think any problem will be caused

PR #533, opened 2024-06-20
Fix batch number

XianzheMa:
Previously, we didn't record the number of passed batches correctly:
We use a `batch_number` which is generated purely from enumeration in `dataloader`. Therefore this number is irrelevant to the number of epochs (only shows how many batches there are in one epoch).
A similar issue exists on the iteration on `StB` when we calculate scores class by class. The number of batches passed in the previous class is not correctly accumulated on the current class.

This PR fixes it.

### DIFF
--- a/modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
+++ b/modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
@@ -42,19 +42,11 @@ class AbstractDownsamplingStrategy(ABC):
         self.requires_remote_computation = True
         self.maximum_keys_in_memory = maximum_keys_in_memory
         self.downsampling_config = downsampling_config
-        self.status_bar_scale = self._compute_status_bar_scale()
-
-    def _compute_status_bar_scale(self) -> int:
-        """
-        This function is used to create the downsampling status bar and handle the training one accordingly.
-
-        For BTS, we return 100 since the training status bar sees all the samples
-        For STB, we return the downsampling_ratio since the training status bar sees only a fraction of points
-        (while the downsampling status bas sees all the points)
-        """
-        if self.downsampling_mode == DownsamplingMode.BATCH_THEN_SAMPLE:
-            return 100
-        return self.downsampling_ratio
+        # the status bar scale is used in conjunction with the total number of samples (after presampling)
+        # and the number of currently trained batches to show current training progress
+        # No matter it is BtS or StB, the number of trained batches should be compared to the total number of samples
+        # divided by the downsampling ratio. Therefore, the status bar scale should be the downsampling ratio.
+        self.status_bar_scale = self.downsampling_ratio
 
     @property
     def downsampling_params(self) -> dict:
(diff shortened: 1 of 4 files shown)


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -374,18 +375,18 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
 
         total_stopw.stop("TotalTrain")
 
-        self._info(f"Finished training: {self._num_samples} samples, {batch_number + 1} batches.")
+        self._info(f"Finished training: {self._num_samples} samples, {passed_batches} batches.")
         self._log["num_samples"] = self._num_samples
         self._log["num_samples_trained"] = trained_batches * self._batch_size
-        self._log["num_batches"] = batch_number + 1
+        self._log["num_batches"] = passed_batches
```
MaxiBoether, line 381: Should we log passed batches and trained batches separately?

MaxiBoether (approved):
Thanks for catching this. My only comment is whether we should log both numbers :)

modyn/trainer_server/internal/trainer/pytorch_trainer.py
XianzheMa: added

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
```diff
@@ -42,19 +42,11 @@ def __init__(
...
-        (while the downsampling status bas sees all the points)
-        """
-        if self.downsampling_mode == DownsamplingMode.BATCH_THEN_SAMPLE:
-            return 100
-        return self.downsampling_ratio
+        # the status bar scale is used in conjunction with the total number of samples (after presampling)
+        # and the number of currently trained batches to show current training progress
+        # No matter it is BtS or StB, the number of trained batches should be compared to the total number of samples
+        # divided by the downsampling ratio. Therefore, the status bar scale should be the downsampling ratio.
+        self.status_bar_scale = self.downsampling_ratio
```
XianzheMa, line 49: There are a bit more changes involved @MaxiBoether , as I found the visual status bar on client side could still be inaccurate. It's in a confusing state but I propose to solve it ultimately with the following approach:

The client gets 4 numbers:
1. the total number of samples after presampling
2. status bar scale
3. the number of seen samples for StB downsampling
4. the number of *trained* samples

On client side, [1, 3] are compared to show StB downsampling progress. [1, 2, 4] are used to show training progress by comparing `total_num_samples / status_bar_scale` with `num_trained_samples`. Now no matter what mode it is in, it is always correct.

What also is changed is in `pytorch_trainer`, the number fed into `update_queue` now are the number of trained batches and the number of trained samples.

MaxiBoether: I think this sounds reasonable. If I understand correctly, for training, you basically want to show `number of trained samples / number of samples we will train on` - which I think is the best solution!

MaxiBoether (approved):

modyn/selector/internal/selector_strategies/downsampling_strategies/abstract_downsampling_strategy.py
XianzheMa: yes exactly!

PR #534, opened 2024-06-20
feat: More sophisticated evaluation logic

robinholzi:
# Motivation

- Analytics tools showed performance issues when run with multiple big logfiles.
- Naming ambiguities in plotting logic
- No support for averaging over multiple runs

# Changes

- Introduce an aggregation tool that transforms multiple pipeline logs into one aggregated (averaged) logfile.
- move analytics plotting app dataframe state into global variable to not reload/rebuild data every time someone selects another pipeline.
- Improve performance of dataframe construction from logfile by not concatenating size 1 dataframes but collecting all logs in tuples and then creating dataframes from them.


robinholzi:
WARNING: state management for the `composite_model_variant` is currently broken for anything but the `Evaluation over time` plot (for the other plot we need to adapt the `_SharedData` state classes like done in `eval_over_time` and not pass composite_model_variant via function argument.

MaxiBoether (approved):
Seems to work smoothly, for now :-) From my side, we can merge this. Thanks so much for the hard work here, this makes analyzing the results so much easier.

robinholzi:
> Seems to work smoothly, for now :-) From my side, we can merge this. Thanks so much for the hard work here, this makes analyzing the results so much easier.

Thanks @MaxiBoether, great to hear that!

PR #535, opened 2024-06-20
fix: Adjust evaluation business logic

robinholzi:
# Motivation

- We want to use the `real_last_timestamp` (start of next training interval - 1 --> marking end of current training interval) only for plotting the boxes in the heatmap plot. For decisions w.r.t. currently active models this doesn't for. E.g. if the next year in a dataset has no data at the year start, our training interval would extend into this next year and therefore it's model won't be considered for the current evaluation interval.
- timetriggers should allow starting a statically defined start point (and not with the first sample), otherwise, the whole schedule is off a couple of days.

# Note

Independently of the bug we fix with the `start_timestamp` in `timetrigger`, this setting allows us to effectively do `pre-training` with the first trigger (e.g. in the continous arxiv dataset we have sparse data from 1988 - ~2005). We could simply start the schedule in year 2005, then the first trigger trains on all previous data.

### DIFF
--- a/modyn/supervisor/internal/utils/time_tools.py
+++ b/modyn/supervisor/internal/utils/time_tools.py
@@ -1,0 +1,18 @@
+import pandas as pd
+
+
+def generate_real_training_end_timestamp(df_trainings: pd.DataFrame) -> pd.Series:
+    """
+    For sparse datasets we want to use next_training_start-1 as training interval end instead of last_timestamp
+    as there could be a long gap between the max(sample_time) in one training batch and the min(sample_time) in
+    the next training batch.
+    e.g. if we want to train for 1.1.2020-31.12.2020 but only have timestamps on 1.1.2020, last_timestamp
+    would be 1.1.2020, but the next training would start on 1.1.2021.
+
+    Args:
+        df_trainings: The pipeline stage execution tracking information including training and model infos.
+
+    Returns:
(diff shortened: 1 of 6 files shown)


modyn/supervisor/internal/utils/time_tools.py
```diff
@@ -0,0 +1,18 @@
+import pandas as pd
+
+
+def generate_real_training_end_timestamp(df_trainings: pd.DataFrame) -> pd.Series:
```
robinholzi, line 4: currently unused, will be used in analytics tool

MaxiBoether (approved):
LGTM. I guess in most cases one should set the start timestamp to 0, if they want a neat triggering boundary. But I guess finding the correct setting here is the task of the user.

MaxiBoether:
i will wait for integrationtests to run through for this one and then merge

robinholzi:
> LGTM. I guess in most cases one should set the start timestamp to 0, if they want a neat triggering boundary. But I guess finding the correct setting here is the task of the user.

For continuous datasets with periodic evaluation not necessarily. There you don't really care where it starts and what the exact bounds are, you just care about time proximity.

robinholzi:
@MaxiBoether could you kick off a cglm pipeline after merging this? I need a pipeline log file with correct training intervals (not affected by the first sample)

PR #537, opened 2024-06-20
feat: Log commit sha

robinholzi:
# Motivation

To be able to link a pipeline rune with the code that has executed it, we want to log the git sha of the last commit in the pipeline logfile.

### DIFF
--- a/modyn/supervisor/internal/utils/git_utils.py
+++ b/modyn/supervisor/internal/utils/git_utils.py
@@ -1,0 +1,9 @@
+import subprocess
+
+
+def get_head_sha() -> str:
+    """When launched in a git repository, return the SHA of the current HEAD."""
+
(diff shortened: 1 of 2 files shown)


MaxiBoether (approved):
n🍨

modyn/supervisor/internal/utils/git_utils.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+
+def get_head_sha() -> str:
+    """When launched in a git repository, return the SHA of the current HEAD."""
+
+    result = subprocess.run(["git", "rev-parse", "HEAD"], capture_output=True, text=True, check=True)
```
MaxiBoether, line 7: Actually I think you might need to set the cwd here. I guess this would take the cwd of wherever we start the script, but I guess the most robust way would be using Path(__file__).parent or something, then we for sure have the correct git reference

MaxiBoether (approved):

Issue #488, opened 2024-06-10
Log training loss


MaxiBoether:
We should at the trainer server (optionally) collect the training loss as well and put it into the trainer server log. this might help to understand performance on the training set without requiring to eval on it.

XianzheMa:
on it

PR #539, opened 2024-06-21
Record training loss

XianzheMa:
Per the PR title, this PR solves issue https://github.com/eth-easl/modyn/issues/488.


modyn/common/grpc/grpc_helpers.py
```diff
@@ -216,40 +216,38 @@ def prepare_start_training_request(
...
+            num_prefetched_partitions=training_config.num_prefetched_partitions,
+            parallel_prefetch_requests=training_config.parallel_prefetch_requests,
+            seed=training_config.seed,  # seed is an optional field which can accept None
+            # tokenizer is an optional field which can accept None
+            tokenizer=PythonString(value=tokenizer) if tokenizer is not None else None,
+            num_samples_to_pass=num_samples_to_pass if num_samples_to_pass is not None else 0,
+            shuffle=training_config.shuffle,
+            enable_accurate_gpu_measurements=training_config.enable_accurate_gpu_measurements,
+            record_loss_every=training_config.record_loss_every,
+        )
```
XianzheMa, line 250: for possibly optional value I have commented (e.g. `tokenizer` and `seed`). Other values have non-optional type and won't be None.

MaxiBoether: I don't get these changes. `cleaned_kwargs` explicitly removes the None fields from the request. Isn't that what we want? For example, now we always pass num_samples_to_pass = 0 to the trainer server, which is a semantic change from before. Why are we changing this?

MaxiBoether (changes requested):
The loss logic itself is ok but I am confused about the gRPC changes since we explicitly did not want to pass the None arguments because in gRPC there is a difference between passing None and no value at all, if I remember correctly. We had issues before with this

modyn/common/grpc/grpc_helpers.py
XianzheMa: My motivation is that it is the only place how we pack the grpc request like this; at the other places we just pass args directly. And it is a bit cumbersome to write the literal arg name in quotes, rather than benefiting from IDE code completion.

Re `num_samples_to_pass`: this is an non-optional field which cannot accept None value (would be captured by mypy, that's why I changed it). 0 is the default value if not set, or set explicitly to 0. It is set to 0 anyway. We detect 0 explicitly on the other side https://github.com/eth-easl/modyn/blob/bf96bfa546c829a4844f25e0143497deb9539e04/modyn/trainer_server/internal/trainer/pytorch_trainer.py#L215.

For optional fields, setting a `None` is exactly the same as not setting it as all (but when reading the field, we read the default value instead of `None`). We have the `start_timestamp` and the `end_timestamp` in evaluation requests as optional fields, https://github.com/eth-easl/modyn/blob/bf96bfa546c829a4844f25e0143497deb9539e04/modyn/protos/evaluator.proto#L15, and we set the to None by default. Nothing has been wrong.

But I am not strongly opinioniated for the change (just consider as an ease of life). If you prefer the previous approach I would revert it :)

MaxiBoether: I tried to find our old discussions on gRPC default fields, but could not find it. I guess intuitively you're definitely right, it should not make a difference. If it doesn't break, we can merge it with your changes

MaxiBoether (approved):
Thanks. i tried to dig into the grpc optional thing again but could not find my old meeting notes with Francesco. I guess it should be ok

Issue #505, opened 2024-06-13
Potentially wrong gradient matrix


MaxiBoether:
This is a duplicate of https://github.com/PatrickZH/DeepCore/issues/13.

In our [AbstractMatrixDownsamplingStrategy](https://github.com/eth-easl/modyn/blob/main/modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py ) implementation, we use the same matrix as defined by DeepCore. As discussed in this issue, the definition of the matrix looks weird, and I am also confused that we don't even need the actual gradient (with torch.no_grad()). We should check out whether this actually makes sense.

XianzheMa:
on it

XianzheMa:
After the potential fix PR, https://github.com/eth-easl/modyn/pull/540, I realize there could be a reason for their way of calculating the gradients.

Essentially, we should use the *full gradients* to select points. But since full gradients are expensive to compute, researchers propose to use the *last layer gradients* to approximate full gradients.

Since it is possible to use *last layer gradients* to approximate, it is also possible to use *gradients of second-to-last layer*, which is the gradients of loss regarding to the embedding. Since the mapping from embedding to output is linear, then the gradients of second-to-last layer is essentially last layer gradients scaled by the mapping layer, which is what they did.

MaxiBoether:
I guess it's not fully clear to me yet:

1) From my understanding indeed, the papers proposing the initial techniques use the last layer gradients. 

2) I guess it's not fully clear why what they do is equivalent to scaling the last layer gradient. For the bias, they take the last layer gradients, and concatenate that with the bias gradient * embedding? I don't immediately see how tat is equivalent to taking `d last layer / d penultimate layer`.

What is your proposal here? Should we keep the DeepCore implementation? Should we add a flag that allows to switch between DeepCore and your PR? I am not sure I fully understand the issue. My gut feeling is that we should follow the literature, but I wouldn't have a problem with adding a flag to switch between both implementations.

XianzheMa:
> I guess it's not fully clear to me yet:
> 
> 1. From my understanding indeed, the papers proposing the initial techniques use the last layer gradients.
> 2. I guess it's not fully clear why what they do is equivalent to scaling the last layer gradient. For the bias, they take the last layer gradients, and concatenate that with the bias gradient * embedding? I don't immediately see how tat is equivalent to taking `d last layer / d penultimate layer`.
> 
[...]

PR #540, opened 2024-06-22
Adding another way of calculating gradients in downsamplers

XianzheMa:
This PR addresses https://github.com/eth-easl/modyn/issues/505.
A flag is introduced, to switch between using purely the last layer gradients, or the concatenation of last layer gradients and penultimate layer gradients, in selection algorithms.


modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -9,7 +9,11 @@
 )
 from modyn.trainer_server.internal.trainer.remote_downsamplers.deepcore_utils.shuffling import _shuffle_list_and_tensor
 
-MatrixContent = Enum("MatrixContent", ["EMBEDDINGS", "GRADIENTS"])
+
+class MatrixContent(Enum):
+    EMBEDDINGS = 1
+    LAST_LAYER_GRADIENTS = 2
+    LAST_TWO_LAYERS_GRADIENTS = 3
```
XianzheMa, line 17: The previous way is not the canonical way to write an enum

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_per_label_remote_downsample_strategy.py
```diff
@@ -6,17 +6,6 @@
...
-        self,
-        pipeline_id: int,
-        trigger_id: int,
-        batch_size: int,
-        params_from_selector: dict,
-        modyn_config: dict,
-        device: str,
-    ):
-        super().__init__(pipeline_id, trigger_id, batch_size, params_from_selector, modyn_config, device)
-        self.requires_data_label_by_label = True
```
XianzheMa, line 19: many subclasses still overwrite this field `requires_data_label_by_label` to `False`. Thus I leave it for subclass to decide. (actually only Craig is the only subclass that previous does not overwrite this field itself.)

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_craig_downsampling.py
```diff
@@ -43,13 +46,16 @@ def __init__(
...
+
         if self.greedy not in OPTIMIZER_CHOICES:
             raise ValueError(
                 f"The required Greedy optimizer is not available. Pick one of the following: {OPTIMIZER_CHOICES}"
             )
 
-        # This class uses the embedding recorder
-        self.requires_coreset_supporting_module = True
+        self.requires_coreset_supporting_module = self.full_grad_approximation == "LastLayerWithEmbedding"
+        self.requires_data_label_by_label = True
```
XianzheMa, line 58: Craig always selects points label by label. Previously this field is set implicitly to True in the base class `AbstractPerLabelRemoteDownsamplingStrategy`

analytics/tools/aggregate_runs/core_aggregation.py
```diff
@@ -3,7 +3,6 @@
 from pathlib import Path
 
 import pandas as pd
-
```
MaxiBoether, line 6: nit, but no need to touch this file in this PR :D

modyn/config/schema/pipeline/sampling/downsampling_config.py
```diff
@@ -74,6 +77,7 @@ class GradMatchDownsamplingConfig(BaseDownsamplingConfig):
 
     strategy: Literal["GradMatch"] = "GradMatch"
     balance: bool = Field(False, description="If True, the samples are balanced.")
+    full_grad_approximation: FullGradApproximation = Field(default="LastLayer")
```
MaxiBoether, line 80: Could you either here or at the definition on `FullGradApproximation` add a description/comment on the meaning of this such that future users understand what this is doing?

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_craig_remote_downsampling.py
```diff
@@ -20,6 +20,7 @@ def get_sampler_config(modyn_config, balance=False):
         "balance": balance,
         "selection_batch": 64,
         "greedy": "NaiveGreedy",
+        "full_grad_approximation": "LastLayerWithEmbedding",
```
MaxiBoether, line 23: This affects all downsamplers: Do we have a test somewhere at least that we do not crash when using `LastLayer` instead? we don't need to write a test where we manually calculate something and then assert equivalence but having some form of validation it does not crash would be nice

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -61,12 +68,26 @@ def inform_samples(
...
         assert self.matrix_content is not None
-
-        if self.matrix_content == MatrixContent.GRADIENTS:
-            new_elements = self._compute_gradients(forward_output, target, embedding)
+        if self.matrix_content == MatrixContent.LAST_LAYER_GRADIENTS:
+            grads_wrt_loss_sum = (
+                self._compute_last_layer_gradient_wrt_loss_sum(self.criterion, forward_output, target).detach().cpu()
+            )
+            grads_wrt_loss_mean = grads_wrt_loss_sum / batch_size
+            new_elements = grads_wrt_loss_mean.detach().cpu()
```
MaxiBoether, line 78: no need to detach again, I think

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_per_label_remote_downsample_strategy.py
MaxiBoether: sounds good!

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
```diff
@@ -101,3 +103,51 @@ def select_points(self) -> tuple[list[int], torch.Tensor]:
...
+        embedding_dim = embedding.shape[1]
+        num_classes = forward_output.shape[1]
+        batch_num = target.shape[0]
+
+        with torch.no_grad():
+            bias_parameters_grads = torch.autograd.grad(loss, forward_output)[0]
+            weight_parameters_grads = embedding.view(batch_num, 1, embedding_dim).repeat(
+                1, num_classes, 1
+            ) * bias_parameters_grads.view(batch_num, num_classes, 1).repeat(1, 1, embedding_dim)
+            gradients = torch.cat([bias_parameters_grads, weight_parameters_grads.flatten(1)], dim=1)
```
MaxiBoether, line 152: I guess I am still confused where this every calculates the gradient for the second-to-last layer , I see that we have d loss / d output but then I only see .view and .repeat operations where I don't get how this would calculate a gradient, but this is also not really within this PR

MaxiBoether (approved):
Thanks, my only concern is (and I think codecov complains as well) that we should at least test whether the new code path does not crash or something. But the changes itself look good and I tried to validate that the old logic stay the same :)

modyn/config/schema/pipeline/sampling/downsampling_config.py
XianzheMa: Make sense. Done

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
XianzheMa: thanks for spotting it

modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_craig_remote_downsampling.py
XianzheMa: make sense. will do

MaxiBoether:
Thanks, the new tests look good. Feel free to merge!

PR #541, opened 2024-06-22
Implement `ratio_max` scaling for downsamplers

MaxiBoether:
Before, we only were able to enter ratios from 0-100 as percentages for downsamplers. With this PR, we allow scaling by a ratio_max factor.

### DIFF
--- a/modyn/config/schema/pipeline/config.py
+++ b/modyn/config/schema/pipeline/config.py
@@ -1,15 +1,16 @@
 from __future__ import annotations
 
-from typing import Optional
+from typing import Optional, Self
 
 from modyn.config.schema.base_model import ModynBaseModel
-from pydantic import Field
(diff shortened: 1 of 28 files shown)


modyn/config/schema/pipeline/config.py
```diff
@@ -32,3 +33,25 @@ class ModynPipelineConfig(ModynBaseModel):
     trigger: TriggerConfig
     selection_strategy: SelectionStrategy
     evaluation: EvaluationConfig | None = Field(None)
+
+    @model_validator(mode="after")
+    def validate_bts_training_selection_works(self) -> Self:
```
MaxiBoether, line 38: If we had this before, I would have run into this a lot earlier :D

XianzheMa (approved):
LGTM

PR #544, opened 2024-06-22
Add `ratio_max` for presampling strategies

MaxiBoether:
I realized we also need to set `ratio_max` for presampling strategies, otherwise they cannot run 12.5% as well.


XianzheMa (approved):
🚀

PR #546, opened 2024-06-24
feat: Refine triggering abstraction

robinholzi:
# Motivation

Let's refine the abstraction for trigger a bit to enable more powerful triggering configuration.

# Changes

Note: we will adjust the business logic in the future to use the new abstraction.

- New configuration models & refinement of existing models
- move trigger configs in dedicated files


modyn/config/schema/pipeline/trigger/ensemble.py
```diff
@@ -0,0 +1,69 @@
+
```
robinholzi, line 1: New, but not yet reflected in the business logic.

docs/pipeline/TRIGGERING.md
```diff
@@ -0,0 +1,59 @@
+# Modyn Triggering
+
+Alongside the simple triggers, Modyn also provides complex triggers that can be used to trigger the training of a model. 
+
+Despite the line being blurry, complex triggers are generally more sophisticated and require more information to be provided to the trigger - they most of the time cannot be pre-determined via a simple configuration entry but require
+reinjested information from the ongoing pipeline run.
```
MaxiBoether, line 6: reingested :D?

docs/pipeline/TRIGGERING.md
```diff
@@ -0,0 +1,59 @@
...
+
+Some complex triggers can only make decisions in batched intervals as they are cannot be efficiently computed
+on a per-sample basis. Here we can find the `DataDrift` and `CostBased` (not yet implemented) based triggers.
+
+Another policy type is the `EnsemblePolicy` which can be used to combine multiple triggers into a single trigger. This can be useful if multiple triggers should be evaluated before the training of a model is triggered.
```
MaxiBoether, line 11: Oh ok sure, sounds good, I mean nice to have but no urgent need for this right now I suppose

modyn/config/schema/pipeline/config.py
```diff
@@ -30,7 +30,7 @@ class ModynPipelineConfig(ModynBaseModel):
 
     training: TrainingConfig
     data: DataConfig
-    trigger: TriggerConfig
+    trigger: TriggerConfig  # type: ignore[valid-type]
```
MaxiBoether, line 33: why this change in mypy? it was not complaining, was it?

modyn/config/schema/pipeline/trigger/drift.py
```diff
@@ -0,0 +1,23 @@
...
+
+__TriggerConfig = ForwardRef("TriggerConfig", is_class=True)
+
+
+class DataDriftTriggerConfig(ModynBaseModel):
+    id: Literal["DataDriftTrigger"] = Field("DataDriftTrigger")
+
+    detection_interval: Optional[__TriggerConfig] = Field(  # type: ignore[valid-type]
+        None, description="The Trigger policy to determine the interval at which drift detection is performed."
+    )  # currently not used
```
MaxiBoether, line 16: I don't get this yet. How do you want to pass a trigger_config to this (and why)? can i pass a drift trigger to a drift trigger? could you elaborate a bit here?

modyn/config/schema/pipeline/trigger/ensemble.py
```diff
@@ -0,0 +1,70 @@
...
+        Function that aggregates the decisions of the individual triggers."""
+        raise NotImplementedError
+
+
+class MajorityVoteEnsembleStrategy(BaseEnsembleStrategy):
+    id: Literal["MajorityVote"] = Field("MajorityVote")
+
+    @property
+    def aggregate_decision_func(self) -> Callable[[dict[str, bool]], bool]:
+        return lambda decisions: sum(decisions.values()) > len(decisions) / 2
```
MaxiBoether, line 23: i am somewhat torn here since we now have business logic in the config files, but this is simple logic so it might be ok...

MaxiBoether (approved):
generally LGTM, had a short question on the new field on the data drift trigger

modyn/config/schema/pipeline/config.py
robinholzi: It was complaining in an earlier version. I think we can remove it now though! Thanks!

docs/pipeline/TRIGGERING.md
robinholzi: I wonder why the typo pre-commit hook didn't detect this😅

docs/pipeline/TRIGGERING.md
robinholzi: Yeah, I probably won't implement this in the next month.

modyn/config/schema/pipeline/config.py
robinholzi: Still required, unfortunately. 
image

modyn/config/schema/pipeline/trigger/drift.py
robinholzi: I noticed that drift the trigger only supports to be evaluated after every n-th sample. However, there's no real reason to not use a time schedule for assessing the drift trigger. Consequently, I wanted to support specifying a TimeTrigger or DataAmountTrigger. However, there's no real reason not to support other trigger types to as they follow the same interface. In practise you would want to use a cheaper policy though to determine the evaluation of the more expensive ones.

modyn/config/schema/pipeline/trigger/ensemble.py
robinholzi: Also considered that. I concluded though that this is the most accessible setup.

PR #547, opened 2024-06-24
Twin RHO Model Step 1: create the Twin RHO Model

XianzheMa:
This PR is the first PR to implement another way of producing holdout set, il model and irreducible loss (typically suitable for small datasets): 
> Split training set into two halves; train two IL models, each on one half. Each model provides the irreducible loss for samples that it was not trained on. The main model is still trained on the original training set (CIFAR10, CIFAR100, CINIC-10).

Our current architecture only allow one trigger id to correspond to one model id. To accommodate two il models within one trigger, I create a "twin model" which internally consists of two il models. During training, each il model will memorize the sample ids it has seen, so that during evaluation each il model will be used for the samples the model hasn't seen.

## How it works

1. At selector, `RHOLossDownsamplingStrategy` randomly samples half of the training set and mark the `used` column in `selector_state_metadata` table of those samples as `True`. The strategy issues a request to train a `RHOLOSSTwinModel` on this TSS. (unimplemented)
2. At trainer server, `RHOLOSSTwinModel` is instantiated. Only the 0th model is trained on this dataset (implemented in this PR).
3. At selector, `RHOLossDownsamplingStrategy` produces the other half of the training set by selecting the samples with `used==False`. The strategy issues a request to finetune this twin model. (unimplemented)
4. At trainer server, `RHOLOSSTwinModel` is instantiated again. Only the 1th model is trained on this dataset (implemented in this PR).
5. At selector, (optionally) clear the `used` flags.
6. At trainer server when training the main model: nothing needed to change as the logic is handled internally in the twin model.

Apparently it is not the optimal way to train a twin RHO model, but it's a very straightforward way and we can optimize it depending on how well it performs.

## Current drawbacks
Due to `used` `RHOLoss` will currently be not compatible with some presampling strategies that also use `used` fields such as `FreshnessSamplingStrategy`.

## Next PR
Implementing step 1 and 3: preparing the split holdout set. 

## How to review
All the main logic is in [modyn/models/rho_loss_twin_model/rho_loss_twin_model.py](https://github.com/eth-easl/modyn/pull/547/files#diff-0f510b51f60a2c4ee867551fed01763b09ca13d3b88b10aac5aca55d83377fdf)


modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+    def set_extra_state(self, state: dict) -> None:
+        self._models_seen_ids = state["models_seen_ids"]
+        # the second time we train on this model, we should switch to the other model
+        self._current_model = 1
+
+    def forward(self, data: torch.Tensor, sample_ids: Optional[list[int]] = None) -> torch.Tensor:
+        assert sample_ids is not None
+        if self.training:
```
MaxiBoether, line 49: where is self.training coming from? fields should be initialized in the constructor, maybe I don't see it. Usually pylint should complain here that you access a field not initialized in the constructor

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
```diff
@@ -0,0 +1,71 @@
...
+        self._models_seen_ids: list[set[int]] = [set(), set()]
+        self._current_model = 0
+
+    def get_extra_state(self) -> Any:
+        return {
+            "models_seen_ids": self._models_seen_ids,
+        }
+
+    def set_extra_state(self, state: dict) -> None:
+        self._models_seen_ids = state["models_seen_ids"]
```
MaxiBoether, line 43: yeah, if this is just a single item, not sure why we use a dict

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
MaxiBoether: also, why not make this decision based on self.current_model?

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
```diff
@@ -0,0 +1,71 @@
...
+        seen_by_model0 = torch.BoolTensor(
+            [sample_id in self._models_seen_ids[0] for sample_id in sample_ids], device=self.device
+        )
+        seen_by_model1 = torch.BoolTensor(
+            [sample_id in self._models_seen_ids[1] for sample_id in sample_ids], device=self.device
+        )
+
+        # assert that a sample is seen by at least one model
+        assert (seen_by_model0 | seen_by_model1).all()
+        # when a sample is seen by both models, we route it to model 0
```
MaxiBoether, line 69: why would that happen?

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
```diff
@@ -0,0 +1,71 @@
...
+        # unsqueeze to make seen_by_model1 broadcastable
+        return torch.where(seen_by_model1.unsqueeze(1), self._models[0](data), self._models[1](data))
```
MaxiBoether, line 71: doesn't this do unncessary computation? i.e. we compute both models but only take what we need? is there a way to make this more efficient?

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
MaxiBoether: but actually I do not understand yet how this is updated. can you maybe add a comment here explaining this? there is nothing in the trainer server that sets this, how does the model know?

MaxiBoether (changes requested):
The changes mostly look good, but I am a bit confused about who calls set_extra_state/updates the current model and it whether we can make this a bit less convoluted. It's not clear to me yet why this dict is necessary. Also I wonder whether we can avoid calling both models :)

modyn/models/dlrm/dlrm.py
```diff
@@ -111,7 +111,8 @@ def reorder_categorical_input(self, cat_input: torch.Tensor) -> torch.Tensor:
         order = self._embedding_ordering.expand(dim0, -1)
         return torch.gather(cat_input, 1, order)
 
-    def forward(self, data: torch.Tensor) -> torch.Tensor:
+    # pylint: disable-next=unused-argument
```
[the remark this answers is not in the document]
XianzheMa: there is no `del sample_ids` in this file. Sorry I didn't get it?

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
```diff
@@ -0,0 +1,71 @@
...
+        self._models = nn.ModuleList(
+            [
+                model_handler(rho_model_config, device, amp).model,
+                model_handler(rho_model_config, device, amp).model,
+            ]
+        )
+        self._models_seen_ids: list[set[int]] = [set(), set()]
+        self._current_model = 0
+
+    def get_extra_state(self) -> Any:
```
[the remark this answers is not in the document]
XianzheMa: This is a utility function that will be called internally by PyTorch, when we fetch the state dict by `.state_dict()` or directly save the model state via `torch.save`. See https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.get_extra_state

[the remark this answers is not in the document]
XianzheMa: This ensures that the `_models_seen_ids` are stored and loaded next time

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
XianzheMa: I will add it. It is an attribute in `nn.Module` that's updated everytime `.train()` or `.eval()` is called.
See https://discuss.pytorch.org/t/what-does-model-training-in-the-transfer-learning-tutorial-mean/82182

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
XianzheMa: good catch. As I reflect the logic here, I should assert that no sample is seen by both models.

But we could have a sample not seen by either one, because maybe we use a presampling policy that decides to include historic samples in this training set.

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
XianzheMa: You are right. I will change it: first check if both models have seen some samples in this batch, and only in this case will I do two forward passes.

modyn/config/schema/pipeline/sampling/downsampling_config.py
```diff
@@ -143,13 +143,26 @@ class RHOLossDownsamplingConfig(BaseDownsamplingConfig):
...
+        description="How much of the training set is used as the holdout set.",
         min=0,
         max=100,
     )
     il_training_config: ILTrainingConfig = Field(description="The configuration for the IL training.")
 
+    @model_validator(mode="after")
+    def validate_holdout_set_ratio(self) -> Self:
+        if self.holdout_set_strategy == "Twin" and self.holdout_set_ratio != 50:
+            raise ValueError("holdout_set_ratio should be 100 for the Twin strategy.")
```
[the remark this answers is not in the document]
XianzheMa: Does not matter actually. In the next PR internally I didn't use this number from configuration but a hard-coded number. This is here just to signify to the user that is has too be a constant.

modyn/models/rho_loss_twin_model/rho_loss_twin_model.py
XianzheMa: see https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.set_extra_state

modyn/models/dlrm/dlrm.py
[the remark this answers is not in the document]
MaxiBoether: My point is that you add a pylint nolint to supress a warning that you dont use a parameter. Instead, you should add a line del sample_ids since that avoids the nolint

MaxiBoether (approved):
Thanks! The updated logic looks good and avoids the double model call when possible. My only comment left is regarding the nolintS

PR #548, opened 2024-06-24
feat: Add unique sequence number for aggregation into supervisor

robinholzi:
# Motivation

We want to be able to average the execution time of pipeline stage executions over multiple pipeline runs to gain more confidence in our results. For that we need sequential identifiers to map the StageLog entries from all logfiles.


MaxiBoether (approved):
LGTM!

PR #550, opened 2024-06-26
Parameterize `drop_last`

XianzheMa:
The reason for this PR is that, for il model training, I don't want to drop the last batch, as the holdout set is already super small (and on il model training no sampling is used, so it does not harm not to drop last batch). But for main model training, usually we default `drop_last` to True.


modyn/config/schema/pipeline/training/config.py
```diff
@@ -98,6 +98,9 @@ class TrainingConfig(ModynBaseModel):
         description="The number of data loader workers on the trainer node that fetch data from storage.", ge=1
     )
     batch_size: int = Field(description="The batch size to be used during training.", ge=1)
+    drop_last_batch: bool = Field(
+        default=True, description="Whether to drop the last batch if it is smaller than the batch size."
```
XianzheMa, line 102: default to True, so we don't have a breaking change after merging this PR

MaxiBoether (approved):
The changes look good but cause a problem in the integration test. When this is fixed, feel free to merge

XianzheMa:
> The changes look good but cause a problem in the integration test. When this is fixed, feel free to merge

The failure is because we have a `_assert_data_size` logic added when merging the batch accumulation. This assertion happens regardless of we are doing batch accumulation or not. When we set `drop_last` to False, naturally the last batch has fewer samples and does not conform to the assertion. Let's hope this time CI passes

PR #551, opened 2024-06-26
feat: Create `DriftDetector` abstraction and introduce alibi-detect

robinholzi:
# Motivation

We want to be able to use both evidently and alibi-detect as they support different drift detection metrics as well as differ in their runtime performance.

For that we need a clean abstraction that allows us to delegate drift detection to the respective library

# Changes

- add `alibi-detect` dependency
- adjust drift detection abstraction via the `DriftDetector` interface
- introduce pydantic configurations models for alibi detect and evidently
- allow the collection and aggregation of multiple drift metrics.
- add tests for our embedding drift detection setup

# Note:

As this is rather big already, I'd like to get this merged before trying to run drift detection pipelines, so we might need slight adjustments in a followup PR.

### DIFF
--- a/experiments/arxiv/compare_trigger_policies/run.py
+++ b/experiments/arxiv/compare_trigger_policies/run.py
@@ -12,6 +12,8 @@ from modyn.config.schema.pipeline import (
 from modyn.config.schema.pipeline.evaluation.strategy.between_two_triggers import BetweenTwoTriggersEvalStrategyConfig
 from modyn.config.schema.pipeline.evaluation.strategy.periodic import PeriodicEvalStrategyConfig
 from modyn.config.schema.pipeline.trigger import DataDriftTriggerConfig
+from modyn.config.schema.pipeline.trigger.drift.aggregation import MajorityVoteDriftAggregationStrategy
+from modyn.config.schema.pipeline.trigger.drift.evidently import EvidentlyMmdDriftMetric
 from modynclient.config.schema.client_config import ModynClientConfig, Supervisor
 
 
@@ -65,14 +67,15 @@ def construct_pipelines() -> list[ModynPipelineConfig]:
         )
 
     for interval, threshold in [
-        (20_000, 0.5),
-        (100_000, 0.6),
-        (100_000, 0.7),
-        (100_000, 0.8),
-        (100_000, 0.9),
-        (250_000, 0.7),
-        (500_000, 0.7),
-        (1_000_000, 0.7),
+        (20_000, 0.15),
+        # (20_000, 0.5),
+        # (100_000, 0.6),
+        # (100_000, 0.7),
+        # (100_000, 0.8),
+        # (100_000, 0.9),
+        # (250_000, 0.7),
+        # (500_000, 0.7),
+        # (1_000_000, 0.7),
     ]:
         pipeline_configs.append(
             gen_pipeline_config(
@@ -80,8 +83,13 @@ def construct_pipelines() -> list[ModynPipelineConfig]:
                 trigger=DataDriftTriggerConfig(
                     detection_interval_data_points=interval,
                     sample_size=5000,
-                    metric="model",
-                    metric_config={"threshold": threshold},
+                    metrics={
+                        "ev_mmd": EvidentlyMmdDriftMetric(threshold=threshold),
+                        # "ev_model": EvidentlyMmdDriftMetric(threshold=threshold),
+                        # "ev_ratio": EvidentlyMmdDriftMetric(threshold=threshold),
+                        # "ev_distance": EvidentlyMmdDriftMetric(threshold=threshold),
+                    },
+                    aggregation_strategy=MajorityVoteDriftAggregationStrategy(),
                 ),
                 eval_handlers=eval_handlers,
             )
--- a/experiments/arxiv/compare_trigger_policies/run.py
+++ b/experiments/arxiv/compare_trigger_policies/run.py
@@ -13,7 +13,7 @@ from modyn.config.schema.pipeline.evaluation.strategy.between_two_triggers impor
 from modyn.config.schema.pipeline.evaluation.strategy.periodic import PeriodicEvalStrategyConfig
 from modyn.config.schema.pipeline.trigger import DataDriftTriggerConfig
 from modyn.config.schema.pipeline.trigger.drift.aggregation import MajorityVoteDriftAggregationStrategy
-from modyn.config.schema.pipeline.trigger.drift.evidently import EvidentlyMmdDriftMetric
+from modyn.config.schema.pipeline.trigger.drift.alibi_detect import AlibiDetectMmdDriftMetric
 from modynclient.config.schema.client_config import ModynClientConfig, Supervisor
 
 
@@ -66,8 +66,8 @@ def construct_pipelines() -> list[ModynPipelineConfig]:
             )
         )
 
-    for interval, threshold in [
-        (20_000, 0.15),
+    for interval in [
+        20_000,
         # (20_000, 0.5),
         # (100_000, 0.6),
         # (100_000, 0.7),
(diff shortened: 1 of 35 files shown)


modyn/tests/supervisor/internal/triggers/drift/test_evidently_detector.py
```diff
@@ -0,0 +1,100 @@
...
+        ("model", EvidentlyDriftDetector({"model": model_drift_metric})),
+        ("ratio", EvidentlyDriftDetector({"ratio": ratio_drift_metric})),
+        ("simple_distance", EvidentlyDriftDetector({"simple_distance": simple_distance_drift_metric})),
+    ]
+    for name, ad in detector:
+        assert isinstance(ad, EvidentlyDriftDetector)
+
+        # on h0
+        results = ad.detect_drift(df_data_ref, df_data_h0)
+        # if name == "mmd":
```
robinholzi, line 77: Commented out as locally I patched the evidently code.

robinholzi: for CI we need to include it (or open a PR in evidently)

experiments/arxiv/compare_trigger_policies/run.py
```diff
@@ -65,23 +67,29 @@ def construct_pipelines() -> list[ModynPipelineConfig]:
         )
 
     for interval, threshold in [
```
robinholzi, line 69: This will change anyways a bit in another PR

experiments/yearbook/compare_trigger_policies/run.py
```diff
@@ -76,8 +78,13 @@ def construct_pipelines() -> None:
                 trigger=DataDriftTriggerConfig(
                     detection_interval_data_points=interval,
                     sample_size=None,
-                    metric="model",
-                    metric_config={"threshold": threshold},
+                    metrics={
+                        "ev_mmd": EvidentlyMmdDriftMetric(threshold=threshold),
+                        # "ev_model": EvidentlyMmdDriftMetric(threshold=threshold),
```
robinholzi, line 83: TODO: add other metrics, leave it for now as this will change a bit anyways

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
```diff
@@ -0,0 +1,57 @@
...
+
+class _AlibiDetectCorrectionMixin(ModynBaseModel):
+    correction: Literal["bonferroni", "fdr"] = Field("bonferroni")
+
+
+class _AlibiDetectAlternativeMixin(ModynBaseModel):
+    alternative_hypothesis: Literal["two-sided", "less", "greater"] = Field("two-sided")
+
+
+class AlibiDetectMmdDriftMetric(_AlibiDetectBaseDriftMetric, AlibiDetectDeviceMixin):
```
MaxiBoether, line 33: How similar are Alibi Detect and Evidently? In a perfect world, instead of having  `AlibiDetect[XXX]Metric` and  `EvidentlyXXXMetric`, we would have an XX metric and a field where we either say alibi detect or evidently, i.e., it's a configuration option of the metric rather than a different metric. Do you think this is feasible?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -156,19 +93,31 @@ def inform(self, new_data: list[tuple[int, int, int]]) -> Generator[int, None, N
             if detection_idx_end <= self.leftover_data_points:
                 continue
 
-            # trigger_id doesn't always start from 0
             if self.previous_trigger_id is None:
-                # if no previous trigger exists, always trigger retraining
-                triggered = True
+                triggered = True  # if no previous trigger exists, always trigger retraining
+                drift_results: dict[str, MetricResult] = {}
```
MaxiBoether, line 98: This is the logic at the beginning, right? Should we have something like having the first trigger time or amount based? Right now not sure when exactly the first trigger would fire, maybe it would make sense to synchronize the first trigger to then have a reference point for drift?

modyn/supervisor/internal/triggers/drift/alibi_detector.py
```diff
@@ -0,0 +1,107 @@
...
+        embeddings_ref: pd.DataFrame | np.ndarray | torch.Tensor,
+        embeddings_cur: pd.DataFrame | np.ndarray | torch.Tensor,
+    ) -> dict[str, MetricResult]:
+        assert isinstance(embeddings_ref, np.ndarray)
+        assert isinstance(embeddings_cur, np.ndarray)
+
+        results: dict[str, MetricResult] = {}
+
+        for metric_ref, config in self.metrics_config.items():
+            metric = _evidently_metric_factory(config, embeddings_ref)
```
MaxiBoether, line 48: probably want to change the name here

modyn/supervisor/internal/triggers/drift/alibi_detector.py
```diff
@@ -0,0 +1,107 @@
...
+    def init_detector(self) -> None:
+        pass
+
+    # to do: reuse reference data, and the configured metrics (incl. kernels)
+
+    @override
+    def detect_drift(
+        self,
+        embeddings_ref: pd.DataFrame | np.ndarray | torch.Tensor,
+        embeddings_cur: pd.DataFrame | np.ndarray | torch.Tensor,
```
MaxiBoether, line 40: just for my understanding because we had this discussion before: do you want to run the detection on embeddings or the data directly or support both? (you mentioned alibi detect supports e.g. image input data)

modyn/tests/supervisor/internal/triggers/drift/test_evidently_detector.py
MaxiBoether: If it affects a single function, can we wrap evidently, e.g. having modyn.common.evidently or sth like that that imports * from evidently, and then overrides evidently.mmd = patched_mmd or something, and then we always import evidently from our package?

modyn/supervisor/internal/grpc_handler.py
```diff
@@ -163,6 +163,7 @@ def get_time_at_storage(self) -> int:
         if not self.connected_to_storage:
             raise ConnectionError("Tried to fetch data from storage, but no connection was made.")
 
+        # pylint: disable=c-extension-no-member
```
MaxiBoether, line 166: why this?

MaxiBoether (changes requested):
generally looks good, I have not commented on nitpicky stuff yet (mostly) since you mentioned you need to clean it still. my major comment would be whether instead of having an AlibiDetectMMDmetric and an evidentlyMMDMetric we can have an MMDMetric with AlibiDetect or Evidently as fields. Not sure if this is feasible (there might be major configuration differences in which they don't align, I guess then it does not make sense), let me know if this is a stupid idea

modyn/supervisor/internal/triggers/drift/alibi_detector.py
robinholzi: Currently, I'm only looking at embeddings in modyn. Raw data would probably need more architectural changes.

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: Good point, I just took the current solution so far, but that makes sense.

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
robinholzi: They support different features like hardware acceleration (incl. device selection),  kernel setup, thresholds and for most the configuration of the hypothesis tests (bootstrap vs. permutation: alibi detect allows to specify the num permutations, while evidently hard codes that). Also the quantile_probability sometimes seems to have an inverted definition (even within the same library, e.g. 0.95 instead of 0.05).

Overall I'd rather keep the abstractions close to their underlying variants.

MaxiBoether: Yeah, makes sense. Then let's keep it how you proposed it.

modyn/supervisor/internal/triggers/drift/alibi_detector.py
MaxiBoether: I see. Let's see whether we need it, I hope it will not be too big architectural changes (should be mostly changing the flow within the alibi detector since the data should be there anyways), depends on the quality of the results. As soon as the selection stuff is written, I can also dive into coding here and join forces.

modyn/supervisor/internal/grpc_handler.py
robinholzi: that's a question to ask pylint, I don't know😅

modyn/tests/supervisor/internal/triggers/drift/test_evidently_detector.py
robinholzi: I just figured that we don't even need to monkey patch anything, we can simply return from our custom adaption of the mmd wrapper class in the metric creation function (in `evidently_detector.py`)

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: Currently, it just takes the first batch acc. to `detection_interval_data_points` and train on it. That's fair for now, maybe we want sth. timestamp based in the future.

modyn/supervisor/internal/triggers/drift/alibi_detector.py
```diff
@@ -0,0 +1,107 @@
...
+        kernel = getattr(alibi_detect.utils.pytorch, config.kernel)
+
+    if isinstance(config, AlibiDetectMmdDriftMetric):
+        assert kernel is not None
+        return MMDDrift(
+            x_ref=embeddings_ref,
+            backend="pytorch",
+            p_val=config.p_val,
+            kernel=kernel,
+            n_permutations=config.num_permutations,
```
MaxiBoether, line 88: In your experiments, you let the pipelines run sometimes without the permutation testing. With this logic, do we always run the hypothesis tests? If so, we need to adjust the code to either support making threshold or hypothesis based decisions, right?

PR #552, opened 2024-06-27
Twin RHO Model Step 2: split the training set and train the twin model

XianzheMa:
should be reviewed after https://github.com/eth-easl/modyn/pull/547


MaxiBoether:
Will review when 547 is merged and the branch is updated

integrationtests/supervisor/integrationtest_supervisor.py
```diff
@@ -139,11 +139,6 @@ def test_rho_loss_pipeline_with_two_triggers() -> None:
         model_ids = [model.model_id for model in models]
         assert latest_il_model_id == max(model_ids)
 
-        # validate the holdout set is in correct proportion
-        # in rho_loss.yaml, each trigger contains 10 images. The holdout set ratio is 30%
-        # so the holdout set should contain 3 images
-        assert all(trigger.num_keys == 3 for trigger in triggers)
```
MaxiBoether, line 145: Why remove this check?

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -46,24 +46,80 @@ def __init__(
...
+            next_rho_trigger_id = rho_state[0] + 1
+            last_model_id = rho_state[1]
+        else:
+            logger.info(f"No previous state found for pipeline {self.rho_pipeline_id}.")
+            next_rho_trigger_id = 0
+            last_model_id = None
+        self._persist_holdout_set(query, next_rho_trigger_id, selector_storage_backend)
+        if self.il_training_config.use_previous_model:
+            if self.holdout_set_strategy == "Twin":
+                raise NotImplementedError("Use previous model currently is not supported for Twin strategy")
```
XianzheMa, line 64: It is implementable; however I need to deal with some nuances. therefore I want to leave it unimplemented in this PR as finetuning the IL model is not majorly used right now

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -46,24 +46,80 @@ def __init__(
     def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: AbstractStorageBackend) -> None:
         if not isinstance(selector_storage_backend, DatabaseStorageBackend):
             raise ValueError("RHOLossDownsamplingStrategy requires a DatabaseStorageBackend")
+
+        probability = self.holdout_set_ratio / 100
```
MaxiBoether, line 50: Ah, I think this should use the new ratio_max parameter as well, shouldn't it?

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -46,24 +46,80 @@ def __init__(
...
-        assert il_model_id is not None
-        config["il_model_id"] = il_model_id
+        state = self._get_latest_rho_state(self.rho_pipeline_id, self._modyn_config)
+        assert state is not None
+        config["il_model_id"] = state[1]
         return config
 
     @staticmethod
-    def _get_latest_il_model_id(rho_pipeline_id: int, modyn_config: dict) -> Optional[int]:
+    def _get_latest_rho_state(rho_pipeline_id: int, modyn_config: dict) -> Optional[Tuple[int, int]]:
```
MaxiBoether, line 120: Is this truly optional? doesn't look like it, seems like we assert is not None

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -154,24 +200,36 @@ def training_set_producer() -> Iterable[tuple[list[tuple[int, float]], dict[str,
...
+    ) -> Select:
+        assert isinstance(selector_storage_backend, DatabaseStorageBackend)
+
+        def _session_callback(session: Session) -> None:
+            session.query(SelectorStateMetadata).filter(
                 SelectorStateMetadata.pipeline_id == main_pipeline_id,
                 SelectorStateMetadata.seen_in_trigger_id == trigger_id,
-            )
-            .order_by(func.random())  # pylint: disable=E1102
-            .limit(target_size)
```
MaxiBoether, line 176: What is your reasoning behind changing the sampling logic? The reason we do it this way is because we want to get exactly `target_size` items. In your case, you'll get a different number of rows each time you sample. Is that a problem?

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -46,24 +46,80 @@ def __init__(
...
+        return stmt
 
     @property
     def downsampling_params(self) -> dict:
         config = super().downsampling_params
         config["rho_pipeline_id"] = self.rho_pipeline_id
-        il_model_id = self._get_latest_il_model_id(self.rho_pipeline_id, self._modyn_config)
-        assert il_model_id is not None
-        config["il_model_id"] = il_model_id
+        state = self._get_latest_rho_state(self.rho_pipeline_id, self._modyn_config)
```
MaxiBoether, line 114: instead of state maybe max_trigger_id, il_model_id = blabla?

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -46,24 +46,80 @@ def __init__(
     def inform_next_trigger(self, next_trigger_id: int, selector_storage_backend: AbstractStorageBackend) -> None:
         if not isinstance(selector_storage_backend, DatabaseStorageBackend):
             raise ValueError("RHOLossDownsamplingStrategy requires a DatabaseStorageBackend")
+
+        probability = self.holdout_set_ratio / 100
+
+        query = self._get_sampling_query(self._pipeline_id, next_trigger_id, probability, selector_storage_backend)
+        rho_state = self._get_latest_rho_state(self.rho_pipeline_id, self._modyn_config)
```
MaxiBoether, line 53: instead of state maybe max_trigger_id, il_model_id = blabla?

MaxiBoether (changes requested):
Thanks! No big comments, just a couple of nitpicky things

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
```diff
@@ -154,24 +200,36 @@ def training_set_producer() -> Iterable[tuple[list[tuple[int, float]], dict[str,
...
+            session.query(SelectorStateMetadata).filter(
                 SelectorStateMetadata.pipeline_id == main_pipeline_id,
                 SelectorStateMetadata.seen_in_trigger_id == trigger_id,
-            )
-            .order_by(func.random())  # pylint: disable=E1102
-            .limit(target_size)
+                func.random() < probability,  # pylint: disable=not-callable
+            ).update({"tmp_version": 1})
+            session.commit()
+
```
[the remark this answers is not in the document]
MaxiBoether: ah lol i mentioned it in my review which i apparently started too early. Yes, I agree, since it is an internal implementation detail of the rho loss strategy, it should be fine. I wasn't sure if you were aware of this change but then all good :)

integrationtests/supervisor/integrationtest_supervisor.py
XianzheMa: because now this size is probabilistically close to 3, but no longer accurate.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: sure will add it

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: At the first trigger, we don't have a latest trigger id and last model id, that's when None should be returned.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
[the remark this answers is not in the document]
XianzheMa: yeah. I also have a unit test for it with 1000 samples and ratio 50%. I assert the subset size to be between 450 and 550. So far it is not flaky😂

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: but the state could be None.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: However when querying `downsampling_params`, at that point (if it is queried at the right time) it is queried from trainer server, and a TSS is already generated. So in that method we assert it to be non-None.

In method `inform_next_trigger`, the state returned could be None.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: here the state could be None.

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: Here it shouldn't be None. But mypy will complain😂

MaxiBoether (approved):
Thanks, on the Optional typing thing I had another comment but feel free to ignore that, it's minor...

modyn/selector/internal/selector_strategies/downsampling_strategies/rho_loss_downsampling_strategy.py
XianzheMa: I agree with your thought on packing! I further considered that the downside of `Tuple[Optional[int], Optional[int]]` is that, it is possible to have inconsistent None-ness, i.e. the first field being None while the second is not, and I need to double assert. That's why I chose this way.

Issue #536, opened 2024-06-20
Speeding up evaluations 


MaxiBoether:
Right now, the supervisor sends one evaluation request per (model, interval) pair. However, often, we want to evaluate the same model on many intervals. Since every gRPC request implies that the evaluator downloads the model, instantiates the model on the GPU, creates processes etc, it might make sense to instead send requests of the form (model, [list of intervals]). Then, the evaluator can instantiate one model instance, only needs to spawn a single process etc.

This reduces the overhead of evaluation in cases where the evaluation time is dominated by the orchestration stuff (e.g. yearbook) instead of forward passes. Needs careful handling at the evaluator, since we now need to reset the state of the Pytorch Evaluator per interval, and need to return the numbers correctly, but is definitly notable and should be implementable.

PR #554, opened 2024-06-30
Batch evaluation intervals into a single request and a single evaluation process

XianzheMa:
Per the title, this PR enables batching evaluations on many intervals into a single evaluation request. This resolves https://github.com/eth-easl/modyn/issues/536. The integration test is adjusted to cover this new functionality.

## How to review

The best way to review is to first take a look at [modyn/protos/evaluator.proto](https://github.com/eth-easl/modyn/pull/554/files#diff-5017e582f2ffeabee6d3cfa42b2ca86f3f7035419df31cc3cff68be568279f04) to see on the API level what is changed.

## Miscellaneous
After this PR, batching is only enabled on server side, on client side still one interval is passed at a time, as I am confused on the current way to generate `EvalRequest`.

After this PR, @robinholzi could you make a PR to collect the intervals associated with the same `id_model` and `dataset_id`, and pack them in one evaluation round? I think just looking at the function `_single_batched_evaluation` in this file [modyn/supervisor/internal/pipeline_executor/evaluation_executor.py](https://github.com/eth-easl/modyn/pull/554/files#diff-f111ee1d407da6a1b5d46b118dd5221da51801ead85ea2c9271b008299c6a11f) should be enough to understand the change / how to make that PR. It should be a very easy and straightforward PR. I am just confused with the data models there. Thank you so much!


MaxiBoether:
Thank you so much! I will go through this tomorrow and also ping @robinholzi to review this since it affects his evaluation stuff

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -125,31 +126,34 @@ def evaluate_model(self, request: EvaluateModelRequest, context: grpc.ServicerCo
...
-        dataset_size = dataset_size_response.num_keys
-        if not dataset_size_response.success:
-            logger.error(
-                f"Total number of keys for dataset {dataset_size_req.dataset_id} cannot be fetched. "
-                f"Evaluation cannot be started."
-            )
-            return EvaluateModelResponse(
-                evaluation_started=False, eval_aborted_reason=EvaluationAbortedReason.DATASET_NOT_FOUND
+        data_sizes: list[int] = []
+        for interval in request.dataset_info.evaluation_intervals:
```
MaxiBoether, line 130: Before I start reviewing, could you clarify what you mean by batching is done at the server? I am a little bit confused. I thought the client needs to group all eval requests by model id, and then submit a single request with all intervals for that model. How can you do this at the server? And this also looks like the server iterates over the intervals. I am a bit confused

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -134,17 +129,14 @@ def run_pipeline_evaluations(
             trigger_id: The trigger id to evaluate.
             training_id: The training id to evaluate.
             model_id: The model id to evaluate.
-            first_timestamp: Start of the interval to evaluate.
-            last_timestamp: End of the interval to evaluate.
+            first_timestamp: Start of the training interval.
```
MaxiBoether, line 132: I have just started to scroll through the PR but given the confusion I would like to clarify this change. These intervals have nothing to do with training, right @robinholzi ? They describe how the eval dataset is sliced. Hence I am not sure if there is a misunderstanding here

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
```diff
@@ -52,6 +52,13 @@ def get_name(self) -> str:
         """
         raise NotImplementedError()
 
+    @abstractmethod
```
robinholzi, line 55: I find this a bit odd, I'd rather create separate Metric objects if they use independent states.

modyn/protos/evaluator.proto
```diff
@@ -54,25 +58,25 @@ message EvaluationStatusRequest { int32 evaluation_id = 1; }
 message EvaluationStatusResponse {
   bool valid = 1;
   bool is_running = 2;
-  bool state_available = 3;
-  bool blocked = 4;
-  optional string exception = 5;
```
robinholzi, line 59: Why do we need to remove these logging fields completely? At least `batches_seen` and `samples_seen` are quite nice to verify that evaluations actually contained the expected amount of data.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: In the Slicing strategy e.g. these intervals are completely independent from training intervals (unless the triggers are configured to follow the same schedule ;) ). I think we should keep the old description.

modyn/supervisor/internal/pipeline_executor/pipeline_executor.py
```diff
@@ -47,11 +40,6 @@
 logger = logging.getLogger(__name__)
 EXCEPTION_EXITCODE = 8
 
-EVAL_RESULT_WRITER_CLASSES: dict[ResultWriterType, type[AbstractEvaluationResultWriter]] = {
```
robinholzi, line 50: Good, I think we can strip these log writers

robinholzi (changes requested):
Thanks so much @XianzheMa for putting some thought into the evaluator side of the evaluation logic! I can surely adjust the batching in supervisor to comply with the new interface! Just a couple of questions before we finalise the grpc interface changes

modyn/protos/evaluator.proto
MaxiBoether: Xianzhe asked whether we can remove the tracking but I thought we were talking about the visual progress bar. The amount of data in each interval we definitely need to track, this is useful debugging information

MaxiBoether, line 59: This is also problematic. We need to check for Exceptions at the evaluator and print them. I thought there was a problem with the visualization of the progress bars which I am fine to delete but I did not want to remove all status message transfer between the evaluator and the supervisor

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -125,31 +126,34 @@ def evaluate_model(self, request: EvaluateModelRequest, context: grpc.ServicerCo
...
+                )
+                return EvaluateModelResponse(
+                    evaluation_started=False, eval_aborted_reason=EvaluationAbortedReason.DATASET_NOT_FOUND
+                )
+
+            if dataset_size == 0:
+                logger.info(
+                    f"Dataset {dataset_size_req.dataset_id} is empty in given time interval. "
+                    f"Evaluation cannot be started."
+                )
```
MaxiBoether, line 152: What happens here with the return if only one out of several intervals is empty? I am a bit unsure whether we can safely return here.

MaxiBoether: I guess we might need to have the eval_aborted_reason= / evaluation_started= per interval, since different intervals might fail/work

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
MaxiBoether: Yes, I'd also say we should instantiate new objects

MaxiBoether (changes requested):
Thank you so much Xianzhe! I hope the changes are not too much work. We can chat with Robin about the status queue thingy, but I think we need a way of e.g. transferring exceptions to the supervisor because we should have those in the logs (I've used them before for debugging, helpful to have it in a single place). I think there are some minor issues in the interface where we need to adjust the evaluation started request, but the overall flow looks nice!

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: but why within the method we pass them to an arg of `training_interval`?
https://github.com/eth-easl/modyn/blob/60d3baa7523d21197dec85262fb8c31662d656c2/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py#L159

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
XianzheMa: Thanks. Will do.

modyn/protos/evaluator.proto
XianzheMa: yeah these few fields are useful visually, to report the progress of the evaluation while it is ongoing. However most evaluations are quite fast (i.e. only on a short interval and one epoch), this is not so useful.

Also when now we batch the evaluation requests, it is more tricky to keep these fields correct. To balance the usefulness and efforts, I decided to remove them.

To verify the evaluation actually contains expected data, in `EvaluationModelResponse` there is a field `dataset_sizes`. To verify the progress is going correctly, we still have the log https://github.com/eth-easl/modyn/blob/60d3baa7523d21197dec85262fb8c31662d656c2/modyn/evaluator/internal/pytorch_evaluator.py#L168.

modyn/tests/supervisor/internal/test_grpc_handler.py
```diff
@@ -349,135 +365,67 @@ def test_wait_for_evaluation_completion(*args):
     handler = GRPCHandler(get_simple_config())
     handler.init_cluster_connection()
     assert handler.evaluator is not None
-    eval_status_queue = mp.Queue()
-    evaluations = {
-        1: EvaluationStatusReporter(
```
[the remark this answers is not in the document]
XianzheMa: Since I removed those fields `num_samples_seen`, etc. (the number to report the internal progress of evaluation on one interval), this queue is no longer used.

modyn/protos/evaluator.proto
```diff
@@ -54,25 +58,25 @@ message EvaluationStatusRequest { int32 evaluation_id = 1; }
...
+  optional string exception = 3;
 }
 
-message EvaluationData {
+message SingleMetricResult {
   string metric = 1;
   float result = 2;
 }
 
+message SingleEvaluationData {
```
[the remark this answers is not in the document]
XianzheMa: I admit that the logic here is confusing, I will add a comment.
So now the evaluation is batched. The evaluation results, first, on the top level, is a list where each entry corresponds to the each interval. Then for the evaluation results for one interval, we have multiple numbers for multiple metrics.

modyn/protos/evaluator.proto
robinholzi: Hmm, stdout logs are virtually useless when running big experiments with a lot of evaluations and you don't have an initial idea that sth. would be wrong.
Currently, we are using this information in the analytics tool to e.g. check that certain evaluation strategy configurations did what they are supposed to do.

modyn/protos/evaluator.proto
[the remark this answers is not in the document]
robinholzi: ahh, makes sense, thanks!

modyn/protos/evaluator.proto
XianzheMa: I see. Can you point me where in the code how any field is used? I am happy to add them back and implement correctly, just to understand their usage.

XianzheMa: anyway, I will add back `batches_seen` and `samples_seen`, together with a new field `current_interval_idx`

XianzheMa: anyway, I will add back `batches_seen` and `samples_seen`, together with a new field `current_interval_idx`

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
MaxiBoether: Unfortunately, I have to disagree here. There are strategies that generate intervals in a fixed distance, e.g. [0,1000] and [1000,2000] etc. For example when running a regular CGLM pipeline, this will occur.

However, I am not sure we need to have a lot of engineering effort here. I see two solutions

1. Include the interval or some interval id that the supervisor generates in the metric reply. Then you immediately have a 1:1 mapping and it should be easy to pass-through

2. instead of failing here, return None / 0 for the metrics and just not start an evaluation. i am not sure if this has side effects (@robinholzi) but might be a quick fix.

However, I do think we need a 1:1 mapping since exceptions can occur. Exceptions sometimes happen (e.g., some random error while fetching data) and then the ordering may be disturbed I think (although it affects all intervals in the current batch of intervals that have not been processed). Not sure how you'd handle those right now?

modyn/protos/evaluator.proto
robinholzi: I just checked I think we only use `dataset_size` ([here](https://github.com/eth-easl/modyn/blob/60d3baa7523d21197dec85262fb8c31662d656c2/modyn/supervisor/internal/pipeline_executor/models.py#L318C1-L319C1)). Sorry for the confusion, I don't know what the semantics of these two fields is then. If you don't find them necessary I'm happy to drop them

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
robinholzi: We definitely need this 1:1 mapping, often times, the test datasets don't have enough samples at the beginning yielding empty evaluations. With batched processing, we wouldn't evaluate anything for one model just because the first evaluation interval has no samples. The same occurs when using a Slicing / periodic strategy that generates smaller intervals (more likely to have no test data).
I think returning a list of EvaluationResults where every EvaluationResults uses (interval_start, interval_end) as unique identifier (for a certain model / dataset combination) is desirable. (@MaxiBoether's option 1)

modyn/protos/evaluator.proto
XianzheMa: okay great! then I will just ask Maxi. Thanks for confirming this.

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
XianzheMa: Thanks a lot, Robin and Maxi for providing extra insight! Then I guess I will use simply the interval_idx, i.e. the index in which the interval is present in the interval list, as the index. I think it is lighter as (interval_start, Interval_end).

For other transient failures during evaluation that manifest themselves as exceptions, currently when one happens, I just fail all requests in this batch, as I think there is something fundamentally wrong. but what do you suggest? @MaxiBoether

MaxiBoether: It depends. If that is the easiest solution that's fine (we don't have a restart logic for failed evaluations anyways currently). Generally, what will happen is that at some point during the evaluation we see an exception and all intervals after the current ones also fail. The intervals for which we've already collected results we could theoretically still transfer the metrics. The most common failure I've seen (not very often, but actually the only one) is the connection between evaluator <-> storage just terminating for whatever reason, so something that did not really affect the numbers for the previous intervals that still worked. However, we can play it safe and just fail that batch

XianzheMa: Okay I made the logic dependent on the client: when an exception is met, is up to the client to decide whether to retrieve results prior to the exception. The server is ready to provide them

modyn/evaluator/internal/metrics/abstract_evaluation_metric.py
XianzheMa: done

modyn/protos/evaluator.proto
```diff
@@ -43,36 +47,42 @@ message EvaluateModelRequest {
...
 message EvaluateModelResponse {
+  // only when all interval evaluations failed, this field will be set to false
   bool evaluation_started = 1;
   int32 evaluation_id = 2;
-  int64 dataset_size = 3;
-  EvaluationAbortedReason eval_aborted_reason = 4;
+  // non-empty only when evaluation_started is true; in this case it has the same size as the number of intervals
+  repeated int64 dataset_sizes = 3;
+  // always has the same size as the number of intervals
+  repeated EvaluationAbortedReason eval_aborted_reasons = 4;
```
robinholzi, line 56: If I understood correctly, an evaluation for a certain interval can fail because of 
1) a problem with the interval (e.g. no data in that interval)
2) a problem with whole request (applies to all intervals in the request)

If that is the case I think the cleanest solution would be to have an error response field on request level and an error field on interval level.

That goes in hand with the `repeated int64 dataset_sizes`. Currently, we can only match the `data_sizes` to the error codes and intervals by assuming consistent indexes in the sequences.

I recommend to add a `repeated EvaluateModelIntervalResponse` to the `EvaluateModelResponse` and add all interval-specific things (dataset_sizes, interval failure, ...) there.

modyn/protos/evaluator.proto
```diff
@@ -43,36 +47,42 @@ message EvaluateModelRequest {
...
+  optional string exception = 3;
 }
 
-message EvaluationData {
+message SingleMetricResult {
   string metric = 1;
   float result = 2;
 }
 
+message SingleEvaluationData {
```
robinholzi, line 72: maybe `EvaluationIntervalData`?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: Yes my question is we pass these two parameters to the `training_interval` of this function, so their description in the docstring should be "training interval" right? Or I mistook anything?

modyn/protos/evaluator.proto
XianzheMa: make sense. will do

modyn/protos/evaluator.proto
XianzheMa: done

modyn/protos/evaluator.proto
XianzheMa: added; make sense thanks

modyn/tests/evaluator/internal/test_pytorch_evaluator.py
```diff
@@ -184,108 +171,93 @@ def test_load_model():
...
-    status = response_queue.get()
-    assert status["num_batches"] == 0
-    assert status["num_samples"] == 0
-
-    # accuracy
-    metric_name, metric_result = metric_result_queue.get()
-    assert metric_name == Accuracy(AccuracyMetricConfig()).get_name()
-    assert metric_result == pytest.approx(1)
+    expected_accuracies = [1.0, 0.01, 0.0, 0.5]
+    expected_f1scores = [0.5, ANY, 0.0, 1 / 3]
```
XianzheMa, line 217: the second f1 score is a number not so easily computed from outside. I just trust the code

XianzheMa:
The integration test definitely passed on our server. Will check tomorrow why

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: Reading the code again, this seems indeed to be the interval of training for this model, so your updated description might be correct. @robinholzi WDYT?

MaxiBoether:
> The integration test definitely passed on our server. Will check tomorrow why

I think there probably is some flakyness somewhere. Sorry, I am too tired to review this properly tonight, I will do it tomorrow morning as well. But I don't think it's because of release/debug/... mode, I think this is rather some concurrency issue somewhere

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -167,45 +188,35 @@ def evaluate_model(self, request: EvaluateModelRequest, context: grpc.ServicerCo
...
-        for configuration_json in metric_configurations:
-            metric = MetricFactory.get_evaluation_metric(configuration_json)
-            if metric.get_name() not in metric_names:
-                metrics.append(metric)
-                metric_names.add(metric.get_name())
-            else:
-                logger.warning(f"Metric {metric.get_name()} is already registered.")
-        return metrics
+        return EvaluateModelResponse(
+            evaluation_started=True,
```
MaxiBoether, line 213: What is the purpose of the  `evaluation_started` field as we now have the responses per interval? As far as I can tell, it's true if at least one interval has been started. Do we still need this field?

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
```diff
@@ -301,20 +282,26 @@ def get_evaluation_result(
...
 
-        for name, result in metric_results:
-            evaluation_data.append(EvaluationData(metric=name, result=result))
-        return EvaluationResultResponse(valid=True, evaluation_data=evaluation_data)
+            evaluation_data.append(single_eval_data)
+        if len(evaluation_data) < len(self._evaluation_dict[evaluation_id].not_failed_interval_ids):
+            logger.error(
+                f"Could not retrieve results for all intervals of evaluation {evaluation_id}. "
+                f"Expected {len(self._evaluation_dict[evaluation_id].not_failed_interval_ids)}, "
+                f"but got {len(evaluation_data)}. Maybe exception happens during evaluation."
```
MaxiBoether, line 302:
```suggestion
                f"but got {len(evaluation_data)}. Maybe an exception happened during evaluation."
```

modyn/evaluator/internal/pytorch_evaluator.py
```diff
@@ -142,38 +128,42 @@ def evaluate(self) -> None:
...
-                    metric.evaluate_dataset(y_true, y_score, self._num_samples)
+                    metric.evaluate_dataset(y_true, y_score, num_samples)
 
-        # We do this since we might also have just non-holistic metrics, in which case len(y_true) always is 0
-        for metric in self._metrics:
-            self._metric_result_queue.put((metric.get_name(), metric.get_evaluation_result()))
+        metric_result = []
+        for metric in metrics:
+            metric_result.append((metric.get_name(), metric.get_evaluation_result()))
+        self._metric_result_queue.put((interval_idx, metric_result))
```
MaxiBoether, line 153: Just for my understanding: The interval idx stems from the index in the list of indices that the supervisor sends, correct? So the index is implicltly given by the supervisor, and the interval is not identified by the [start, end] timestamps, correct?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -292,40 +295,31 @@ def _single_evaluation(self, log: StageLog, eval_status_queue: Queue, eval_req:
...
+        self.grpc.wait_for_evaluation_completion(response.evaluation_id)
+        eval_data = self.grpc.get_evaluation_results(response.evaluation_id)
+        self.grpc.cleanup_evaluations([response.evaluation_id])
+        # here we assume only one interval is evaluated
+        eval_results: dict = {"dataset_size": response.interval_responses[0].dataset_size, "metrics": []}
+
+        for single_eval_data in eval_data:
+            if single_eval_data.interval_index == 0:
+                for metric in single_eval_data.evaluation_data:
+                    eval_results["metrics"].append({"name": metric.metric, "result": metric.result})
```
MaxiBoether, line 319: I don't understand the if == 0 yet. Why is that valid?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -292,40 +295,31 @@ def _single_evaluation(self, log: StageLog, eval_status_queue: Queue, eval_req:
...
         if not response.evaluation_started:
-            log.info.failure_reason = EvaluationAbortedReason.DESCRIPTOR.values_by_number[
-                response.eval_aborted_reason
-            ].name
+            failure_reason = get_failure_reason(response.interval_responses[0].eval_aborted_reason)
             logger.error(
-                f"Evaluation for model {eval_req.id_model} on split {eval_req.interval_start} to "
-                f"{eval_req.interval_end} not started with reason: {log.info.failure_reason}."
+                f"Evaluation for model {model_id_to_eval} on split {interval_start} to "
+                f"{interval_end} not started with reason: {failure_reason}."
```
MaxiBoether, line 305: Not sure this log is correct. It's a bit confusing but I think those are the interval boundaries of the model (where it was trained), not of the interval on which we evaluate. @robinholzi ?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -292,40 +295,31 @@ def _single_evaluation(self, log: StageLog, eval_status_queue: Queue, eval_req:
...
-            else {}
-        )
+        logger.info(f"Evaluation started for model {model_id_to_eval} on split {interval_start} " f"to {interval_end}.")
+        self.grpc.wait_for_evaluation_completion(response.evaluation_id)
+        eval_data = self.grpc.get_evaluation_results(response.evaluation_id)
+        self.grpc.cleanup_evaluations([response.evaluation_id])
+        # here we assume only one interval is evaluated
+        eval_results: dict = {"dataset_size": response.interval_responses[0].dataset_size, "metrics": []}
+
+        for single_eval_data in eval_data:
```
MaxiBoether, line 316: I think related to my other question on the interval_index I have a bigger question: It does not appear that we have easy access to the interval index here. Basically there is a 1:1 mapping between interval index and the interval on which we evaluate. Instead of somehow re-creating the mapping at the supervisor, it might be a lot easier to use a tuple (eval_interval_start, eval_interval_end) at the evaluator as the identifier of an interval, no? Because then we can just return that in the results instead of this index which we don't easily have access to. WDYT? If it's unclear, I will be at the office in ~1 hour so we can chat about it

MaxiBoether (changes requested):
Thank you! My only big comment besides nitpicky stuff is on the interval_idx. I think we should switch to using the interval boundaries cause looking at the supervisor, it would simplify the logic for the follow up PR if we directly have the interval bounds instead of some idx. I did not immediately find the reason for the integrationtest failure, though :(

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: I realized it's probably due to the fact we only have a single interval, but then I am not sure why we have an if instead of just asserting it's 0 since otherwise we have a problem in the current implementation :D

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
XianzheMa: in that sense we don't need it. It is just a slight ease of life so that the client can easily decide if I want to wait for the evaluation completion, or I can just abort, without going through each interval response. what do you think?

modyn/evaluator/internal/pytorch_evaluator.py
XianzheMa: yes. My thought is that [start, end] is somewhat complex, and also not unique in the semantics:
[0, 2000] and [None, 2000] means the same interval.

Also on the supervisor side, when collecting the evaluation results, it is easier to just index into the interval lists, rather than going through the list and matching the `[start, end]` pattern. What do you think?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: you are right. I was just being lazy in the sense that the result collection logic needs to be rewritten by @robinholzi anyway. But I will change it as you suggest!

modyn/evaluator/internal/pytorch_evaluator.py
MaxiBoether: Maybe i am missing something but i dont see where at the supervisor we currently have this mapping. I guess my concern mostly is that we run into a discrepancy of indices between evaluator and supervisor. But ok if its very easy to have this mapping at the supervisor as well, then we can keep your approach

modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py
MaxiBoether: Ok makes sense :) maybe we can add a brief comment to the proto file that this is a field for convenience?

XianzheMa: will do!

modyn/evaluator/internal/pytorch_evaluator.py
XianzheMa: no the indices between the evaluator and supervisor are consistent.
The supervisor naturally have a list of intervals, because it is the supervisor that put this interval list into the evaluation request and send the request.

How the indices are tracked on evaluator:
1. On `evaluator_grpc_servicer.py`, we enumerate over the original interval list and record the ids that didn't fail. https://github.com/eth-easl/modyn/blob/94efd22fb40dbdaf9c7adc5be2c63e3cb4f680c8/modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py#L146
2. This `not_failed_interval_ids` is passed to the `EvaluationInfo` together with the original full interval list https://github.com/eth-easl/modyn/blob/94efd22fb40dbdaf9c7adc5be2c63e3cb4f680c8/modyn/evaluator/internal/grpc/evaluator_grpc_servicer.py#L197.
3. With these two fields, the `pytorch_evaluator` can only execute the not failed intervals and associate them with the right ids in the `metric_result_queue` https://github.com/eth-easl/modyn/blob/94efd22fb40dbdaf9c7adc5be2c63e3cb4f680c8/modyn/evaluator/internal/pytorch_evaluator.py#L157

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: I think it is correct (at least it is the same interval as before this PR). We use the fields from the `EvalRequest` to launch the request. 
https://github.com/eth-easl/modyn/blob/94efd22fb40dbdaf9c7adc5be2c63e3cb4f680c8/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py#L242

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: I see your point. My thought is that, later, this function will be rewritten so that it takes a list of intervals, packs them inside the `EvaluateModelRequest`, and then contact the evaluator to launch evaluation. So this function always has the full list of intervals when reading the response from the evaluator.

That's why I feel it is more convenient to give an index so here this `single_batched_evaluation` function can directly index to the original interval list. @robinholzi what's your thought here?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: done

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
XianzheMa: Also, it is possible that (for whatever reason) the client sends a request with duplicate intervals inside, e.g. two [100, 200], in this case we cannot distinguish based on the interval itself.

XianzheMa:
> > The integration test definitely passed on our server. Will check tomorrow why
> 
> I think there probably is some flakyness somewhere. Sorry, I am too tired to review this properly tonight, I will do it tomorrow morning as well. But I don't think it's because of release/debug/... mode, I think this is rather some concurrency issue somewhere

everything should be fixed now. Let's see

MaxiBoether (approved):
LGTM! Let's wait for @robinholzi's review (and clarification regarding the docstring), but I do not have further comments. Thank you so much!

modyn/evaluator/internal/pytorch_evaluator.py
robinholzi: Can we let the supervisor encode the indices explicitly in the payload (not only by the position in an array)? Then the evaluator just refers to the id it was given from the payload if we build indices with enumeration no both sides that sounds like a potential pitfall where things start to fail silently, just because someone made wrong assumptions about interval ordering.

robinholzi (approved):

XianzheMa:
The last commit only involves a comment, and before the commit the CI passed. So I just merged without waiting more.

PR #555, opened 2024-07-03
feat: Create `DriftDetector` abstraction and introduce alibi-detect

robinholzi:
# Motivation

We want to be able to use both evidently and alibi-detect as they support different drift detection metrics as well as differ in their runtime performance.

For that we need a clean abstraction that allows us to delegate drift detection to the respective library

# Changes

- add `alibi-detect` dependency
- adjust drift detection abstraction via the `DriftDetector` interface
- introduce pydantic configurations models for alibi detect and evidently
- allow the collection and aggregation of multiple drift metrics.
- add tests for our embedding drift detection setup

# Note:

As this is rather big already, I'd like to get this merged before trying to run drift detection pipelines, so we might need slight adjustments in a followup PR.


experiments/arxiv/compare_trigger_policies/run.py
```diff
@@ -11,85 +12,145 @@
...
+        gpu_device="cuda:1",
+        warmup_until=_FIRST_TIMESTAMP,
+    ),
+    # cold training startup vs warmup phase
+    1: Experiment(
+        # to verify online composite model determination logic
+        name="arxiv-timetrigger-warm-start",
+        eval_handlers=[
+            construct_slicing_eval_handler("90d", int(pd.to_datetime("2000-01-01").timestamp())),
+            # construct_between_trigger_eval_handler()  # TODO: reenable for arxiv_kaggle_all
```
robinholzi, line 131: let's leave the todos in here, it's just an experiment config

analytics/app/pages/plots/eval_heatmap.py
```diff
@@ -86,7 +86,7 @@ def gen_figure(
         label_map = {v: full_refs[k] for k, v in mapping[()].items()}
```
MaxiBoether, line 87: For the changes in eval plotting I trust that this makes sense :D

modyn/config/__init__.py
```diff
@@ -12,7 +12,7 @@ def read_modyn_config(path: Path) -> ModynConfig:
     Args:
         path: Path to the configuration file.
 
-    Returns:
+    Returns: e
```
MaxiBoether, line 15: typo?

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class _AlibiDetectBaseDriftMetric(ModynBaseModel):
+    p_val: float = Field(0.05, description="The p-value threshold for the drift detection.")
+
+    # to be added in the future:
+
+    # x_ref_preprocessed
+    # preprocess_at_init
+    # configure_kernel_from_x_ref
```
MaxiBoether, line 14: Do we need these for the drift experiments with Modyn? Sorry, not entirely sure what parameters are required.

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class AlibiDetectMmdDriftMetric(_AlibiDetectBaseDriftMetric, AlibiDetectDeviceMixin):
+    id: Literal["AlibiDetectMmdDriftMetric"] = Field("AlibiDetectMmdDriftMetric")
+    num_permutations: int = Field(
+        100,
+        description=(
+            "The number of permutations used in the MMD hypothesis permutation test. 100 is the alibi-detect default."
+        ),
+    )
```
MaxiBoether, line 40: What if we don't want to do hypothesis testing and just specify a threshold? Is that supported currently?

modyn/supervisor/internal/triggers/drift/drift_detector.py
```diff
@@ -0,0 +1,24 @@
+from abc import ABC
+
+import numpy as np
+import pandas as pd
+import torch
+from modyn.config.schema.pipeline.trigger.drift.config import DriftMetric
+from modyn.config.schema.pipeline.trigger.drift.result import MetricResult
+
+
+class DriftDetector(ABC):
```
MaxiBoether, line 10: Is this for your custom MMD implementation?

MaxiBoether (approved):
Mostly looks good, I guess my main question is on running hypothesis tests. I suppose we should do the windowing logic in a follow up PR?

modyn/supervisor/internal/triggers/drift/drift_detector.py
robinholzi: No, it's just the interface that abstracts from alibi-detect and evidently. I removed my custom implementation for now as we don't need it.

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
robinholzi: alibi detect needs to run permutation testing. When I referred to disabling it, I just defaulted it to 1 permutation. I will make this field nullable and default it to 1 in the business logic.

modyn/supervisor/internal/triggers/drift/drift_detector.py
MaxiBoether: ah sorry, i did not realize it's the abstract base class

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
robinholzi: I tried to tweak them, they seem not to have big effects. I have already added them to the interface locally so we can set them if we like. For the paper they won't be of any use though.

modyn/config/__init__.py
robinholzi: Yes, thanks!

modyn/config/schema/pipeline/trigger/drift/alibi_detect.py
MaxiBoether: Sounds like a good plan

MaxiBoether (approved):

MaxiBoether:
changes LGTM

modyn/config/examples/modyn_config.yaml
```diff
@@ -67,7 +70,7 @@ storage:
         selector_batch_size: 256,
       },
       {
-        name: "yearbook-test",
+        name: "yearbook_test",
```
XianzheMa, line 73: @robinholzi let's try to make as few breaking changes as possible until the paper submission, because on the sigmod branch there are scripts with hard-coded dataset names. Thanks! 

The scripts won't run properly without a name change. CC @MaxiBoether 

PS: not related to this PR, just want to bring it to attention that we shouldn't change the default value in Pydantic configs in following PRs until submission, because otherwise the scripts on sigmod branch will run suddently with different configs without notice.

PR #558, opened 2024-07-03
Add rho training log to the pipeline file

XianzheMa:
Per the title, this PR actually adds the log of downsampling phase at selector to the selector log, hence further to the pipeline log. The intention here is to see how well the il model is trained with the `record_loss_every`.

### DIFF
--- a/modyn/config/schema/pipeline/sampling/downsampling_config.py
+++ b/modyn/config/schema/pipeline/sampling/downsampling_config.py
@@ -147,6 +147,9 @@ class ILTrainingConfig(TrainingConfig):
     initial_model: Literal["random"] = "random"
     initial_model_id: Literal[None] = None
     checkpointing: NoCheckpointingConfig = Field(default=NoCheckpointingConfig())
+    drop_last_batch: bool = Field(
+        default=False, description="Whether to drop the last batch if it is smaller than the batch size."
+    )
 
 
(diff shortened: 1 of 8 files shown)


modyn/config/schema/pipeline/sampling/downsampling_config.py
```diff
@@ -147,6 +147,9 @@ class ILTrainingConfig(TrainingConfig):
     initial_model: Literal["random"] = "random"
     initial_model_id: Literal[None] = None
     checkpointing: NoCheckpointingConfig = Field(default=NoCheckpointingConfig())
+    drop_last_batch: bool = Field(
+        default=False, description="Whether to drop the last batch if it is smaller than the batch size."
+    )
```
XianzheMa, line 152: The default value for the normal training config is `True` because we want to drop the last batch.

Here this is for the training for an il model (which is used to guide the data selection for the main model). We want to disable last batch dropping because we don't use selection strategies on il model so the previous exception triggered by not dropping last batch is avoided. Also we want to use as much data s possible.

MaxiBoether (approved):
LGTM! (when the unit tests are fixed)

robinholzi (approved):

XianzheMa:
The unit test failed last time last night was due to flakiness, not something really wrong :)

PR #559, opened 2024-07-03
feat!: Batch evaluations in evaluator

robinholzi:
# Motivation

We want to utilize the batching logic introduced in https://github.com/eth-easl/modyn/pull/554 by grouping requests in the supervisor.

This will break the analytics tool because of the adjusted logging. Once we have a completed run I'll open a followup PR.

### DIFF
--- a/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
+++ b/modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 import pickle
 import sys
+from collections import defaultdict
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from functools import partial
@@ -229,10 +230,10 @@ class EvaluationExecutor:
         logs = SupervisorLogs()
 
         # do batching by (dataset_id, model_id)
-        eval_requests_by_model: dict[tuple[str, int], list[EvalRequest]] = {}
+        eval_requests_by_model: dict[tuple[str, int], list[EvalRequest]] = defaultdict(list)
         for eval_req in eval_requests:
             key_ = (eval_req.dataset_id, eval_req.id_model)
-            eval_requests_by_model[key_] = eval_requests_by_model.get(key_, []) + [eval_req]
+            eval_requests_by_model[key_] += [eval_req]
 
         def worker_func(model_eval_req: tuple[tuple[str, int], list[EvalRequest]]) -> StageLog:
             """
@@ -354,13 +355,25 @@ class EvaluationExecutor:
         eval_data = self.grpc.get_evaluation_results(response.evaluation_id)
         self.grpc.cleanup_evaluations([response.evaluation_id])
 
-        eval_results: list[tuple[str | None, dict[str, Any]]] = []
         assert len(eval_data) == len(intervals), f"We expected {len(intervals)} intervals, but got {len(eval_data)}."
         assert eval_data
 
-        # as the dataset_size comes from the `EvaluateModelResponse` and response is a dense list (no intervals are
-        # skipped), we build a list of results with the same order as the intervals.
-        # the metrics will be filled in the next loop that unwraps `EvaluationResultResponse`.
+        eval_results: list[tuple[str | None, dict[str, Any]]] = []
+
+        # ---------------------------------------------- Result Builder ---------------------------------------------- #
+        # The `eval_results` list is a list of tuples. Each tuple contains a failure reason (if any) and a dictionary
+        # with the evaluation results. The order of the tuples corresponds to the order of the intervals.
+        #
+        # response.interval_responses contains the evaluation results for each interval in the same order as the
+        # intervals in the request. Failed evaluations are marked with a failure reason.
+
+        # Metric results come from the `EvaluateModelResponse` and are stored in the `evaluation_data` field. This
+        # only contains the metrics for the intervals that were successfully evaluated.
+        #
+        # Therefore we first build a list of results with the same order as the intervals. The metrics will be filled in
+        # the next loop that unwraps `EvaluationResultResponse`.
+        # ----------------------------------------------------- . ---------------------------------------------------- #
+
         for interval_response in response.interval_responses:
             if interval_response.eval_aborted_reason != EvaluationAbortedReason.NOT_ABORTED:
                 reason = get_failure_reason(interval_response.eval_aborted_reason)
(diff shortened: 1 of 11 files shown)


MaxiBoether:
Lmk when i should review this

XianzheMa:
I think it should be called "batching requests" because we do not actually parallelize them?

robinholzi:
> I think it should be called "batching requests" because we do not actually parallelize them?

Ah right, the evaluator doesn't do parallelization (yet)

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -226,33 +228,61 @@ def _launch_evaluations_async(
         tasks: list[Future[StageLog]] = []
         logs = SupervisorLogs()
 
-        def worker_func(eval_req: EvalRequest) -> StageLog:
+        # do batching by (dataset_id, model_id)
+        eval_requests_by_model: dict[tuple[str, int], list[EvalRequest]] = {}
+        for eval_req in eval_requests:
+            key_ = (eval_req.dataset_id, eval_req.id_model)
+            eval_requests_by_model[key_] = eval_requests_by_model.get(key_, []) + [eval_req]
```
MaxiBoether, line 235: nit but if you'd want to you could use a defaultdict here and just .append() instead of having the get logic

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -264,24 +294,28 @@ def worker_func(eval_req: EvalRequest) -> StageLog:
...
+            f"Evaluation Starts for model {model_id_to_eval} and dataset {dataset_id} on intervals:  {intervals}."
         )
         dataset_config = next((d for d in self.pipeline.evaluation.datasets if d.dataset_id == dataset_id))
 
         request = GRPCHandler.prepare_evaluation_request(
             dataset_config.model_dump(by_alias=True),
             model_id_to_eval,
             self.pipeline.evaluation.device,
-            intervals=[(interval_start, interval_end)],
+            intervals=cast(list[tuple[int | None, int | None]], intervals),
```
MaxiBoether, line 318: why is this cast necessary? isn't this the input type of `intervals`?

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -295,30 +329,52 @@ def _single_batched_evaluation(
...
+        for interval_response in response.interval_responses:
+            if interval_response.eval_aborted_reason != EvaluationAbortedReason.NOT_ABORTED:
+                reason = get_failure_reason(interval_response.eval_aborted_reason)
+                eval_results.append((reason, {}))
+            else:
+                eval_results.append((None, {"dataset_size": interval_response.dataset_size, "metrics": []}))
+
+        for interval_result in eval_data:
+            interval_idx = interval_result.interval_index
+            assert eval_results[interval_idx][0] is None, "Evaluation failed, no metrics should be present."
```
MaxiBoether, line 373: I'm sure there is a logic here but I don't understand. We are asserting for every interval that the failure reason is None. But what about the failed evaluations? Then this assertion will fail, won't it? Should we really crash then, instead of just not logging any metrics like before? Also, I don't fully understand the assertion message here

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -295,30 +329,52 @@ def _single_batched_evaluation(
...
+            eval_results[interval_idx][1]["metrics"] = [
+                {"name": metric.metric, "result": metric.result} for metric in interval_result.evaluation_data
+            ]
+        return eval_results
```
MaxiBoether, line 377: Could you maybe elaborate in a comment how you construct eval_results here? I am a bit confused why there are two loops

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: actually in the function that calls this (the worker func) we handle failure reasons but i am not sure about this assertion and what its purpose is

MaxiBoether (approved):
Thanks Robin! Overall looks good, I had one comment on an assertion and loops building the eval_results where I don't get what's happening and whether that is correct

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: prepare_evaluation_request allows the start to be nullable, here it's nonnullable. And mypy doesn't allow narrowing those types in a nested setting.

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: I assumed that in the `eval_data` response, unsuccessful evaluations wouldn't show up.

see this comment in the evaluator.proto

```

  // Since not every interval evaluation from EvaluateModelRequest may be successful,
  // the EvaluationIntervalData contained in the EvaluationResultResponse must explicitly specify what interval this
  // evaluation data corresponds to. The interval_index is the index of the interval in the list
  // Datainfo.evaluation_intervals in the EvaluateModelRequest.
  // For example if Datainfo.evaluation_intervals have 3 intervals, [interval1, interval2, interval3],
  // and interval2 fails. Then the EvaluationResultResponse will have 2 EvaluationIntervalData, one with interval_index
  // 0 (which corresponds to interval1) and the other with interval_index 2 (which corresponds to interval3).
```

if that assumption does not hold, we will find out soon enough. Better to have one assertion too much than missing oneone😅

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
robinholzi: There are two grpc calls that one that returns the `dataset_size` and potential failure messages, the other one a sparse (only successful intervals) the metric results.

robinholzi: I added a comment

modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
MaxiBoether: Ahh, sorry, I got confused (as expected) over the index stuff

MaxiBoether (approved):
LGTM!

robinholzi:
I'll merge this now with the rho-loss unit test failing

XianzheMa:
the failed one is a slightly flaky test. no problem at all

PR #560, opened 2024-07-03
Support threshold-based triggering for alibi-detect MMD

MaxiBoether:
Per the title

### DIFF
--- a/benchmark/wildtime_benchmarks/example_pipelines/data_drift_trigger/huffpost_datadrift.yaml
+++ b/benchmark/wildtime_benchmarks/example_pipelines/data_drift_trigger/huffpost_datadrift.yaml
@@ -46,6 +46,7 @@ trigger:
   metrics:
     ev_mmd:
       id: AlibiDetectMmdDriftMetric
+      num_permutations: 1000
   aggregation_strategy:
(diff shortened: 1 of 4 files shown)


benchmark/wildtime_benchmarks/example_pipelines/data_drift_trigger/huffpost_datadrift.yaml
```diff
@@ -46,6 +46,7 @@ trigger:
   metrics:
     ev_mmd:
       id: AlibiDetectMmdDriftMetric
+      num_permutations: 1000
```
robinholzi, line 49: do we want to do permutation testing after all?

robinholzi: or is this just to make the unit tests happy?

MaxiBoether: this was just to make the unit tests happy, since this an example pipeline. these are not the experiments i want to run :D

robinholzi (approved):
🪨 solid

PR #562, opened 2024-07-04
fix: Adjust currently trained model to be defined for dataset end

robinholzi:
# Motivation

With the last training having finished before the last k evaluation intervals, those intervals will use the last model as a currently active model, however no currently trained model exists. The evaluation plots will therefore be empty at the end skewing the average of the evaluation metric.

# Changes

Redefine the currently trained model to be defined on for the dataset end. The last model is allowed to be used as currently trained and currenly active model at the same time


MaxiBoether (approved):
n🍨

PR #564, opened 2024-07-04
feat: Add more powerful drift windowing strategies, warmup and dynamic thresholds

robinholzi:
# Motivation

Allow for more powerful drift windowing, dynamic drift threshold as well as a warmup period (for calibration)


modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -1,12 +1,35 @@
 from __future__ import annotations
```
robinholzi, line 2: This file needs the most thorough review!

robinholzi:
@MaxiBoether fyi. this should be ready now :) 
** I haven't run a pipeline with this yet, so maybe we still need some minor tweaks in a followup PR.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -200,16 +238,32 @@ def _run_detection(self) -> tuple[bool, dict[str, MetricResult]]:
...
+                if isinstance(metric_result.distance, float)
+                else float(mean(metric_result.distance))
+            )
+            drift_results[metric_name].is_drift = self.decision_engines[metric_name].evaluate_decision(
+                distance, metric_result.is_drift
+            )
+
         logger.info(f"[DataDriftDetector][Dataset {self.dataloader_info.dataset_id}]" + f"[Result] {drift_results}")
+        if is_warmup:
+            return False, {}  # unused dummy
```
MaxiBoether, line 266: not sure what you mean with this comment

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -127,39 +137,67 @@ def inform(
...
+            initial_batch_pointer += len(next_detection_interval)
+            new_key_ts = new_key_ts[self._sample_left_until_detection :]
+
+            # The first ever detection will always trigger
+            if not self._triggered_once or len(self.warmup_intervals) < (self.config.warmup_intervals or 0):
+                # If we've never triggered before, always trigger
+                self._triggered_once = True
+                triggered = True
+                drift_results: dict[str, MetricResult] = {}
+                self.warmup_intervals.append(list(self._windows.reference_))
```
MaxiBoether, line 164: why do we always append this, even if not using warmups?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -200,16 +238,32 @@ def _run_detection(self) -> tuple[bool, dict[str, MetricResult]]:
...
+            **self.evidently_detector.detect_drift(reference_embeddings_df, current_embeddings_df, is_warmup),
+            **self.alibi_detector.detect_drift(reference_embeddings, current_embeddings, is_warmup),
         }
+
+        # make the final decisions with the decision engines
+        for metric_name, metric_result in drift_results.items():
+            distance = (
+                metric_result.distance
+                if isinstance(metric_result.distance, float)
+                else float(mean(metric_result.distance))
```
MaxiBoether, line 258: why is this aggregation needed? maybe we can add a comment here

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -200,16 +238,32 @@ def _run_detection(self) -> tuple[bool, dict[str, MetricResult]]:
...
+            )
+            drift_results[metric_name].is_drift = self.decision_engines[metric_name].evaluate_decision(
```
MaxiBoether, line 260: I think one part that I find confusing in naming is that we have drift _detectors_. Detectors detect drift and return a binary is_drift variable. However now we don't necessarily  use the `is_drift` result. I think the separation of concerns is not clear yet. For example:

https://github.com/eth-easl/modyn/blob/0d4217642fd04607a1c02d7cdd5ab35b689980fc/modyn/supervisor/internal/triggers/drift/detector/alibi.py#L70C17-L70C25

here we implement the threshold check. 

we reimplement the check then in the decision engine. 

I think we want to refactor the interactions between detector, aggregation logic, and the decision engine.

modyn/supervisor/internal/triggers/drift/detection_window/timestamp.py
```diff
@@ -0,0 +1,75 @@
+from collections import deque
```
MaxiBoether, line 1: codecov is complaining about several missing tests here

MaxiBoether (changes requested):
Thank you, Robin. I did a first pass. While I did not fully grasp everything yet, I think my major feedback for the next iteration of this PR is that we should think about the interactions between the different components and potentially document them in some README.

In particular, we have aggregation logic for decisions in the pydantic module, we have the drift _detectors_ and then some duplicated decision logic in the decision engine. I think we want to consolidate the how we make drift/no drift decisions in a central place somehow. I also propose to reconsider the naming of the windowing manager.  Last, I also encountered some duplication in the new pydantic classes.

I hope my comments make sense. Basically I think the naming and levels of abstraction need some adjustment. Happy to discuss tomorrow or via Github comments!

modyn/supervisor/internal/triggers/models.py
```diff
@@ -7,12 +7,13 @@
 
 class DriftTriggerEvalLog(ModynBaseModel):
```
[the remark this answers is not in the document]
robinholzi: iirc, this is a class used for logging. We might also use it in the business logic, but that's completely fine as pydantic v2 is fast and we don't do complex validation that would result in vastly different performance compared to dataclasses.

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
robinholzi: Yep, I'll add some tests here

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class AmountWindowingStrategy(_BaseWindowingStrategy):
+    id: Literal["AmountWindowingStrategy"] = Field("AmountWindowingStrategy")
+    amount_ref: int = Field(1000, description="How many data points should fit in the reference window", ge=1)
+    amount_cur: int = Field(1000, description="How many data points should fit in the current window", ge=1)
+
+    @property
+    def current_buffer_size(self) -> int | None:
+        return self.amount_cur
```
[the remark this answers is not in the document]
robinholzi: The model is used as a Union of several `WindowingStrategies`. Some don't let the user choose this field and, therefore, return None. The None here is for type consistency. Arguably it might also work to return the narrowest type and hope that the Union type resolves the property to a Union return value.

modyn/config/schema/pipeline/trigger/drift/detection_window/timestamp.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
robinholzi: I tried `detection_window/time.py`, but that resulted in clashes with the standard library, producing weird errors. Not sure how to name this, maybe "time_.py"?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -36,34 +56,24 @@ def __init__(self, config: DataDriftTriggerConfig):
...
+            config.detection_interval_data_points
+        )  # allows to detect drift in a fixed interval
+        self._windows = _setup_detection_window_manager(config.windowing_strategy)
         self._triggered_once = False
 
         self.evidently_detector = EvidentlyDriftDetector(config.metrics)
         self.alibi_detector = AlibiDriftDetector(config.metrics)
+        self.decision_engines = _setup_decision_engines(config)
+
+        # list of reference windows for each warmup interval
```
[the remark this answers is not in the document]
robinholzi: no, I don't see any reason to have this.

modyn/config/schema/pipeline/trigger/drift/detection_window/window.py
```diff
@@ -0,0 +1,65 @@
...
+    allow_overlap: bool = Field(
+        False,
+        description=(
+            "Whether the windows are allowed to overlap. This is useful for time-based windows."
+            "If set to False, the current window will be reset after each trigger."
+        ),
+    )
+
+
+class AmountWindowingStrategy(_BaseWindowingStrategy):
```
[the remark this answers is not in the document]
robinholzi: yes, this file shouldn't contain the amount and time window strategies

modyn/config/schema/pipeline/trigger/drift/detection_window/window.py
```diff
@@ -0,0 +1,65 @@
...
+    id: Literal["AmountWindowingStrategy"] = Field("AmountWindowingStrategy")
+    amount_ref: int = Field(1000, description="How many data points should fit in the reference window", ge=1)
+    amount_cur: int = Field(1000, description="How many data points should fit in the current window", ge=1)
+
+    @property
```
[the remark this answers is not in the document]
robinholzi: removed

modyn/config/schema/pipeline/trigger/drift/detection_window/window.py
```diff
@@ -0,0 +1,65 @@
...
+        return num * SECONDS_PER_UNIT[unit]
+
+    @cached_property
+    def limit_seconds_cur(self) -> int:
+        unit = str(self.limit_cur)[-1:]
+        num = int(str(self.limit_cur)[:-1])
+        return num * SECONDS_PER_UNIT[unit]
+
+    @property
+    def current_buffer_size(self) -> int | None:
```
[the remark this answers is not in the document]
robinholzi: removed

modyn/config/schema/pipeline/trigger/drift/metric.py
```diff
@@ -0,0 +1,47 @@
...
+
+from modyn.config.schema.base_model import ModynBaseModel
+from pydantic import Field
+
+
+class HypothesisTestDecisionCriterion(ModynBaseModel):
+    id: Literal["HypothesisTestDecisionCriterion"] = "HypothesisTestDecisionCriterion"
+
+    @property
+    def needs_calibration(self) -> bool:
```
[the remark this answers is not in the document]
robinholzi: let's make it shorter and use add a field `needs_calibration` that has a unary literal type (either `Literal[True]` or `Literal[False]` dep. on the class)

modyn/supervisor/internal/triggers/drift/detection_window/timestamp.py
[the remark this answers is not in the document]
robinholzi: Let's see what the next commit will result in. The imports used the duplicated file `window.py`. There should be tests in theory.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -1,12 +1,31 @@
 from __future__ import annotations
```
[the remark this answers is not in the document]
robinholzi: I'll a some docs once we converged on a naming convention.
What about:
- `DataDriftTrigger`: With `Trigger` as the root "policy" class that the supervisor interacts with (like timetrigger, ...)
- **windows**: your correct that it's a wrapper around the actual window dequeues, and it implements the different strategy. I'm hesitant to call it `WindowStrategy` (and postfix the pydantic model with `Config`) as I wouldn't suspect a composite relation to the actual dequeues in `strategy` - it sounds functional/stateless. At the same time `DetectionWindows` is better, but hides the aspect that there are strategies. Maybe `AbstractDetectionWindows` and then subclasses `TimeDetectionWindows`, ...? wdyt
- `Detector`: wrapper of the detection library that supports different metrics and distance value generation (also raw decisions via hypothesis tests, which make no sense in our setting). I think the name is okay.
- `DecisionEngine`. A helper component that supports different ways to derive a drift decision from raw drift detections (hyp. tests, not useful), and the raw distance scores. Either forwarding the raw drift detection, comparing the distance with a threshold, or implementing this confidence level dynamic policy (that has its own queue as memory for historic distance observation used for calibration). Maybe `DecisionPolicy` would be an appropriate name?

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: the dictionary is just a placeholder as we don't have any useful object to return in the warmup region, i remove the comment as it doesn't really help

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: good catch, simply a bug

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: We could implement sth. like with an EnsemblePolicy (I think we already have the configuration models for that). However, for drift, we don't want to recompute embeddings if we want to evaluate two different metrics (e.g., mmd and classifier); that's why we do aggregation here.

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: I find it very unlikely that we will ever use the drift decisions from the detectors, as hypothesis testing is a dead end. The threshold comparison in `alibi.py` is a simple attempt to get the same interface as we get from evidently. With all my offline exploration in mind, I'd propose to remove all the decision support from the detectors (so they will only serve the distance computation --> rename to e.g. `DistanceDevice`?). Then, we could strip all the hypothesis configuration support. Wdyt?

modyn/supervisor/internal/triggers/datadrifttrigger.py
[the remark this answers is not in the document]
robinholzi: related https://github.com/eth-easl/modyn/pull/564#discussion_r1707974252

modyn/supervisor/internal/triggers/drift/decision_engine.py
```diff
@@ -0,0 +1,76 @@
+from abc import ABC, abstractmethod
+from collections import deque
+from typing import Deque
+
+from modyn.config.schema.pipeline.trigger.drift.metric import DynamicThresholdCriterion, ThresholdDecisionCriterion
+
+
+class DriftDecisionEngine(ABC):
+    """
+    Decision policy that will make the binary is_drift decisions based on the similarity/distance metrics.
```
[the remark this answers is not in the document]
robinholzi: `aggregation_strategy` works on the result of the decision engines from multiple metrics (mmd, classifier, ...).

modyn/supervisor/internal/triggers/drift/decision_engine.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Each drift decision wraps one DriftMetric and observes it's time series of distance values.
```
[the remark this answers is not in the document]
robinholzi: I think we resolved that in person by now.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -168,7 +206,9 @@ def inform_previous_model(self, previous_model_id: int) -> None:
 
     # --------------------------------------------------- INTERNAL --------------------------------------------------- #
 
-    def _run_detection(self) -> tuple[bool, dict[str, MetricResult]]:
+    def _run_detection(
+        self, reference: list[tuple[int, int]], current: list[tuple[int, int]], is_warmup: bool
```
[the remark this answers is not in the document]
robinholzi: We also use this function at the end of the warmup period to compute a series of drift distances for calibration of dynamic policies
```python
  if len(self.warmup_intervals) > 0:
      # we can ignore the results as the decision criteria will keep track of the warmup results
      # internally
      for warmup_interval in self.warmup_intervals:
          # we generate the calibration with different reference windows, the latest model and
          # the current window
          _warmup_triggered, _warmup_results = self._run_detection(
              warmup_interval,
              list(self._windows.current_),
              is_warmup=True,
          )
```

docs/pipeline/triggering/DRIFT_TRIGGERS.md
```diff
@@ -0,0 +1,231 @@
+# Drift Triggering
```
robinholzi, line 1: This should explain how drift based triggering works in modyn now. Wdyt?

docs/pipeline/triggering/DRIFT_TRIGGERS.md
```diff
@@ -0,0 +1,231 @@
...
+        +void inform_trigger()
+    }
+
+    DetectionWindows <|-- AmountDetectionWindows
+    DetectionWindows <|-- TimeDetectionWindows
+```
+
+### DriftDetector Hierarchy
+
+The `DriftDetector` class is an abstract base class for detectors like `AlibiDriftDetector` and `EvidentlyDriftDetector`, which use different metrics to measure the distance between the current and reference data distributions. It generates distance values.
```
MaxiBoether, line 94: Maybe add one sentence here that the underlying packages these detectors rely on also support making decisions, we implement our own decision logic in our drift decision policy, and ignore the output from the detectors.

docs/pipeline/triggering/DRIFT_TRIGGERS.md
```diff
@@ -0,0 +1,231 @@
...
+
+    DecisionCriterion <|-- ThresholdDecisionCriterion
+    DecisionCriterion <|-- DynamicThresholdCriterion
+```
+
+### DriftDecisionPolicy Hierarchy
+
+The `DriftDecisionPolicy` class is an abstract base class for policies like `ThresholdDecisionPolicy`, `DynamicDecisionPolicy`, and `HypothesisTestDecisionPolicy`.
+
+Each Decision policy wraps one DriftMetric (e.g. MMD, CVM, KS) and one DecisionCriterion (e.g. Threshold, DynamicThreshold, HypothesisTest) to make a decision based on the distance metric. It e.g. observes the series of distance value measurements from it's `DriftMetric` and makes a decision after having calibrated on the seen distances.
```
MaxiBoether, line 196:
```suggestion
Each decision policy wraps one DriftMetric (e.g. MMD, CVM, KS) and one DecisionCriterion (e.g. Threshold, DynamicThreshold, HypothesisTest) to make a decision based on the distance metric. It e.g. observes the series of distance value measurements from it's `DriftMetric` and makes a decision after having calibrated on the seen distances.
```

modyn/config/schema/pipeline/trigger/drift/config.py
```diff
@@ -21,51 +22,35 @@
...
+        description="Which windowing strategy to use for current and reference data",
     )
-
-    reset_current_window_on_trigger: bool = Field(
-        False, description="Whether the current window should be reset on trigger or rather be extended."
+    warmup_intervals: int | None = Field(
+        None,
+        description=(
+            "The number of intervals before starting to use the drift detection. Some "
+            "`DecisionCriteria` use this to calibrate the threshold. During the warmup, every interval will cause "
```
MaxiBoether, line 46:
```suggestion
            "`DecisionCriteria` use this to calibrate the threshold. During the warmup, every detection will cause "
```

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+from modyn.config.schema.base_model import ModynBaseModel
+from modyn.const.regex import REGEX_TIME_UNIT
+from modyn.utils.utils import SECONDS_PER_UNIT
+
+
+class _BaseWindowingStrategy(ModynBaseModel):
+    allow_overlap: bool = Field(
+        False,
+        description=(
+            "Whether the windows are allowed to overlap. This is useful for time-based windows."
```
MaxiBoether, line 14: Why is it not useful for amount based windows? In my understanding this is relevant for the time after a trigger is performed, and we could have overlaps also in amount windows?

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+        return num * SECONDS_PER_UNIT[unit]
+
+    @cached_property
+    def limit_seconds_cur(self) -> int:
+        unit = str(self.limit_cur)[-1:]
+        num = int(str(self.limit_cur)[:-1])
+        return num * SECONDS_PER_UNIT[unit]
+
+
+DriftWindowingStrategy = Annotated[
```
MaxiBoether: We also define this in the __init__.py

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Annotated, Literal
+
+from pydantic import Field
+
+from modyn.config.schema.base_model import ModynBaseModel
+from modyn.const.regex import REGEX_TIME_UNIT
+from modyn.utils.utils import SECONDS_PER_UNIT
+
+
+class _BaseWindowingStrategy(ModynBaseModel):
```
MaxiBoether, line 10: should we move this to init.py / separate file? seems arbitrary that the base strategy lives in the amount window and the time window has to import it

MaxiBoether: actually, we also have window.py. so this is duplicated

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, config: ThresholdDecisionCriterion):
+        self.config = config
+
+    def evaluate_decision(self, distance: float) -> bool:
+        return distance >= self.config.threshold
+
+
+class DynamicDecisionPolicy(DriftDecisionPolicy):
+    """Decision policy that will make the binary is_drift decisions based on a
+    dynamic threshold."""
```
MaxiBoether, line 41: Maybe you could add one more sentence here explaining how the dynamic threshold is set, i.e., the logic around `num_more_extreme`

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__()
+        self.config = config
+
+        # using maxlen the deque will automatically remove the oldest elements if the buffers are full
+        self.current: deque[tuple[int, int]] = deque(maxlen=config.amount_cur)
+        self.current_reservoir: deque[tuple[int, int]] = deque(maxlen=max(0, config.amount_ref - config.amount_cur))
+        self.reference: deque[tuple[int, int]] = deque(maxlen=config.amount_ref)
+
+        # in overlapping mode (we need dedicated buffer to keep track of the new samples that are not in
+        # the reference buffer, yet). The current_ and current_reservoir_ are not enough as after
```
MaxiBoether, line 19: this sentence is somewhat not complete/weird

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class AmountDetectionWindows(DetectionWindows):
+    def __init__(self, config: AmountWindowingStrategy):
+        super().__init__()
+        self.config = config
+
+        # using maxlen the deque will automatically remove the oldest elements if the buffers are full
+        self.current: deque[tuple[int, int]] = deque(maxlen=config.amount_cur)
+        self.current_reservoir: deque[tuple[int, int]] = deque(maxlen=max(0, config.amount_ref - config.amount_cur))
```
MaxiBoether, line 15: maybe add a comment for what the reservoir is needed?

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if self.config.allow_overlap:
+            # use the dedicated buffer that tracks the new elements to be copied to reference on trigger
+            self.exclusive_current.extend(data)
+        else:
+            # use the existing buffers
+            num_pushed_out = len(self.current) + len(data) - self.config.amount_cur
+
+            # move data from current window to reservoir
+            for pushed_out in self.current:
```
MaxiBoether, line 37: does just iterating over a deque remove it? otherwise I'd have suggested to use a while loop here instead of this for loop with a break pattern

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+                if num_pushed_out == 0:
```
MaxiBoether, line 38: maybe `remaining_pushes`?

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            # First, move data from the reservoir window to the reference window
+            self.reference.extend(self.current_reservoir)
+            self.current_reservoir.clear()
+
+            # Move data from current to reference window
+            self.reference.extend(self.current)
+
+            # if allow_overlap, don't reset the current window
+            self.current.clear()
```
MaxiBoether, line 60: I don't get this comment yet. We're in the else branch here, hence allow_overlap is False

modyn/supervisor/internal/triggers/drift/detection_window/window.py
```diff
@@ -0,0 +1,31 @@
...
+from collections import deque
+
+
+class DetectionWindows(ABC):
+    """Wrapper and manager for the drift detection windows include reference,
+    current and reservoir window.
+
+    All windows contain tuples with (sample_id, timestamp).
+
+    The manager is responsible for the following tasks:
```
MaxiBoether, line 11:
```suggestion
    This class is responsible for the following tasks:
```

modyn/supervisor/internal/triggers/drift/detection_window/window.py
```diff
@@ -0,0 +1,31 @@
...
+    If the reference window is bigger than the current window, we still want to fill up the whole reference window
+    after a trigger by taking |reference| elements from the current window.
+    We therefore need to keep track of the elements that exceed the current window but should still be transferred
+    to the reference window. If something is is dropped from the reservoir, it won't ever be used again in any window.
+    """
+
+    def __init__(self) -> None:
+        self.current: deque[tuple[int, int]] = deque()
+        self.current_reservoir: deque[tuple[int, int]] = deque()
+        self.reference: deque[tuple[int, int]] = deque()
```
MaxiBoether, line 25: the child classes currently also initialize that

modyn/supervisor/internal/triggers/drift/detector/evidently.py
```diff
@@ -44,7 +45,13 @@ def detect_drift(
         column_mapping = ColumnMapping(embeddings={EVIDENTLY_COLUMN_MAPPING_NAME: embeddings_ref.columns})
 
         # https://docs.evidentlyai.com/user-guide/customization/embeddings-drift-parameters
-        report = Report(metrics=[self.evidently_metrics[name] for name in self.evidently_metrics])
+        report = Report(
+            metrics=[
+                self.evidently_metrics[name][1]
+                for name in self.evidently_metrics
+                if not is_warmup or self.evidently_metrics[name][0].decision_criterion.needs_calibration
+            ]
```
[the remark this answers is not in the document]
MaxiBoether: i guess what we could do is instantiate another trigger (e.g., a timetrigger or amount trigger) and pass that to the drift trigger. the drift trigger could instantiate this trigger and during warmup, forward the calls to the simpler policy, while taking care of building its internal state. WDYT?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -36,34 +67,27 @@ def __init__(self, config: DataDriftTriggerConfig):
         self.encoder_downloader: EmbeddingEncoderDownloader | None = None
         self.embedding_encoder: EmbeddingEncoder | None = None
 
-        self._reference_window: list[tuple[int, int]] = []
-        self._current_window: list[tuple[int, int]] = []
-        self._total_items_in_current_window = 0
+        self._sample_left_until_detection = (
+            config.detection_interval_data_points
+        )  # allows to detect drift in a fixed interval
+        self._windows = _setup_detection_window_manager(config.windowing_strategy)
```
MaxiBoether, line 73: rename to setup detection windows?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -36,34 +67,27 @@ def __init__(self, config: DataDriftTriggerConfig):
...
         self._triggered_once = False
 
         self.evidently_detector = EvidentlyDriftDetector(config.metrics)
         self.alibi_detector = AlibiDriftDetector(config.metrics)
 
+        # Every decision policy wraps one metric and is responsible for making decisions based on the metric's results
+        # and the metric's range of distance values
+        self.decision_policies = _setup_decision_engines(config)
```
MaxiBoether, line 81: rename to setup decision policies?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -36,34 +67,27 @@ def __init__(self, config: DataDriftTriggerConfig):
...
+
+        # list of reference windows for each warmup interval
+        self.warmup_intervals: list[list[tuple[int, int]]] = []
```
MaxiBoether, line 84: I guess it's still unclear to me here why we can have multiple warmup intervals

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -127,41 +154,85 @@ def inform(
...
+            self._sample_left_until_detection = self.config.detection_interval_data_points
+
+            # The first ever detection will always trigger
+            if not self._triggered_once or len(self.warmup_intervals) < (self.config.warmup_intervals or 0):
+                # If we've never triggered before, always trigger
+                self._triggered_once = True
+                triggered = True
+                drift_results: dict[str, MetricResult] = {}
+                if len(self.warmup_intervals) < (self.config.warmup_intervals or 0):
+                    self.warmup_intervals.append(list(self._windows.reference))
```
MaxiBoether, line 190: it seems we currently don't have a test for the warmup case (according to codecov) (but I am not too sure about the list of intervals in the first case. in my understanding so far warmup is either True or False)

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -127,41 +154,85 @@ def inform(
...
+
+            else:
+                # Run the detection
+
+                # if this is the first non warmup detection, we inform the metrics that use decision criteria
+                # with calibration requirements about the warmup intervals so they can calibrate their thresholds
+                if len(self.warmup_intervals) > 0:
+                    # we can ignore the results as the decision criteria will keep track of the warmup results
+                    # internally
+                    if self._any_metric_needs_calibration():
```
MaxiBoether, line 200: According to codecov, the calibration/warmup logic does not seem to be tested here

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -203,16 +277,33 @@ def _run_detection(self) -> tuple[bool, dict[str, MetricResult]]:
 
         # Compute embeddings
         assert self.embedding_encoder is not None
+
+        # tbd. reuse the embeddings as long as the reference window is not updated
```
MaxiBoether, line 281: should this become an issue with a proper TODO?

modyn/supervisor/internal/triggers/datadrifttrigger.py
MaxiBoether: I still don't fully understand this. We already have the ensemble logic in case we have e.g. a threshold decision engine with MMD and a threshold decision policy with euclidean, and then we can e.g. do majority vote. I don't understand here yet why we averaging multiple metrics is required and what is the use case.

docs/pipeline/triggering/DRIFT_TRIGGERS.md
MaxiBoether, line 196: Should we explain the terminology we use in context of the dynamic policy somewhere? For a new reader of the drift trigger file, the terms calibiration and warmup might still be confusing, might be good to define them somewhere / explain the calibration logic

modyn/supervisor/internal/triggers/datadrifttrigger.py
MaxiBoether: metrics can also live on completely different scales I suppose (like 5000 and 0.005 - it really depends on the metric itself). Can this just be averaged?

MaxiBoether (changes requested):
Thank you, Robin! This is looking good. Most of my comments are nitpicky, I had a couple of smaller questions in some parts

modyn/supervisor/internal/triggers/drift/detector/evidently.py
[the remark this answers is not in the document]
robinholzi: Yes, that's what I had in mind.

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: I think i got confused myself when writing the comment there. This averaging is only creating scalar distance values from an array of distance values returned by an alibi-detect metric (some of the metrics there don't return a single distance value but a distance value for every sample)

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: Yes, I will add a test for the warmup logic

modyn/supervisor/internal/triggers/datadrifttrigger.py
MaxiBoether: Ahhh, ok. Maybe we should move this logic then to the detector level? Because from the detector interface, the returned metric should be a scalar value, not a list, shouldn't it?

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
robinholzi: You're right. We can also decide the overlap based on the sample IDs. When I first wrote this, I only had the sample times available in the offline exploration. The implementation already supports it :)

modyn/config/schema/pipeline/trigger/drift/detection_window/amount.py
robinholzi: Was already resolved at the time of review submission.

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
robinholzi: No, the `remaining_pushes` variable that we computed before that signals the number of elements that will be dropped through the `self.current.extend(data)` at the end. As we might still need those elements in the reference buffer. We copy `remaining_pushes` to the reservoir manually beforehand.

robinholzi: I added a comment to make that clearer

modyn/supervisor/internal/triggers/drift/detection_window/amount.py
robinholzi: I shuffled the logic a bit after the initial draft before the sigmod submission. Probably forgot to remove the comment, thanks for the hint!

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: Assume we have 5 warmup intervals. Then we first collect the first 5 references windows and do the triggering decision according to a simple time/amount based warmup policy. After finishing the warmup we have collected a couple of data intervals and run the drift detection with the most recent current window and all of the collected references window. Each of the reference windows generates a distance value together with the most recent current window. With this series of distances we can calibrate the dynamic threshold decision policy. Does that make sense?

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: I added a quite extensive test, good that you mentioned this, there were still a couple of bugs in the implementation.

modyn/supervisor/internal/triggers/datadrifttrigger.py
MaxiBoether: Okay, I think my first level of confusion stemmed from the fact that I believed this means we somehow have more than 1 warmup phase. But I understand now this is about waiting for `n` windows until warmup ends.

What I am now confused about is why we need to cache the reference windows here. Couldn't we, on the fly during warmup, do the drift detection (without issuing a trigger), and remember the distance values? In my understanding, if we have a warmup duration of 5, we'd need to cache 5 distance metrics. I don't understand yet why we need to cache the windows instead. You say that you run the detection with the current window for all reference windows, but why? We want to calibrate the dynamic threshold. For this, shouldn't we use the current data as it will be later on? (Sorry, hard to articulate what I mean - basically during the actual runtime after warmup, we use current/reference windows, but they are not 5 windows apart. This might influence the magnitude of the distance values, since during warmup if we compare a very old reference window to a new current window that might lead to a larger distance value.)

robinholzi: Yes, this influences the magnitude, and that's exactly why we do it. After the warmup phase, we are on a "diagonal" element of the distance matrix. However, if we also use the corresponding reference/current window pairs from every interval in the warmup phase, we would see only close-to-zero if not only `0` distances, making the calibration useless. In my offline exploration, those policies then only took the diagonal walk. I treat this cache of windows more as a implementational detail that leads to a successful calibration.

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: Yes, let's do that, I also thought of that, but back then, I wanted the full detector results in the logs. Now I'm confident of the results so we can do that.

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -180,21 +159,42 @@ def inform(
...
+
                 drift_results: dict[str, MetricResult] = {}
                 if len(self.warmup_intervals) < (self.config.warmup_intervals or 0):
-                    self.warmup_intervals.append(list(self._windows.reference))
+                    # for the first detection the reference window is empty, therefore adding the current window
+                    self.warmup_intervals.append(
+                        list(self._windows.reference if self._triggered_once else self._windows.current)
+                    )
+
+                self._triggered_once = True
```
MaxiBoether, line 188: I am wondering about the case where

1) we dont have warmup (=> triggered once is checked)
2) we specify a warmup trigger

what if that warmup trigger does not trigger?

then we set triggered once to true, despite potentially not triggering. is that intended?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -278,7 +310,7 @@ def _run_detection(
         # Compute embeddings
         assert self.embedding_encoder is not None
 
-        # tbd. reuse the embeddings as long as the reference window is not updated
+        # TODO: reuse the embeddings as long as the reference window is not updated
```
MaxiBoether: could we open an issue about this so we dont forget

modyn/supervisor/internal/triggers/datadrifttrigger.py
MaxiBoether: Okay, I see. I guess you're going to write about that in your thesis, if you already have a text statement prepared about this, would be good to explain this in a comment or the markdown documentation i suppose, because that was not obvious to me

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -38,7 +41,16 @@ def evaluate_decision(self, distance: float) -> bool:
 
 class DynamicDecisionPolicy(DriftDecisionPolicy):
     """Decision policy that will make the binary is_drift decisions based on a
-    dynamic threshold."""
+    dynamic threshold.
+
+    We compare a new distance value with the series of previous distance values
+    and decide if it's more extreme than a certain percentile of the series. Therefore we count the
+    `num_more_extreme` values that are greater than the new distance and compare it with the
+    `percentile` threshold.
```
MaxiBoether, line 49: For example here we could explain how we build this, or in the drift triggers.md

MaxiBoether (approved):
Thank you!! Some last nitpicky comments, feel free to merge afterwards.

modyn/supervisor/internal/triggers/datadrifttrigger.py
robinholzi: if the warmup policies doesn't trigger in the first interval, that doesn't matter, for the first interval we force a trigger

```python
triggered = (
    delegated_trigger_results is not None
    if self._triggered_once
    else True  # first candidate always triggers
)
```

modyn/supervisor/internal/triggers/drift/decision_policy.py
robinholzi: I went for `DRIFT_TRIGGERS.md`

robinholzi: I just copied a rough draft for the description of our solution; in my thesis, I'll use visualizations. It's rather hard to convey the issues here with plain text.

PR #569, opened 2024-07-08
feat: Add `SIGMOD` plots

robinholzi:
# Motivation

As we need the logfile models to bring the data into the correct shape, this logic needs to reside within this repo.

Optionally we could only generate the data here (store in .csvs) and put the actual plotting logic somewhere else. I don't have any preference on this.


MaxiBoether (approved):
Since this has been tested by the both of us, LGTM! I guess we need to update to the main branch here?

PR #487, opened 2024-06-09
[ignore] experiment stuff

MaxiBoether:
when i lie in bed contemplating my life choices (e.g., which learning rate I chose for CLOC), then I sometimes like to scroll over my code and it makes it a lot easier to have an open PR to find my branch ;)

PR #579, opened 2024-07-22
feat: Add `nbstripout`

robinholzi:
# Motivation

As we have a couple of Jupyter notebooks in the repo now, let's ensure nobody commits the notebooks without clearing the outputs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,10 @@ repos:
     rev: v1.20.9
     hooks:
       - id: typos
+  - repo: https://github.com/kynan/nbstripout
+    rev: 0.7.1
+    hooks:
+      - id: nbstripout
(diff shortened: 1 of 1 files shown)


MaxiBoether (approved):
Generally LGTM, there seems to be a mypy error from a new mypy version. I wonder whether nbstripout can also be part of the CI? I guess it's better to also enforce this in the CI rather than just use pre commit hooks.

robinholzi:
> Generally LGTM, there seems to be a mypy error from a new mypy version. I wonder whether nbstripout can also be part of the CI? I guess it's better to also enforce this in the CI rather than just use pre commit hooks.

Yep, we can easily run the pre-commit hooks in the CI :) I wanted to wait with this until we enable ruff on the whole code base. Do you intend to merge https://github.com/eth-easl/modyn/pull/487? If yes, I'd rather wait until this is done to avoid conflicts.

PR #580, opened 2024-07-23
Split Criteo

XianzheMa:


modyn/trainer_server/internal/dataset/key_sources/__init__.py
```diff
@@ -3,14 +3,11 @@
...
 
 import os
-from enum import Enum
 
 from .abstract_key_source import AbstractKeySource  # noqa: F401
 from .local_key_source import LocalKeySource  # noqa: F401
 from .selector_key_source import SelectorKeySource  # noqa: F401
 
-KeySourceNames = Enum("KeySource", ["LOCAL_DOWNSAMPLER", "SELECTOR"])
-
```
XianzheMa, line 13: not used anywhere, and complained by mypy for different names (KeySourceNames vs KeySource)

benchmark/criteo_1TB/preprocessing/train_eval_split.py
```diff
@@ -0,0 +1,142 @@
...
+        target_criteo_path.mkdir(parents=True)
+    # create the train and test directories
+    target_train_path = target_criteo_path / "train"
+    target_test_path = target_criteo_path / "test"
+    if not target_train_path.exists():
+        logger.info(f"Creating target train path: {target_train_path}")
+        target_train_path.mkdir(parents=True)
+    if not target_test_path.exists():
+        logger.info(f"Creating target test path: {target_test_path}")
+        target_test_path.mkdir(parents=True)
```
MaxiBoether, line 89: i guess could also just do path.mkdir(exists_ok=True) but then we cannot print, whatever you prefer

modyn/trainer_server/internal/dataset/extra_local_eval/binary_file_wrapper.py
```diff
@@ -85,3 +85,11 @@ def get_samples_from_indices(self, indices: list) -> list[bytes]:
 
         samples = [data[(idx * self.record_size) + self.label_size : (idx + 1) * self.record_size] for idx in indices]
         return samples
+
+    def persist_sub_file(self, indices: list, sub_file_path: str) -> None:
+        with open(self.file_path, "rb") as file:
```
MaxiBoether, line 90: Since this is not really using functionalites from the BinaryFileWrapper, let's move this into the train eval split python file. The "problem" is that this file here is not really supposed to exist within modyn, it's just for running local benchmarks, we should not rely on it. Let's move it to the criteo specific file.

MaxiBoether (approved):
Thanks and also thanks for fixing CI! My only point is that we should not rely on the python binary file wrapper as I lay out in the comment.

XianzheMa:
since linter-checks passed, we merge it.

PR #582, opened 2024-07-25
Port over miscellaneous changes from SIGMOD branch

MaxiBoether:


modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -745,6 +745,7 @@ def _setup_downsampling(
         self._downsampler = self._instantiate_downsampler(
             strategy_name, downsampler_config, modyn_config, self._criterion_nored
         )
+        self._log["received_downsampler_config"] = downsampler_config
```
XianzheMa, line 748: this can be deleted as it is already there at line 750/751

modyn/supervisor/internal/triggers/embedding_encoder_utils/embedding_encoder_downloader.py
```diff
@@ -24,12 +25,13 @@ class EmbeddingEncoderDownloader:
...
     def __init__(
         self,
-        modyn_config: dict,
+        modyn_config: ModynConfig,
         pipeline_id: int,
         base_dir: pathlib.Path,
         model_storage_address: str,
     ):
-        self.modyn_config = modyn_config
+        # TODO(MaxiBoether): Update this class to use the model
```
robinholzi, line 33: Not sure how much effort this is here, I remember that I have tried to integrate the models into the database components when introducing the pydantic models, that was a nightmare with all the tests that don't set all required fields. I wouldn't mind if we remove this TODO.

robinholzi (approved):
Didn't check the new dataset wrappers, but the rest seems fine

PR #583, opened 2024-07-26
docs: Enhance documentation

robinholzi:
# Motivation

The pipeline state machine graph was a bit outdated. Also added some cross-references and further documentation markdowns to the `docs/` dir.

This PR sets the basis for #442


MaxiBoether (approved):
Thanks for fixing this!

PR #584, opened 2024-07-31
refactor: Introduce ruff and other linting tools

robinholzi:
# Motivation

Previously, we always had to run the `python_compliance.sh`  script, which had intrinsic conflicts so that black and other linters changed each other's reformatting.

I sort was also not even executed on relevant directories like `modern`, leaving us with only sparsely formatted code.

Let's migrate to a ruff, a blazingly fast modern Python linter, and use both ruff linting and formatting on the whole codebase (including directories like `experiments`)!

CI workflows were adjusted accordingly.


robinholzi:
** Currently, GitHub Actions faces a partial outage; the CI workflow probably still needs some adjustments, so let's wait until https://www.githubstatus.com/ is green again!

.github/workflows/workflow.yaml
```diff
@@ -10,37 +10,39 @@ defaults:
...
-        uses: ./.github/actions/mamba
-
-      - name: Flake8
-        run: |
-          micromamba run -n modyn flake8 --version
-          micromamba run -n modyn flake8 modyn --statistics
+      - name: Checkout branch
+        uses: actions/checkout@v4
+      - name: Run pre-commit-conda
+        uses: quantco/pre-commit-conda@v1
```
MaxiBoether, line 21: Any reason for not using the mamba/action.yml that we use for the other jobs?

.github/workflows/workflow.yaml
```diff
@@ -10,37 +10,39 @@ defaults:
...
+          environment-file: environment.yml
+          cache-environment: true
+      - name: Install repository
         run: |
-          micromamba run -n modyn mypy --version
-          micromamba run -n modyn mypy modyn
+          python -m pip install --no-build-isolation --no-deps --disable-pip-version-check -e .
+          micromamba run -n modyn pip install -r dev-requirements.txt
+      - name: Run mypy
+        run: mypy .
```
MaxiBoether, line 45: Similarily here, we used to hide the setup of micromamba in the mamba action, not in the task itself. Any reason for changing the mypy job?

.github/workflows/workflow.yaml
```diff
@@ -58,40 +60,6 @@ jobs:
           micromamba run -n modyn pylint --version
           micromamba run -n modyn pylint modyn
 
-  isort:
```
MaxiBoether, line 61: just to be sure, ruff replaces isort as well?

modyn/storage/src/internal/database/sql/SQLUpdateDataset.sql
```diff
@@ -1,4 +1,4 @@
-R"(UPDATE datasets 
+R"(UPDATE datasets
```
MaxiBoether, line 1: Ah, I don't want to vouch for this working. Maybe we want to exclude .sql files from removing whitespaces. if newlines are not equal to a white space here, this would mean `UPDATE datasetsSET base_path` which would be problematic.

scripts/python_compliance.sh
```diff
@@ -16,50 +16,7 @@ else
     exit 1
 fi
 
-
-echo "Running auto-formatters"
```
MaxiBoether, line 20: We should document somewhere how to run autoformatting now (and potentially, run it when running this script as before, unless you think something speaks against it)

MaxiBoether (changes requested):
Generally LGTM! Thanks so much for taking care of this. 2 notes:

1) Of course before merging we should figure out why CI is consistently failing.

2) Could we document somewhere how to run the formatting? IIRC, the pre commit hook directly modifies my commit inline, whenever I `git commit` right? Because I have faced more than once that autoformatting destroyed something. Hence, I prefer to run autoformatting in a new commit. The pre commit checks action job now covers several checks (ruff, nbstripout, ...), correct?

.github/workflows/workflow.yaml
robinholzi: The local action does some things we don't need for this use case.

.github/workflows/workflow.yaml
robinholzi: If it's just the micromamba setup and nothing else, the indirection through a local actions doesn't really benefit us.

.github/workflows/workflow.yaml
robinholzi: Yes, https://docs.astral.sh/ruff/rules/#isort-i

modyn/storage/src/internal/database/sql/SQLUpdateDataset.sql
robinholzi: Right, we are using `.sql` files in the most unintuitive way through C++ preprocessor imports😅 In April I spent a couple of hours searching for a better way, but didn't find anything that C++ would accept. Seemingly, you cannot load files in compile time / constexpr functions😅

robinholzi: Excluding directories is possible :)

scripts/python_compliance.sh
robinholzi: autoformatting is done automatically through the pre-commit hooks. You should stage your changes before running a git commit to be able to undo things. If you don't want the pre-commit hooks to run on every commit, you should alias your git commit command with a `-n` flag. That's nothing that should be configured centrally.

robinholzi: I didn't want to include pylint in the pre-commit hooks, as it's prohibitively slow (20-30 sec.). The other hooks run in <5s total. It's still in the python_compliance script though (for CI)

modyn/storage/src/internal/database/sql/SQLUpdateDataset.sql
robinholzi: Doesn't seem to be harmful though

.github/workflows/workflow.yaml
MaxiBoether: But I mean the local action takes care of caching, no? In case of no updates, we can restore the environment from cache, speeding up the CI - that is what the local action does

.github/workflows/workflow.yaml
MaxiBoether: Similar to my other comment, it caches the environment though, doesn't it? And that speeds up the action

MaxiBoether:
We can merge this after figuring out the setup action. The reason I suggest to use that one is that it caches the environment.

MaxiBoether (approved):
Giving my approval here but let's figure out the caching thing

robinholzi:
> Giving my approval here but let's figure out the caching thing

I think we don't need this fancy caching setup with date and cache version. I'm removing this and defaulting to the default cache environment across all setup-micromamba occurrences. Let me you if there's a strong reason to have this, than I can restore it in a follow-up PR.

.github/workflows/workflow.yaml
robinholzi: we already did that with the `cache-environment` micromamba flag, just with a different (default) cache-key. Now we use the default cache environment everywhere.

.github/workflows/workflow.yaml
robinholzi: The action runs in <30 sec. We don't need to optimize this further.

PR #585, opened 2024-08-12
fix: Restore old caching in CI

robinholzi:
**Benefit**: central configuration of environment setup.

**Note**: mypy ci job will run ~1 min slower due to additional (unnecessary) dependencies in `dev-requirements.txt` and the apt install from the local action.


MaxiBoether (approved):
Thanks!

PR #586, opened 2024-08-12
WIP: Add simple performance and cost aware policies

robinholzi:
# WORK IN PROGRESS...

Will probably be split into multiple PRs after completion.

PR #588, opened 2024-08-15
refactor: Prepare performance aware triggers

robinholzi:
# Motivation

To reuse logic and have a more organised triggering setup, I'd like to make the following adjustments:

# Changes

- Moving core evaluation logic out of `pytorch` evaluator so it's reusable for performance-trigger evaluations.
- drop unused resultwriters
- move `get_embeddings` and `TriggerDatasets` to a more central place as they will be reuse for performance aware triggers.

I already made some comments in https://github.com/eth-easl/modyn/pull/586 that explain what is moved where.


modyn/evaluator/internal/metric_factory.py
```diff
@@ -1,23 +1,45 @@
...
-    def prepare_metrics(metrics: list[AbstractEvaluationMetric]) -> bool:
-        contains_holistic = False
-        for metric in metrics:
+    def prepare_metrics(metrics: dict[str, AbstractEvaluationMetric]) -> bool:
+        for metric in metrics.values():
             metric.deserialize_evaluation_transformer()
             if isinstance(metric, AbstractHolisticMetric):
                 return True
-        return contains_holistic
+        return False
```
MaxiBoether, line 45: This will return true as soon as we reach a holistic metric. This means for metrics after that we won't call the deserialize function. I think we should instead change return True to contains_holistic = True and return that at the end

modyn/supervisor/internal/triggers/drift/embedding/embeddings.py
```diff
@@ -0,0 +1,40 @@
+import torch
+from torch.utils.data import DataLoader
+
+from modyn.supervisor.internal.triggers.utils.model.manager import ModelManager
+
+
+def get_embeddings(model_manager: ModelManager, dataloader: DataLoader) -> torch.Tensor:
```
MaxiBoether, line 7: This needs a test

modyn/supervisor/internal/triggers/utils/model/manager.py
```diff
@@ -11,10 +11,10 @@
 logger = logging.getLogger(__name__)
 
 
-class EmbeddingEncoder:
-    """The EmbeddingEncoder stores a model and its metadata.
+class ModelManager:
```
MaxiBoether, line 14: can we find a better term than manager? it's actually considered to be a bit of a code smell to call something manager. maybe there is a gang of four pattern that matches this?

modyn/supervisor/internal/triggers/drift/utils.py
```diff
@@ -0,0 +1,10 @@
+import pandas as pd
+import torch
+
+
+def convert_tensor_to_df(t: torch.Tensor, column_name_prefix: str | None = None) -> pd.DataFrame:
```
MaxiBoether, line 5: This function is missing a test

MaxiBoether (changes requested):
Thank you, this generally looks good but i had a few nitpicky comments

modyn/supervisor/internal/triggers/drift/embedding/embeddings.py
robinholzi: Sure, I'll add one. This function also didn't have one before, though😅

modyn/supervisor/internal/triggers/utils/model/manager.py
robinholzi: `Wrapper`? `EmbeddingEncoder` was definitely off, as the class doesn't have anything to do with either embeddings or encoding😅

modyn/supervisor/internal/triggers/drift/utils.py
robinholzi: again technical debt🥶😅

modyn/evaluator/internal/metric_factory.py
robinholzi: good catch!

MaxiBoether (approved):
Thanks. Sorry, i didn't realize those functions were not tested before...

modyn/evaluator/internal/metric_factory.py
```diff
@@ -1,23 +1,46 @@
...
     @staticmethod
-    def prepare_metrics(metrics: list[AbstractEvaluationMetric]) -> bool:
+    def prepare_metrics(metrics: dict[str, AbstractEvaluationMetric]) -> bool:
         contains_holistic = False
-        for metric in metrics:
+        for metric in metrics.values():
             metric.deserialize_evaluation_transformer()
             if isinstance(metric, AbstractHolisticMetric):
-                return True
+                contains_holistic = True
```
XianzheMa, line 45: thanks for catching this

PR #589, opened 2024-08-15
feat: Add config models for performance aware triggers

robinholzi:
# Motivation

Our new performance-aware triggers (which are an intermediate step towards cost-aware triggers) need to define the configuration space.

** target branch marks the dependency of the PRs: this PR depends on #588


modyn/const/types.py
```diff
@@ -12,3 +12,24 @@ class TimeResolution(str, Enum):
     HOUR = "hour"
     MINUTE = "minute"
     SECOND = "second"
+
+
+TriggerEvaluationMode = Literal["hindsight", "lookahead"]
+"""
+- hindsight: The trigger is evaluated on the data that has already been seen. Once a we are certain about a threshold
+    crossing, the trigger is executed. Triggers are only performed after the threshold was crossed.
```
MaxiBoether, line 20: This is only for performance triggers, though? it reads a bit general

modyn/config/schema/pipeline/trigger/performance/performance.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class PerformanceTriggerEvaluationConfig(ModynBaseModel):
+    # max_samples_per_evaluation: int = Field(
+    #     1000,
+    #     description=(
+    #         "The maximum number of samples that should be used for the evaluation. If more are available, "
+    #         "downsampling will be used"
+    #     ),
+    # )  # TODO: currently not yet supported by the evaluator, we use a evaluation dataset that is sufficiently small
```
MaxiBoether, line 20: why is something to be supported by the evaluator? couldnt the trigger subsample the set?

modyn/config/schema/pipeline/trigger/performance/performance.py
```diff
@@ -0,0 +1,95 @@
...
+        None, description="Function used to transform the label (tensors of integers)."
+    )
+
+    @field_validator("dataset")
+    @classmethod
+    def validate_metrics(cls, dataset: EvalDataConfig) -> EvalDataConfig:
+        """Assert that we have at least the accuracy metric."""
+        if not any(metric.name == "Accuracy" for metric in dataset.metrics):
+            raise ValueError("The accuracy metric is required for the performance trigger.")
+        return dataset
```
MaxiBoether, line 34: Actually why do we need accuracy? If a user wants to trigger based on ROC-AUC (e.g., in recommendation systems), why not? I feel like that is not our beer

modyn/config/schema/pipeline/trigger/performance/criterion.py
```diff
@@ -0,0 +1,99 @@
...
+    allowed_deviation: float = Field(
+        0.05,
+        description=(
+            "The allowed deviation from the expected performance. Will only trigger if the performance is "
+            "below the expected performance minus the allowed deviation."
+        ),
+    )
+
+
+# TODO: drift: rolling average + x % (instead of percentile)
```
MaxiBoether, line 44: don't get this todo, should either expand or remove

modyn/config/schema/pipeline/trigger/performance/criterion.py
```diff
@@ -0,0 +1,99 @@
+from typing import Annotated, Literal
+
+from pydantic import Field
+
+from modyn.config.schema.base_model import ModynBaseModel
+
+# -------------------------------------------------------------------------------------------------------------------- #
+#                                                 PerformanceCriterion                                                 #
+# -------------------------------------------------------------------------------------------------------------------- #
+# uses the evaluation results directory to derive a triggering decision
```
MaxiBoether, line 10: directory?

modyn/config/schema/pipeline/trigger/performance/performance.py
```diff
@@ -0,0 +1,95 @@
...
+            "Every criterion is linked to a metric. Some of the criteria implicitly only work on accuracy which is "
+            "the default metric that is always generated and cannot be disabled. To define a "
+            "`StaticPerformanceThresholdCriterion` on Accuracy, the evaluation config has to define the accuracy metric."
+            # TODO: maybe support custom aggregation functions
+        ),
+        min_length=1,
+    )
+
+    mode: TriggerEvaluationMode = Field(
+        "hindsight",
```
MaxiBoether, line 70: As I probably have mentioned before, for the review of the actual algorithm, it will be helpful to have a README or something explaining this again. THanks!

MaxiBoether (changes requested):
This also looks good but I had a few nitpicky comments

modyn/const/types.py
robinholzi: probably also for the cost triggers

modyn/config/schema/pipeline/trigger/performance/performance.py
robinholzi: Iirc we only filter on the elevator side by specifying the first and last timestamp of the evaluation interval. As we currently don't use the evaluator at all, we can surely sample on the supervisor side. I also don't want to adjust the grpc interface again at this point for time reasons.

modyn/config/schema/pipeline/trigger/performance/performance.py
robinholzi: We don't use the accuracy results but extract the number of misclassifications from them (3rd decision criterion for performance aware triggers)

modyn/config/schema/pipeline/trigger/performance/criterion.py
robinholzi: already part of another open PR, I'll remove it.

modyn/config/schema/pipeline/trigger/performance/criterion.py
robinholzi: It doesn't make sense; probably a code pilot artifact😅

modyn/config/schema/pipeline/trigger/performance/performance.py
robinholzi: Okay, I'll open a separate documentation PR after the this here.

robinholzi: #596

modyn/config/schema/pipeline/trigger/performance/performance.py
MaxiBoether: Okay, maybe we can either add a comment here or only enforce this when actually usign this decision criterion, otherwise the changes LGTM

MaxiBoether (approved):
see my one comment on the accuracy stuff, otherwise LGTM

modyn/config/schema/pipeline/trigger/performance/performance.py
robinholzi: ** The accuracy information is only used for the misclassification counting, the thresholds-based approaches allow all kinds of user-defined metrics :)

PR #590, opened 2024-08-15
feat: Add PerformanceTrigger decision policies 

robinholzi:
# Motivation

As a second last step, we add the business logic for the decision policies that implement different decision criterion models.

depends on https://github.com/eth-easl/modyn/pull/589


modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -0,0 +1,202 @@
...
+
+        self.cumulated_avoidable_misclassifications += round(new_avoidable_misclassifications)
+
+        if mode == "hindsight":
+            return self.cumulated_avoidable_misclassifications >= self.config.avoidable_misclassification_threshold
+
+        if mode == "lookahead":
+            # past misclassifications already exceed the threshold, forecasting not needed
+            if self.cumulated_avoidable_misclassifications >= self.config.avoidable_misclassification_threshold:
+                return True
```
MaxiBoether, line 180: What I don't see yet is how we reset this after a trigger. Otherwise we will always keep triggering, won't we? (The threshold might increase but my intuition would be to reset the cumulation buffer after trigger)

MaxiBoether (changes requested):
I think I confused myself due to the "accuracy" vs "performance" naming. Maybe you could elaborate on this based on my comments/adjust, and then I will take a look again. Generally, the logic looked fine, I just want to make sure I understand what is happening

modyn/supervisor/internal/triggers/performance/data_density_tracker.py
```diff
@@ -0,0 +1,72 @@
...
+    def previous_batch_samples(self) -> int:
+        """Returns the number of samples in the last batch."""
+        assert len(self.batch_memory) > 0, "No data in memory, calibration needed."
+        return self.batch_memory[-1][0]
+
+    def needs_calibration(self) -> bool:
+        """Checks if the tracker has enough data for a forecast."""
+        return len(self.batch_memory) == 0
+
+    def forecast_density(self, method: ForecastingMethod = "ridge_regression") -> float:
```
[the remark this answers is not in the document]
robinholzi: Ah, sure, I thought I had already refactored this. I probably forgot this file :)

modyn/supervisor/internal/triggers/performance/data_density_tracker.py
```diff
@@ -0,0 +1,72 @@
...
+        num_seconds = (
+            data[-1][1] - self.previous_batch_end_time
+            if self.previous_batch_end_time is not None
+            else data[-1][1] - data[0][1]
+        )
+
+        self.batch_memory.append((len(data), num_seconds))
+        self.previous_batch_end_time = data[-1][1]
+
+    def previous_batch_samples(self) -> int:
```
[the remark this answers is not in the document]
robinholzi: `previous_batch_samples` (now `previous_batch_num_samples`) is used by other components (`DecisionPolicy`) and therefore requires an assertion. If the client hasn't had at least one information before the call to `previous_batch_num_samples`, he should see a descriptive error.
On the other hand, `previous_batch_end_time` (now `_previous_batch_end_time`) is only used internally. Therefore, we can set a default `None.` For `previous_batch_num_samples` we don't want to expose the `None` type externally.

modyn/supervisor/internal/triggers/performance/performance_tracker.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        num_samples: int,
+        num_misclassifications: int,
+        evaluation_scores: dict[str, float],
+    ) -> None:
+        """Informs the tracker about a new trigger and resets the memory."""
+        self.trigger_evaluation_memory.append((num_samples, num_misclassifications, evaluation_scores))
+
+        # first element in the new series is the performance right after trigger
+        self.since_last_trigger = [(num_samples, num_misclassifications, evaluation_scores)]
```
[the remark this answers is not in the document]
robinholzi: Not really, for the first evaluation after a trigger (reset of `since_last_trigger`), we add the value to both series (`trigger_evaluation_memory` and `since_last_trigger`). However, the two series will eventually contain different items. With every new trigger `trigger_evaluation_memory` will grow.  `since_last_trigger` only grows with every non triggering evaluation and is reset at a trigger. So, we cannot maintain the values in the same structure as they track two semantically different time series.

modyn/supervisor/internal/triggers/performance/decision_policy.py
robinholzi:
```python
    def inform_trigger(self) -> None:
        """Resets the cumulated avoidable misclassifications."""
        self.cumulated_avoidable_misclassifications = 0
```

modyn/supervisor/internal/triggers/performance/performance_tracker.py
```diff
@@ -0,0 +1,127 @@
...
+        """Informs the tracker about a new evaluation."""
+        self.since_last_trigger.append((num_samples, num_misclassifications, evaluation_scores))
+
+    def inform_trigger(
+        self,
+        num_samples: int,
+        num_misclassifications: int,
+        evaluation_scores: dict[str, float],
+    ) -> None:
+        """Informs the tracker about a new trigger and resets the memory."""
```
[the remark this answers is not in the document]
robinholzi: `since_last_trigger` is reset, `trigger_evaluation_memory` is just extended

modyn/supervisor/internal/triggers/performance/performance_tracker.py
```diff
@@ -0,0 +1,127 @@
...
+        self.trigger_evaluation_memory.append((num_samples, num_misclassifications, evaluation_scores))
+
+        # first element in the new series is the performance right after trigger
+        self.since_last_trigger = [(num_samples, num_misclassifications, evaluation_scores)]
+
+    def previous_batch_num_misclassifications(self) -> int:
+        """Returns the number of misclassifications in the previous batch."""
+        return self.since_last_trigger[-1][1]
+
+    def forecast_expected_accuracy(self, method: ForecastingMethod = "ridge_regression") -> float:
```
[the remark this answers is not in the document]
robinholzi: Ah I see, the times series forecasting method should not include `performance` in its name. Maybe `forecast_next_time_series_value`?

modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -0,0 +1,201 @@
...
+        performance_tracker: PerformanceTracker,
+        mode: TriggerEvaluationMode,
+        method: ForecastingMethod,
+    ) -> bool:
+        if mode == "hindsight":
+            return evaluation_scores[self.config.metric] < self.config.metric_threshold
+
+        return (evaluation_scores[self.config.metric] < self.config.metric_threshold) or (
+            performance_tracker.forecast_next_performance(mode) < self.config.metric_threshold
+        )
```
MaxiBoether, line 74: Should we log the reason for the decision / the forecasted value somewhere?

modyn/supervisor/internal/triggers/performance/performance_tracker.py
```diff
@@ -0,0 +1,134 @@
...
+
+        Returns:
+            The forecasted accuracy.
+        """
+        return forecast_next_time_series_value(
+            observations=[1 - p[1] / p[0] for p in self.trigger_evaluation_memory],
+            method=method,
+        )
+
+    def forecast_next_accuracy(self, method: ForecastingMethod = "ridge_regression") -> float:
```
MaxiBoether, line 87: I think my question about this is why we need it. We already enforce the user to use Accuracy, hence we could use  `forecast_expected_performance` with accuracy instead of "manually" calculating the accuracy here?

MaxiBoether (approved):
Thank you, the code is very readable now. I had one comment whether you want to log something (might make sense if you want to understand what happened), and I am still not sure why we need an extra function for accuracy forecasting

modyn/supervisor/internal/triggers/performance/performance_tracker.py
robinholzi: I see your point; generally, I'd prefer your solution. However, this assumption that we force the user to specify accuracy is on the configuration level. For this component, I'd rather keep it independent of this assumption and use stick with the duality of `performance` / `accuracy.`

modyn/supervisor/internal/triggers/performance/decision_policy.py
robinholzi: Good idea that might be useful! 

I cannot integrate this fully in this PR, though, as several of the follow-up PRs define code that also needs adjustment for this. If I run into problems when executing the policy, I'll come back to this and adjust the return type of `evaluate_decision` to `tuple[bool, str]`, outputting the decision reason.

PR #591, opened 2024-08-15
feat: Add `PerformanceTrigger`

robinholzi:
# Motivation

As a last step, we implement the trigger interface for the `PerformanceTrigger` and utilise the features of our trackers and decision policies.

depends on https://github.com/eth-easl/modyn/pull/590


modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -0,0 +1,302 @@
...
+class PerformanceTrigger(Trigger):
+    """Trigger based on the performance of the model.
+
+    We support a simple performance drift approach that compares the
+    most recent model performance with an expected performance value
+    that can be static or dynamic through a rolling average.
+
+    Additionally we support a regret based approach where the number of
+    avoidable misclassifications (misclassifications that could have
+    been avoided if we would have triggered) is compared to a threshold.
```
MaxiBoether, line 55: I wonder if we should somehow differentiate between these two cases. In the one case, we compare any metric (ROC/AUC, Accuracy, F1...) against a threshold. In the other case, we pick one particular metric (cumulative avoidable misclassifications) and compare that. The only thing that these two have in common is that we compare sth to a threshold. Should we maybe introduce two (sub)classes for these two cases? It could also not make sense, feel free to challenge this (as always), but I feel like we're putting two policies on different levels of abstraction into this class here.

modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -0,0 +1,302 @@
...
+
+            # Update the remaining data
+            processing_head_in_batch += len(next_detection_interval)
+            new_key_ts = new_key_ts[len(next_detection_interval) :]
+
+            # Reset for next detection
+            self._sample_left_until_detection = self.config.detection_interval_data_points
+
+            # Run the evaluation (even if we don't use the result, e.g. for the first forced trigger)
+            self.data_density.inform_data(next_detection_interval)
```
MaxiBoether, line 132: Do we always need to keep track of data density? In my understanding, we only need it when forecasting, don't we?

modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -0,0 +1,302 @@
...
+                # If we've never triggered before, always trigger
+                self._triggered_once = True
+                triggered = True
+
+                # in the first interval we don't evaluate the decision policies as they might require one trigger
+                # to have happened before in order to derive forecasts
+
+            else:
+                # evaluate the decision policies
+                any_policy_triggered = False
```
MaxiBoether, line 157: nit: i think you can just use triggered and say triggered |= policy_decisions[policy_name]

modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -0,0 +1,302 @@
...
+        assert self.most_recent_model_id is not None
+        assert self.dataloader_info is not None
+        assert self.model_downloader is not None
+        assert self.context and self.context.pipeline_config is not None
+
+        evaluation_dataloader = prepare_trigger_dataloader_fixed_keys(
+            self.dataloader_info, [key for key, _ in interval_data]
+        )
+
+        # Download most recent model as model manager
```
MaxiBoether, line 228: nit: manager has been renamed

MaxiBoether (changes requested):
The logic itself is very understandable and looks good. My major question is on the duality of using the avoidable misclassifcations vs some metric for triggering. WDYT?

modyn/supervisor/internal/triggers/performancetrigger.py
robinholzi: changes from earlier are not merged in, yet

modyn/supervisor/internal/triggers/performancetrigger.py
robinholzi: Yes, we need it iff at least one of the `DecisionCriterion` needs it. Otherwise, it's unused. However, it's quite cheap to keep track of this, so I'd rather always do it for less code complexity.

modyn/supervisor/internal/triggers/performancetrigger.py
robinholzi: I think the current setup is quite suitable, as it's similar to the drift setup where we observe different drift metrics, each having its own `DecisionEngine`. Here, we have different `DecisionEngine`s directly in the `PerformancePolicy`. Of course, we could force the user to split it up into different root-level triggers and use the soon-to-come `EnsembleTriggers` to aggregate them. However, those come with one major problem:

Like embedding generation, evaluations are expensive. They require us to maintain/download models, fetch samples, and run inferences. This is a large chunk of the runtime of these PerformanceTriggers, and it is shared across all `DecisionEngines` and their metrics. Therefore I propose to use those nested `DecisionCriterion` instances (for both avoidable misclassification as well as metric threshold approaches)

The subclass difference between the approaches is already reflected through the hierarchies on `DecisionPolicy` and `DecisionCriterion` (pydantic models) level.

modyn/supervisor/internal/triggers/performancetrigger.py
MaxiBoether: Okay, I was just thinking about the memory requirements of this, but as long as it's not a problem, we can leave it as is

modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -0,0 +1,300 @@
...
+            num_samples=num_samples,
+            num_misclassifications=num_misclassifications,
+            evaluation_scores=evaluation_scores,
+        )
+
+    # ---------------------------------------------------------------------------------------------------------------- #
+    #                                                     Internal                                                     #
+    # ---------------------------------------------------------------------------------------------------------------- #
+
+    def _run_evaluation(self, interval_data: list[tuple[int, int]]) -> tuple[int, int, dict[str, float]]:
```
MaxiBoether, line 215: This function seems to be missing a test

MaxiBoether (approved):
Thanks, LGTM. The only thing I noticed is that `_run_evaluation` seems to be not tested, you might want to add a simple test there to check this works as expected.

modyn/supervisor/internal/triggers/performancetrigger.py
robinholzi: didn't add a tests as the equivalent method of datadrifttrigger also didn't have one. I'll add one for this here :)

MaxiBoether: if it's much overhead, then i guess its fine if you see it working properly. we didnt have codecov at the time so probably I just often missed missing tests :D

robinholzi: Okay, I won't add tests for this function. This would result in more mocking than actual testing.

PR #593, opened 2024-08-18
feat: Add `DynamicRollingAverageThresholdPolicy`

robinholzi:
# Motivation

Alongside the percentile-based dynamic threshold calibration, we also want to try out and offer a rolling average approach that checks for the deviation between the current distance and the rolling average and see if it exceeds a certain (relative/absolute) threshold.


robinholzi:
@MaxiBoether This should also be fairly straightforward, just another variation of dynamic thresholds

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -41,27 +41,58 @@ def evaluate_decision(self, distance: float) -> bool:
...
     """Decision policy that will make the binary is_drift decisions based on a
-    dynamic threshold.
+    dynamic threshold."""
+
+    def __init__(self, config: DynamicThresholdCriterion):
+        self.config = config
+        self.score_observations: deque = deque(maxlen=self.config.window_size)
+
+
+class DynamicPercentileThresholdPolicy(DynamicDecisionPolicy):
```
robinholzi, line 51: I also adjusted this decision policy to align with my (unpushed) implementation in my offline drift exploration. Seems more bulletproof and easier to understand.

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -41,27 +41,58 @@ def evaluate_decision(self, distance: float) -> bool:
...
+                ),
+                len(sorted_observations) - 1,
+            )
+        ]  # - 1 from length to index space
+        self.score_observations.append(distance)
+
+        return distance > threshold
+
+
+class DynamicRollingAverageThresholdPolicy(DynamicDecisionPolicy):
```
robinholzi, line 82: new

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -41,27 +41,58 @@ def evaluate_decision(self, distance: float) -> bool:
...
+    def evaluate_decision(self, distance: float) -> bool:
+        if len(self.score_observations) == 0:
+            self.score_observations.append(distance)
+            return True
+
+        sorted_observations = list(sorted(self.score_observations))
+
+        threshold = sorted_observations[
+            min(
+                max(
```
MaxiBoether, line 70: I don't really see why the min/max will be needed if basically you want the nth element in a sorted list, but i guess the robustness cannot hurt

modyn/supervisor/internal/triggers/drift/decision_policy.py
```diff
@@ -41,27 +41,58 @@ def evaluate_decision(self, distance: float) -> bool:
...
+                    0,
+                    int(len(sorted_observations) * (1.0 - self.config.percentile)) - 1,
```
MaxiBoether, line 72: why len() * perc - 1 instead of just rounding?

MaxiBoether (approved):
LGTM, although I could not immediately understand why the min(max) stuff can occur

modyn/supervisor/internal/triggers/drift/decision_policy.py
robinholzi: we compute an index here, they `-1` should just account for the fact the python indexes are 0-based.

I had the following semantics in mind:
- Assume we have 10 observations ready
- user requests 0% significance: `len(sorted_observations) * (1.0 - self.config.percentile) = 10`
- user requests 10% significance: --> `9`
- with the `-1` we get valid indexes

What happens to the cases in between (e.g. `5%`) is a design decision. Rounding is actually desired here! The `-1` is still needed, though.

PR #594, opened 2024-08-18
fix: Add `nltk` dependency

robinholzi:
CI started failing because of missing `nltk`. I suspect that some of our packages remove transitive dependency to `nltk` or introduced this new dependency without recording it in the package dependencies.


MaxiBoether (approved):
Thanks!

PR #595, opened 2024-08-18
chore: Adjustments for triggering experiments

robinholzi:
# Bugfixes and Experiment configs


modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -97,13 +97,21 @@ def evaluate_decision(
         mode: TriggerEvaluationMode,
         method: ForecastingMethod,
     ) -> bool:
-        threshold = performance_tracker.forecast_expected_performance(mode) - self.config.allowed_deviation
+        expected = performance_tracker.forecast_expected_performance(mode)
```
MaxiBoether, line 100: as discussed, this call is invalid

MaxiBoether: what i dont understand still is why in the hindsight mode we need any forecasting (let it be the rolling average). i understood it as "if our accuracy on the evaluation interval deviates too much, we trigger". but apparently something more is going on?

modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -97,13 +97,21 @@ def evaluate_decision(
...
+            if self.config.absolute:
+                return deviation >= self.config.deviation
+            return deviation >= self.config.deviation * expected
 
-        return (evaluation_scores[self.config.metric] < threshold) or (
-            performance_tracker.forecast_next_performance(mode) < threshold
+        if self.config.absolute:
+            allowed_absolute_deviation = self.config.deviation
+        else:
+            allowed_absolute_deviation = self.config.deviation * expected
```
MaxiBoether, line 111: you could also always calculate this and then use that in hindsight mode as well. might simplify this function a bit

modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -97,13 +97,21 @@ def evaluate_decision(
...
+
+        return (deviation >= allowed_absolute_deviation) or (
```
MaxiBoether, line 113: i guess you could also remove the "if hindsight" codepath, just have this first thing (which is basically hindsight), and then "if lookahead" set another bool to the result of this (or false if not lookahead). reduces the branches here, had some trouble parsing it at first

MaxiBoether (approved):
Thanks, changes as mostly discussed :)

modyn/supervisor/internal/triggers/performance/decision_policy.py
robinholzi: With 
```python
forecast_expected_performance(
            mode, method="rolling_average"
        )
```
this is equivalent to saying, if current values deviates more than X % of the rolling average, then trigger. We simply just reuse the averaging from the forecasting method. It's not a true forecast.

MaxiBoether: as discussde, let's rename expected to "optimal" performance for this

PR #596, opened 2024-08-19
docs: Document `PerformanceTrigger` approach

robinholzi:
# Motivation

Add an outline that describes how our `PerformanceTriggers` work.


modyn/tests/supervisor/internal/triggers/test_datadrifttrigger.py
```diff
@@ -13,13 +13,13 @@
     AlibiDetectMmdDriftMetric,
 )
 from modyn.config.schema.pipeline.trigger.drift.config import AmountWindowingStrategy
-from modyn.config.schema.pipeline.trigger.drift.detection_window import (
-    TimeWindowingStrategy,
-)
-from modyn.config.schema.pipeline.trigger.drift.metric import (
```
MaxiBoether, line 19: i feel like your ruff or whatever has a stricter line limit than the repo because this keeps changing despite the CI running through :D

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
+# Performance Triggering
+
+## Overview
+
+The `PerformanceTrigger` is a trigger mechanism that evaluates the performance of a model over time and makes decisions based on performance-observing criteria. Unlike simple triggers, the `PerformanceTrigger` considers user selected performance metrics and can evaluate decisions based on static or dynamic performance thresholds, as well as the number of avoidable misclassifications.
+
+We define `avoidable misclassifications` as the misclassifications that could have been avoided if the model had triggered earlier. They can be see as the lost benefits of not triggering. We derive the avoidable misclassification estimates by observing two time series:
+
+1. The series of evaluation scores directly after the last executed triggers (with fresh models). We define those results as `expected performance`.
```
MaxiBoether, line 9: What does "directly" mean exactly? One metric we get? Do you stop after e.g. 2 evaluations? How is this defined?

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
...
+        +float forecast_expected_accuracy(method)
+        +float forecast_next_accuracy(method)
+        +float forecast_expected_performance(metric, method)
+        +float forecast_next_performance(metric, method)
+    }
+```
+
+### `PerformanceTriggerCriterion` Hierarchy
+
+The PerformanceTriggerCriterion class is an abstract configuration base class for criteria like `StaticPerformanceThresholdCriterion` and `DynamicPerformanceThresholdCriterion`, which define how decisions are made based on drift metrics. Even if the threshold generation is somewhat different to the DecisionCriteria in the drift triggers, the concept is similar. The configuration made in those models will determine the decision process within the `PerformanceDecisionPolicy` classes.
```
MaxiBoether, line 88: drift metrics?

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
...
+
+We define `avoidable misclassifications` as the misclassifications that could have been avoided if the model had triggered earlier. They can be see as the lost benefits of not triggering. We derive the avoidable misclassification estimates by observing two time series:
+
+1. The series of evaluation scores directly after the last executed triggers (with fresh models). We define those results as `expected performance`.
+
+2. The series of evaluation SINCE the last trigger on policy update points where we could have triggered but decided not to. This series is used to estimate the avoidable misclassifications. We define those results as `actual/effective performance`.
+
+The delta between the expected and the actual performance is then used to estimate the avoidable misclassifications.
+
+This mechanism relies on performance and data density tracking to observe trends. Decision policies then make binary decisions (trigger or not) based on this information.
```
MaxiBoether, line 15: data density occurs here the first time. i think this is still confusing me a bit because i don't know exactly what it is in this context and how you use it.

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
...
+
+For the performance and misclassification estimations we can always take two approaches: `hindsight` and `lookahead`. The `hindsight` approach uses the most recent data to estimate the performance and misclassifications. The `lookahead` approach uses the forecasted performance and misclassifications to make decisions.
```
MaxiBoether, line 17: The text above this paragraph reads like you always go for hindsight. you keep track of performance, and calculate a delta to the performance after a trigger. I don't understand yet exactly how the same setting also works for the lookahead approach, since you use the existing data to calculate a performance delta. this to me sounds like a hindsight approach, not sure where the forcast comes in

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
...
+
+1. **StaticPerformanceThresholdDecisionPolicy**: Triggers when the performance metric falls below a static threshold.
+2. **DynamicPerformanceThresholdDecisionPolicy**: Triggers when the performance metric deviates beyond a dynamic threshold, based on a rolling average.
+3. **StaticNumberAvoidableMisclassificationDecisionPolicy**: Triggers when the number of avoidable misclassifications exceeds a predefined static threshold.
+
+<details>
+<summary><b>What are avoidable misclassification?</b></summary>
+
+Avoidable misclassifications are the misclassifications that could have been avoided if the model had triggered earlier. They can be seen as the lost benefits of not triggering. Every evaluation point we estimate the avoidable misclassifications by comparing the expected performance (the performance of the model if we had triggered) with the actual performance (the performance of the model since the last trigger). Avoidable misclassifications are then derived from the difference between the expected and actual accuracy and the estimated query density.
```
MaxiBoether, line 138: This makes a lot of sense to me, but only in the hindsight setting

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
```diff
@@ -0,0 +1,180 @@
...
+
+In hindsight mode, decisions are based solely on the current observed performance metrics. This mode checks if the most recent performance falls below a predefined threshold or deviates significantly from the expected performance, based on past evaluations. It is a retrospective approach, relying on already observed data to determine if a trigger should be invoked. Triggering happens post-factum when the performance has already degraded / a threshold has been crossed.
+
+On the other hand, forecasting mode looks forward and incorporates predictions of future performance into the decision-making process. This mode not only evaluates the current performance but also forecasts future performance and data density to anticipate potential issues before they occur. For example, in the static number avoidable misclassification policy, forecasting mode predicts whether the cumulative number of avoidable misclassifications will exceed the threshold before the next evaluation. By doing so, it aims to prevent performance degradation by triggering preemptively, rather than waiting for the problem to manifest.
```
MaxiBoether, line 148: I guess what is unclear to me is a) the role of data density here (what is it, what is it needed for) and 2) if I understand correctly, you predict the number of avoidable missclassifications? If this is the case, maybe we should make clear that's its potentially difficult to predict sudden drift? because i was wondering how e.g. the drift in yearbook would have been predicted

MaxiBoether (changes requested):
Thank you! I mostly had questions on data density / forecasting mode, the rest LGTM!

modyn/tests/supervisor/internal/triggers/test_datadrifttrigger.py
robinholzi: That's right; it also changes a couple of other things on auto-save, but those will be reverted by the pre-commit hooks. The only thing that stays in the shorter lines iff a trailing comma is used. So far, it hasn't bothered me enough to investigate it. All my other code bases have shorter line length settings😅

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
robinholzi: Copy-paste issue😅

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
robinholzi: 1) see other threads; I'll add more background info for this.
2) That's true; it's not possible to forecast sudden drift with this approach. Not sure if it's possible with any approach at all. The whole idea is inspired by slowly/gradually drifting datasets like we see in arxiv. In those instances, we would never see sudden drift but build a cumulative "regret" metric (which is the number of misclassifications here). For the forecasted variant, we use the same expected performance (that is derived from the series of evaluations directly after triggers) but use a forecasted accuracy instead of the last accuracy measurement on the last batch. The forecasted accuracy is currently only a linear extrapolation through a ridge regression model (more sophisticated time series forecasting methods are possible obviously)

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
robinholzi: `directly` in the sense of `right after`. Basically in the `inform_new_model` call of the trigger. This is done after every trigger, we DO NOT stop this after a certain number of evaluations.

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
robinholzi: The idea of the `lookahead` mode isn't that different from the `hindsight` approach. I went for this first because the backward-looking definition is easier to understand. After understanding that the step to forecasting isn't that far anymore.

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
MaxiBoether: Okay, it might make sense to just say "evaluate the fresh model on the last evaluation interval", if I understand correctly that is what's happening?

MaxiBoether (approved):
Thank you, this LGTM!

docs/pipeline/triggering/PERFORMANCE_TRIGGERS.md
MaxiBoether: That makes sense. I wonder if something like a cumulative regret could also be built for drift (or if in drift, one could check out the drift w.r.t. some older data) but yeah thanks for clarifying the density.

robinholzi: For the number of misclassifications this cumulative approach feels rather natural. For drift I'm not sure if it makes sense to remember drift areas that have already been passed. I thought of this in the `CostTrigger` domain, where you have a regret metric (maybe performance, maybe drift, maybe data amount based incorporation latency) and a cost estimate (training time). However, for drift this would almost be impossible to configure for a user. We need this conversion rate hyperparameter from the user. And I don't see any semantic connection / heuristic a user could employ to derive a relation between an arbitrary drift distance sum and training time.

PR #597, opened 2024-08-19
feat: Implement `EnsembleTrigger`

robinholzi:
# Motivation

As we have had the `EnsebleTriggerConfig` for quite some time, I have already fully integrated it.

# Note:

Support for the new (unmerged) trigger policies `PerformanceTrigger` and `CostTrigger` must still be added.


modyn/config/schema/pipeline/trigger/ensemble.py
```diff
@@ -51,15 +58,28 @@ def aggregate_decision_func(self) -> Callable[[dict[str, bool]], bool]:
...
 
-    policies: dict[str, __TriggerConfig] = Field(  # type: ignore[valid-type]
+    detection_interval_data_points: int = Field(
+        1000,
+        description=(
+            "The number of samples in the interval after which drift detection is performed. "
+            "Some policies don't require this time-dicretization. However some do, e.g. therefore, this parameter "
+            "is required also here."
+        ),
+        ge=1,
```
MaxiBoether, line 78: I think here i am unsure why we need it on the ensemble level. If the time trigger triggers more frequently than e.g. the drift trigger, and we do majority vote, it is what it is. Why can't the drift trigger buffer this and run the detections, and if it has not seen enough data, just return False? I don't currently see why it's the job of the ensemble to care about this,  instead of just forwarding it to the sub triggers

MaxiBoether (changes requested):

modyn/config/schema/pipeline/trigger/ensemble.py
robinholzi: Let's maybe discuss that in person!

robinholzi: Okay, I decided to go one step further and implement a more dynamic approach without ensemble level batching. I hope that's closer to what you imagined?

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+class EnsembleTrigger(Trigger):
+    """Evaluates multiple subtriggers and triggers if a certain after an
+    aggregation function aggregates subtrigger decisions to True.
+
+    Idea:
+        The EnsembleTrigger always evaluates and caches the next triggering index
+        from the generators of each subtrigger. The evaluations of all subsequent subtrigger indexes are
+        deferred in a lazy manner.
+
+        We assume that after a subtrigger will always like to see a trigger after the next index in the generator.
```
MaxiBoether, line 34: I don't think this sentence is grammatically correct? I at least don't understand it :D

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+            trigger_name: trigger.inform(new_data, log) for trigger_name, trigger in self.subtriggers.items()
+        }
+
+        # Indicates whether the ensemble trigger has passed a triggering index
+        # with processing_head for a subtrigger since the last trigger.
+        # Will be reset after every trigger.
+        subtrigger_decision_cache = dict(
+            {trigger_name: False for trigger_name in self.subtriggers},
+            **self.last_inform_decisions,
+        )
```
MaxiBoether, line 76: so this is basically a defaultdict with False, and you merge last_inform_decisions into that defaultdict? not sure whether using a defaultdict explicitly here will make the code more compact or not

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+                raise NotImplementedError("Nested ensemble triggers are not supported.")
+            else:
+                raise ValueError(f"Unknown trigger config: {trigger_config}")
+        return subtriggers
+
+    def _reset_subtrigger_decision_cache(self, subtrigger_decision_cache: dict[str, bool]) -> None:
+        subtrigger_decision_cache.clear()
+        subtrigger_decision_cache.update({trigger_name: False for trigger_name in self.subtriggers})
+
+    def _update_outdated_next_trigger_indexes(
```
MaxiBoether, line 148: maybe: `_update_next_subtrigger_index_cache` to make more clear what is being updated?

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+        self,
+        processing_head: int,
+        subtrigger_generators: dict[str, Generator[int, None, None]],
+        next_subtrigger_index_cache: dict[str, int | None],
+    ) -> None:
+        """Updates the next trigger indexes in the cache to be in the future
+        w.r.t. the current trigger index.
```
MaxiBoether, line 155: w.r.t. the processing head?

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+        Supposed to be called after a new trigger index was found or at the beginning of the inform method to
+        sync. with the new data batch.
+
+        Note: Arguments are modified in place.
+        """
+        for trigger_name, generator in subtrigger_generators.items():
+            # generator is not exhausted
+            if next_subtrigger_index_cache[trigger_name] is not None:
+                try:
+                    while cast(dict[str, int], next_subtrigger_index_cache)[trigger_name] <= processing_head:
```
MaxiBoether, line 166: maybe you can remove the casts by instead of using None use -2 (or two variables like NO_TRIGGER_YET = -1 and EMPTY_GENERATOR = -2)?

modyn/supervisor/internal/triggers/ensembletrigger.py
```diff
@@ -0,0 +1,230 @@
...
+            next_trigger_idx = self._find_next_trigger_index(
+                processing_head,
+                new_data,
+                subtrigger_decision_cache,
+                next_subtrigger_index_cache,
+                log,
+            )
+            if next_trigger_idx is None:
+                # complete remaining inform calls regardless of the decision, any of the decision will lead to a
+                # non triggering aggregation result
```
MaxiBoether, line 105: I don't understand this - how do we complete something here, if we immediately break out of the loop?

modyn/tests/supervisor/internal/triggers/performance/test_ensembletrigger.py
```diff
@@ -0,0 +1,270 @@
...
+        subtrigger_decision_cache=subtrigger_decision_cache,
+        next_subtrigger_index_cache=next_subtrigger_index_cache,
+        log=trigger_log,
+    )
+    assert result is None
+    assert len(trigger_log.evaluations) == 0
+
+
+def test_inform_trigger(ensemble_trigger_config: EnsembleTriggerConfig) -> None:
+    trigger = EnsembleTrigger(ensemble_trigger_config)
```
MaxiBoether, line 204: I have a hard time parsing this test, but basically I am trying to understand whether this also tests the state across inform calls. Do we have a test case that checks whether storing the old decisions (that some trigger is `True` already when starting inform) works as expected? If not, can you add one? Alternatively, you could also extend the `test_find_next_trigger_index` above maybe with a dict that already contains true

MaxiBoether (approved):
Thank you. I had a few small comments. I didn't realize this needs some caching logic because in mind I only had en ensemble that triggers if any subtrigger triggers but yeah with an arbitrary strategy such as min n we need this sophisticated logic indeed.

modyn/tests/supervisor/internal/triggers/performance/test_ensembletrigger.py
robinholzi: For the first part, yes `test_inform_multi_batch` should test exactly that. For the latter, yes, it makes sense to explicitly tests that in `find_next_trigger_index`

modyn/supervisor/internal/triggers/ensembletrigger.py
robinholzi: Actually, I forgot to implement what I was describing with this comment. We always have to let all subtrigger see all data as they need it for their internal state. After a non trigger and after a trigger we have to consume all generator indexes up to the newest processing head.

modyn/supervisor/internal/triggers/ensembletrigger.py
robinholzi: Yep, `-2` sounds good. Mypy is simply too stupid to derive the strongest postcondition for the guard clauses.

PR #598, opened 2024-08-20
feat: Add config models for `CostTriggers`

robinholzi:
# Motivation

Based on the ideas of https://arxiv.org/pdf/2310.04216 and https://cse.hkust.edu.hk/~weiwa/papers/continuum-socc18.pdf, we want to also evaluate the effectiveness of cost-aware triggering by combining it with our misclassification based `PerformanceTrigger`s.

Introducing the configuration space here.


modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -0,0 +1,74 @@
...
+    """Base class for cost aware trigger policies."""
+
+    cost_tracking_window_size: int = Field(
+        description="How many trigger into the past should be considered for the linear cost w.r.t. data amount model."
+    )
+
+
+class DataIncorporationLatencyCostTriggerConfig(_CostTriggerConfig):
+    """Cost aware trigger policy configuration that uses the data incorporation
+    latency as a regret metric.
```
MaxiBoether, line 25: I assume this latency describes how many seconds/minutes/hours a data has not been trained on, right? Do you base this off the timestamp of the data point (which could be in 1970) or based on the current timestamp at the supervisor?

modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -0,0 +1,74 @@
...
+
+    While no trigger occurs the un-integrated data is cumulated and the sum of all seconds of every unincorporated data
```
MaxiBoether, line 27: I don't get that sentence. I understand you cumulate the latencies, but how do you also cumulate the data? Or is this redundant?

modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -0,0 +1,74 @@
...
+    point is added up.
+
+    `incorporation_delay_per_training_second` servers as a conversion rate between budget (training time) and regret
```
MaxiBoether, line 30: I don't understand this, unfortunately. Is this our threshold that our cumulative metriic may not cross? I don't understand yet how you can convert between training time and incorporation latency.

modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -0,0 +1,74 @@
...
+    metric (incorporation latency).
+
+    When the regret metric exceeds the training-time based budget metric, a trigger is fired.
```
MaxiBoether, line 33: Sorry, I might just be remembering wrongly - a cost/accuracy tradeoff is not happening here, only time based? Is that exactly what you were trying to tell me? Maybe we should also document this., if that is true.

modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -0,0 +1,74 @@
...
+        description=(
+            "How many seconds of samples having spent in the incorporation queue are we willing to accept "
+            "per second of training time saved."
+        )
+    )
+
+
+class AvoidableMisclassificationCostTriggerConfig(_CostTriggerConfig, _InternalPerformanceTriggerConfig):
+    """Cost aware trigger policy configuration that using the number of
+    avoidable misclassifications integration latency as a regret metric.
```
MaxiBoether, line 49: What I don't understand is whether your regret here is just the number of avoidable misclassificaitons (cumulated) or something that also takes into account time (in seconds or sth) because you say latency

MaxiBoether (changes requested):
Thank you! I had a couple of questions on this one

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: It's based on the timestamps of the batches (first and last sample) at the policy. The supervisor time is not used at all.

However, your comment made me think of the sample timestamps. Currently, I just assumed that if, e.g., the duration of the last batch is 10s and there were 50 samples, then 50*10s of latency will be added for the last interval (+ the additional latency for all data before the last interval). However, we could use the fine-grain time resolution of every sample to, e.g., let samples at the end of the batch have a smaller contribution than the earlier ones. It's a bit more painful to implement this but probably desireable, wdyt? If you agree, I'll adjust the other PR in that regard.

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: `DataIncorporationLatencyCostTriggerConfig` is only cost-aware but not performance-aware. 
The policy is inspired by https://cse.hkust.edu.hk/~weiwa/papers/continuum-socc18.pdf (p. 7; 4.3).

It's time-sensitive and sensitive to data/query density. If there are e.g. periods where almost no data comes in,  triggers are avoided. In some sense, this is a combination of time- and amount-based approaches. 

This isn't too beneficial for our evaluation setting as we weight our evaluations by time and not by data amount.

I expect more from the second cost policy that includes the misclassifiations.

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: I suspect the cumulated number of misclassifications is very unstable and badly conditioned on user input. That's why we penalize every second that an avoidable misclassification remains unaddressed (no trigger). It's a result of combining this data incorporation latency idea with the static number of misclassification performance trigger.

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: From https://cse.hkust.edu.hk/~weiwa/papers/continuum-socc18.pdf (p. 7):

> we propose a cost-aware policy for fast data incorporation at low training cost. We model the cost sensitivity of a user with a _**“knob” parameter w**_

`incorporation_delay_per_training_second` reflects this "knob parameter w".

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: From the samples, we can build an unintegrated samples over a time curve.
The area under this curve is the (cumulated) latency.

modyn/config/schema/pipeline/trigger/cost/cost.py
MaxiBoether: I guess I am not fully sure yet what you mean by "duration of a batch". Is that max timestamp - min timestamp in the batch? But I mean yeah each sample has its own latency, this is nothing on a batch level. I understand it as total latency = \sum_{all not yet trained samples} sample latency, and sample latency = max timestamp observerd so far - sample timestamp.  But I might be missing something in this interpretation

modyn/config/schema/pipeline/trigger/cost/cost.py
MaxiBoether: Okay, since we're discrete I am just thinking of it as a summation over all individual latencies, but I am not sure whether this is true.

modyn/config/schema/pipeline/trigger/cost/cost.py
MaxiBoether: The way I understand it, `DataIncorporationLatencyCostTriggerConfig` triggers when the accumulated regret (sum over all latencies as I mentioned in another comment) crosses a threshold. But I start to believe I am wrong in this assumption. I don't understand yet what the role of this w in the Modyn setting is. Because we say the training cost is linearily dependend on the number of samples, I still don't get the role training time plays here, and why we need to convert between training time and this latency.

modyn/config/schema/pipeline/trigger/cost/cost.py
MaxiBoether: Okay, maybe you could add exactly this explanation to the docstring. " avoidable misclassifications integration latency" contains this but honestly this comment helps much in understanding it.

MaxiBoether: However, if this is cost-performance, and data incorporation is cost-only, maybe we should not keep them in the same file, but instead have a file for cost-only triggers and one for cost-performance aware triggers?

robinholzi: I think for the config models this is fine, cost is the overarching theme for both of this flavours. The respective trigger will both be subclasses of `CostTrigger` but within their own files.

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: Yes, instead a computationally and memory-wise more expensive alternative would be to store all samples since the last trigger and on every evaluation compute the time distance (latency) to the current time and build the sum of these latencies. The approach we take here is iterative and doesn't require remembering the timestamps of the elements.

modyn/config/schema/pipeline/trigger/cost/cost.py
robinholzi: with batch i refer to the data intervals established through the `detection_interval` config variable (and not the supervisor batches)

MaxiBoether: my comment was more about the duration of a batch, but as discussed that is max timestamp - end timestamp of last batch.

MaxiBoether (approved):

PR #599, opened 2024-08-20
feat: Cost tracker for `CostTriggers`

robinholzi:
# Motivation

We want to use a linear model ($T(D) :=aD+b$, $T :=$ time, $D :=$ samples) to estimate training costs/time for our cost-aware policies.

`CostTracker` keeps track of previous training costs and provides the model prediction functionality.


modyn/supervisor/internal/triggers/cost/cost_tracker.py
```diff
@@ -0,0 +1,44 @@
...
+        self._linear_model.fit(samples, observations)
+
+    def needs_calibration(self) -> bool:
+        """Checks if the tracker has enough data for a forecast.
+
+        After one trigger inform, the tracker forecasts with a constant
+        model. With the second trigger the model can learn the
+        y-intercept and start making meaningful forecasts.
+        """
+        return len(self.measurements) == 0 or not hasattr(self._linear_model, "coef_")
```
MaxiBoether, line 39: can the second condition ever happen if the first is not true?

modyn/supervisor/internal/triggers/cost/cost_tracker.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def forecast_training_time(self, number_samples: int) -> float:
+        """Forecasts the training time for a given number of samples."""
+        assert not self.needs_calibration(), "The tracker needs more data to make a forecast."
+        return self._linear_model.predict(np.array([[number_samples]]))[0]
```
MaxiBoether, line 44: i suppose there is a reason why we need the additional dimensionality here, insetad of just having np.array([number_samples])

MaxiBoether (approved):
Thanks :)

modyn/supervisor/internal/triggers/cost/cost_tracker.py
robinholzi: yes, scikit learn demands this shape for one dimensional input vectors

modyn/supervisor/internal/triggers/cost/cost_tracker.py
robinholzi: No, they should be identical. I added both intentionally to signal this to the user and have some fallback in case of unforeseen inconsistent states.

PR #600, opened 2024-08-20
docs: Add docs for `CostTrigger`

robinholzi:
# Motivation

Lays out the architecture and core functionality of our new cost triggers

Also updating all other triggering docs along the way alongside some renamings to establish consistent naming


MaxiBoether:
should I also review this already?

robinholzi:
> should I also review this already?

@MaxiBoether I will most likely revisit this after all other cost-related PRs are merged (because of some renamings, etc.). However, you can take a rough first look. The conversion rate description should probably also be extended here.

robinholzi:
@MaxiBoether Should be ready, but it isn't time-critical.

docs/pipeline/triggers/OVERVIEW.md
```diff
@@ -0,0 +1,101 @@
+# Modyn Triggering
+
+Alongside the simple triggers, Modyn also provides complex triggers that can be used to trigger the training of a model.
+
+Despite the line being blurry, complex triggers are generally more sophisticated and require more information to be provided to the trigger - they most of the time cannot be pre-determined via a simple configuration entry but require
+reingested information from the ongoing pipeline run.
+
+Some complex triggers can only make decisions in batched intervals as they are cannot be efficiently computed
+on a per-sample basis. Here we can find the `DataDrift` and `Cost` based triggers.
```
MaxiBoether, line 9: (I know this just got moved in this PR) Maybe we can add one sentence like: The input to the trigger is already batched by the supervisor, this batching is done for performance-reasons. The `BatchedTriggers` operate on a separate level of batching to decouple this (?).

IDK, actually I am not sure why we did not use the batching parameter from the supervisor in the first place, since both batchings are somewhat for performance reasons

docs/pipeline/triggers/COST_TRIGGER.md
```diff
@@ -0,0 +1,177 @@
...
+
+    style PerformanceTriggerMixin fill:#DDDDDD,stroke:#A9A9A9,stroke-width:2px
+
+    style DataIncorporationLatencyCostTrigger fill:#C5DEB8,stroke:#A9A9A9,stroke-width:2px
+    style AvoidableMisclassificationCostTrigger fill:#C5DEB8,stroke:#A9A9A9,stroke-width:2px
+```
+
+### `CostTrigger` Hierarchy
+
+Both `DataIncorporationLatencyCostTrigger` and `AvoidableMisclassificationCostTrigger` track the cost of triggering and convert a regret metric (e.g., data incorporation latency or avoidable misclassification latency) into the training time unit with a user-defined conversion factor.
```
MaxiBoether, line 93: while data incorporation latency is intuitive (and explained below), I don't have an intuitive understanding of avoidable misclassification latency

docs/pipeline/triggers/OVERVIEW.md
robinholzi: I'll add the comment that I think the two levels of batching make sense, at least in the experimental pipeline mode.
The supervisor-level batching is mostly motivated by the memory constraints of the host system. As we have quite a lot of storage available on the sgs-gpu machines, we can easily set these batches to a high number and avoid extra loops through the supervisor logic.
The batching on trigger levels is dependent on the policy type and motivated by both performance reasons (we cannot detect drift on every sample as detection is rather expensive), and semantic constraints (detecting drift on very small data batches or evaluating on a really small number of samples doesn't make sense)

docs/pipeline/triggers/COST_TRIGGER.md
robinholzi: Conceptually, it's not that different from data incorporation latency. It's simply the time that misclassified samples stayed unaddressed through retraining. 

I would even argue that it makes more sense than data incorporation latency as misclassifications are eventually the metric that should incentivize retraining, so we don't want them to stay "unaddressed" for too long. The regular number of samples is rather unhelpful. For a stationary dataset, the model's predictive capabilities could stay constant while data incorporation latency cumulates up, leading to another **unnecessary** trigger. But let's see what the experiments will result in.

MaxiBoether (approved):
LGTM!

PR #601, opened 2024-08-20
feat: Add `CostTrigger`

robinholzi:
# Motivation

Following up https://github.com/eth-easl/modyn/pull/599, https://github.com/eth-easl/modyn/pull/602, https://github.com/eth-easl/modyn/pull/604 we want to implement a CostTrigger base class that reflects the tradeoff-based trigger between a cost and a regret metric.

Furthermore, we add a data incorporation latency-based cost trigger as the first variant (subclass) of `CostTrigger`.

https://github.com/eth-easl/modyn/pull/605 will add a second subclass based on misclassifications

PR #602, opened 2024-08-20
feat: Add `IncorporationLatencyTracker`

robinholzi:
# Motivation

We want a class that maintains a regret metric and computes a cumulative regret metric with an area under-curve approach.


robinholzi:
@MaxiBoether This is what we just talked about in person, feel free to take a look and check whether it matches my verbal description :)

robinholzi:
Taken from https://github.com/eth-easl/modyn/pull/601/files:
```python
# Updates
batch_start = self._previous_batch_end_time or next_detection_interval[0][1]
batch_duration = next_detection_interval[-1][1] - batch_start
```

modyn/supervisor/internal/triggers/cost/incorporation_latency_tracker.py
```diff
@@ -0,0 +1,80 @@
...
+        regrets_durations = [(end_timestamp - timestamp, regret) for timestamp, regret in regrets]
+        new_regret = sum(regret for _, regret in regrets)
+
+        # rectangular shape of the area under the curve for the recently arrived regrets
+        new_regret_latency = sum(duration * regret for duration, regret in regrets_durations)
+
+        # old regret units that still contribute to area under curve (rectangular shape)
+        old_regret_latency = self._current_regret * period_duration
+
+        self._current_regret += new_regret
```
MaxiBoether, line 72: I am trying to understand the meaning of the two local variables but I don't understand them yet.

My understanding is the following: If we have 10 samples since the last trigger, and now see a batch that has a duration of 20, we would want to calculate the regret of the batch and add 10 * 20 to it, because every sample since the last trigger gets an additional 20 timeunits latency.

In that sense, `self._current_regret` should keep track of the number of seen samples. It should be 10, since it does not matter what the regret value within that batch was. It should only matter that all 10 samples get another 20 timeunits regret when the next batch comes. But since new_regret is based on the timestamps, I don't see how this is possible. Might be easier to discuss this in person on the whiteboard.

robinholzi: - `self._current_regret` reflects the most recent y value of the `number of samples since last trigger` curve. 
- `self._cumulative_latency_regret` reflects the area under this curve since the last trigger.

I hope this image helps
image

MaxiBoether: "self._current_regret reflects the most recent y value of the number of samples since last trigger curve."

That would be what I expect to see here. In my example, that would be 10 (because we already saw 10 samples, and each of them gets an additional regret of 20). However, in line 72, we do `self._current_regret += new_regret`. And new_regret is not equal to the number of samples, but instead a per-sample regret value.

robinholzi: in the `data-incorporation-latency-cost-trigger` the regret is `1` for every sample, the time is only relevent for the `regret_latency`. I guess this definition of `regret` is a bit unintuitive as one might expect that regret already captures the time dimension. If you have a better name for the abstraction from data amount (or number of misclassifications in the second policy), I'm happy to adopt that.

MaxiBoether: Hm, what would be the value of this in case of number of misclassifications? Because for these, we don't have a latency, right? It's just a linear curve? Hence, I would think in that case, regret would also be 1, and we should not have `old_regret_latency = self._current_regret * period_duration`. I am basically unsure in what other cases despite the latency that we already discussed this latency tracker will be used and what other values than 1 this can take and why

robinholzi: I suspect that only using misclassification would yield the same problems as in the data incorporation latency case. Therefore, I was also planning to take this latency sum approach. For the base regret value, we I'd therfore use number of misclassifications and area under curve would be misclassification integration latency.
I'm not 100% sure yet how I will implement that on the trigger side. As we don't have access to what exact samples were misclassified (not planning to refactor that in the evaluation logic) I will either use the `add_latency(...)` function which only takes the aggregated regret for the last interval or I'll use `add_latencies` like before with a `regret` value of `expected_misclassification_rate` for every sample (instead of the `1` in the base case)

robinholzi: Overall, I think the current interface gives most flexibility to subsequent policy implementations / adaptions.

MaxiBoether (approved):

PR #603, opened 2024-08-23
fix: `PerformanceTrigger` batch pointer

robinholzi:
# Motivation

We previously didn't assert the exact indexes in the tests and used returned wrong batch pointers because of the internal unprocessed samples.


MaxiBoether (approved):
Thanks for the fix

PR #604, opened 2024-08-24
refactor: Add `BatchedTrigger`

robinholzi:
# Motivation

To eliminate the batching code duplication in the trigger implementations by moving it into a dedicated abstract base class.

### DIFF
--- a/modyn/supervisor/internal/triggers/amounttrigger.py
+++ b/modyn/supervisor/internal/triggers/amounttrigger.py
@@ -1,5 +1,7 @@
 from collections.abc import Generator
 
+from typing_extensions import override
+
 from modyn.config.schema.pipeline import DataAmountTriggerConfig
 from modyn.supervisor.internal.triggers.models import TriggerPolicyEvaluationLog
 from modyn.supervisor.internal.triggers.trigger import Trigger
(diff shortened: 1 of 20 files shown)


modyn/supervisor/internal/triggers/amounttrigger.py
```diff
@@ -16,8 +18,11 @@ def __init__(self, config: DataAmountTriggerConfig):
 
         assert self.data_points_for_trigger > 0, "data_points_for_trigger needs to be at least 1"
 
+    @override
```
MaxiBoether, line 21: so far we don't use this decorator in the codebase. any reason to introduce it now here?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -70,7 +73,7 @@ def __init__(self, config: DataDriftTriggerConfig):
         self.model: StatefulModel | None = None
 
         self._sample_left_until_detection = (
```
MaxiBoether, line 75: This should be set by the new parent class now?

MaxiBoether (approved):
Thank you, I only had minor comments, this makes reading the batched triggers much easier.

modyn/supervisor/internal/triggers/amounttrigger.py
robinholzi: For mypy to tell me when I forgot to update some signature of a function😅 The decorator is afaik only typing sugar, and doesn't affect runtime behaviour.

PR #605, opened 2024-08-26
feat: Add Avoidable misclassification based `CostTrigger`

robinholzi:
# Motivation

Potentially the final trigger to be implemented 🎈

A combination of `PerformanceTrigger` (based on misclassifications) and latency-based `CostTrigger`.

Mostly shuffling of code into a `PerformanceTriggerMixin` class that can be reused in the actual `PerformanceTrigger` and in the `AvoidableMisclassificationCostTrigger`


robinholzi:
@MaxiBoether Here we go :)

modyn/supervisor/internal/triggers/cost/costtrigger.py
```diff
@@ -29,13 +29,10 @@ def __init__(self, config: CostTriggerConfig):
         self.config = config
```
MaxiBoether, line 29: unrelated to this PR but currently we have both a cost_trigger and costtrigger.py living in this directory

MaxiBoether: also the decision policy.py file seems to not contain any code

modyn/supervisor/internal/triggers/cost/avoidablemissclassification_costtrigger.py
```diff
@@ -38,13 +49,39 @@ def inform_new_model(
...
+        estimated_new_avoidable_misclassifications, _ = (
+            self.misclassification_estimator.estimate_avoidable_misclassifications(
+                update_interval_samples=self.config.evaluation_interval_data_points,
+                data_density=self.data_density,
+                performance_tracker=self.performance_tracker,
+                method=self.config.forecasting_method,
+            )
+        )
+
+        return self._incorporation_latency_tracker.add_latency(
```
MaxiBoether, line 85: I think this might be related to the doc PR but I still don't have a good understanding how how the num of avoidable misclassifications functions as a regret. for the incorporation latencies it's clear to me that each sample's regret grows with each timestep it's not incorporated. how come that the num avoided samples's regret also grows over time? this - in my understanding - is a static value and does not get more severe over time (unlike the incorporation latency). Can you elaborate on this either in the docstring or in the docs PR potentially? thank you!

MaxiBoether (changes requested):
Thanks - I had a couple of comments on docs (again, talking about regrets :D),  but also some code stuff

modyn/supervisor/internal/triggers/cost/avoidablemissclassification_costtrigger.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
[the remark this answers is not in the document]
robinholzi: Oh snap, forgot about that will all the refactoring, thanks :)

modyn/supervisor/internal/triggers/cost/incorporation_latency_tracker.py
```diff
@@ -31,8 +31,11 @@ def add_latency(self, regret: float, batch_duration: float) -> float:
         Returns:
             Most recent cumulative regret value.
         """
+
+        # newly arrived `regret` has existed for `batch_duration / 2` seconds on average;
+        # old regret persists for the entire `batch_duration`
```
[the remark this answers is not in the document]
robinholzi: Yes, with the currently interface this is the best we can do. We don't know the timestamps of the misclassified samples. Even less we know about the time distribution of `avoidable` misclassifications which is a virtual (estimated) metric.

modyn/supervisor/internal/triggers/performancetrigger.py
```diff
@@ -58,40 +44,20 @@ class PerformanceTrigger(BatchedTrigger):
...
-            else None
-        )
 
     @override
     def init_trigger(self, context: TriggerContext) -> None:
-        self.context = context
-        self._init_dataloader_info()
-        self._init_model_downloader()
+        # Call PerformanceTriggerMixin's init_trigger method to initialize the internal performance detection state
+        self._init_trigger(context)
```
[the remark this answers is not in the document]
robinholzi: Yes, making this more explicit is definitely beneficial!

modyn/supervisor/internal/triggers/performance/misclassification_estimator.py
```diff
@@ -0,0 +1,77 @@
...
+from modyn.supervisor.internal.triggers.performance.data_density_tracker import (
+    DataDensityTracker,
+)
+from modyn.supervisor.internal.triggers.performance.performance_tracker import (
+    PerformanceTracker,
+)
+
+
+class NumberAvoidableMisclassificationEstimator:
+    """Stateless tracker of the number of avoidable misclassifications.
```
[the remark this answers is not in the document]
robinholzi: Not sure if I would call immutable configuration variables "stateful". I'll adjust that to `immutable`, making it clearer that no state is changed through member functions.

modyn/supervisor/internal/triggers/performance/misclassification_estimator.py
```diff
@@ -0,0 +1,77 @@
...
+
+        We support both the "hindsight" and "forecast" mode.
+
+        In the "hindsight" mode, the decision is made based on the current performance and the cumulated avoidable misclassifications.
+
+        Returns:
+            Tuple of the number of avoidable misclassifications and the forecasted number of avoidable misclassifications.
+        """
+
+        # [BACKWARD LOOKING]
```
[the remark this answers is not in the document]
robinholzi: I'll adjust that to `# compute new_avoidable_misclassifications ` and `# compute new_avoidable_misclassifications` (which are the backward and forward-looking parts that are computed here)

modyn/supervisor/internal/triggers/cost/avoidablemissclassification_costtrigger.py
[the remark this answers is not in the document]
robinholzi: We are always doing the forecast, and returning the backward looking estimate and the forward looking estimate. The caller can then decide which variant to choose for the policy. As our forecasting methods are very cheap computationally at this point, it's fair to do that. Especially as forward-looking callers need both results for their decision (if backward-looking result exceeds the threshold, then the forward-looking one shouldn't overrule this). If we should introduce more complex forecasting here, the refactoring shouldn't be too complicated.

modyn/supervisor/internal/triggers/cost/costtrigger.py
robinholzi: I wasn't fully sure whether to put the cost triggers (3x) into the /triggers directory or into a separate directory. For the other triggers, the trigger-specific subdirectories have only contained additional logic so far, and we have only had one `*trigger.py` for every policy type. When refining the docs PR earlier, I decided to put those files into `/triggers`. Thanks for the reminder here :)

modyn/supervisor/internal/triggers/cost/avoidablemissclassification_costtrigger.py
robinholzi: I'll add a comment in both places. Let's continue [here](https://github.com/eth-easl/modyn/pull/600#discussion_r1735130216)

modyn/tests/supervisor/internal/triggers/performance/test_performancetrigger_mixin.py
```diff
@@ -0,0 +1,146 @@
...
+    assert trigger.most_recent_model_id == 42
+    assert trigger.model_refresh_needed  # would be reset if _run_evaluation wasn't mocked
+
+    mock_evaluation.assert_called_once_with(interval_data=last_detection_interval)
+    mock_inform_trigger.assert_called_once_with(
+        num_samples=5, num_misclassifications=2, evaluation_scores={"Accuracy": 0.6}
+    )
+
+
+# Note: we don't test _run_evaluation as this would require more mocking than actual testing
```
MaxiBoether, line 120: I briefly searched and didn't find something, but if there is something like  a nocov pragma, then we could add that to run evaluation

MaxiBoether (approved):
This looks good to me! Basically no other comments!

modyn/tests/supervisor/internal/triggers/performance/test_performancetrigger_mixin.py
robinholzi: `# pragma: no cover` should do the trick!

PR #606, opened 2024-08-29
refactor: `DataLoaderInfo` dataclass

robinholzi:
# Motivation

Small quality of life improvement.  I haven't integrated the model into all the dataset implementations yet, even though most of the arguments are already in the `DataLoaderInfo`. I don't want to jump into this rabbit whole just now 😅

### DIFF
--- a/modyn/supervisor/internal/triggers/utils/datasets/dataloader_info.py
+++ b/modyn/supervisor/internal/triggers/utils/datasets/dataloader_info.py
@@ -1,30 +1,20 @@
 # TODO(415): Unify with similar classes in trainer_server and evaluator
+
+from dataclasses import dataclass
+
+
+@dataclass
 class DataLoaderInfo:
-    def __init__(
-        self,
-        pipeline_id: int,
-        dataset_id: str,
-        num_dataloaders: int,
-        batch_size: int,
-        bytes_parser: str,
-        transform_list: list[str],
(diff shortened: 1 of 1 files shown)


modyn/supervisor/internal/triggers/utils/datasets/dataloader_info.py
```diff
@@ -1,30 +1,20 @@
...
+    num_dataloaders: int
+    batch_size: int
+    bytes_parser: str
+    transform_list: list[str]
+    storage_address: str
+    selector_address: str
+    num_prefetched_partitions: int
+    parallel_prefetch_requests: int
+    shuffle: bool
+    tokenizer: str | None = None
```
MaxiBoether, line 19: there was no default for this before - are we sure nothing breaks by adding a default here?

robinholzi: What's the point of allowing `None` while not using it as the default? If any part of the code didn't specify the argument before, this was already an undefined behavior and should have been flagged by mypy. If `None` breaks something here, we should also not allow it to be passed.

MaxiBoether: fair

MaxiBoether (approved):

PR #607, opened 2024-09-01
feat: Add `Hellinger` distance

robinholzi:
# Motivation

Support `Hellinger distance` even though it's not directly supported by `evidently` or `alibi-detect`.

### DIFF
--- a/modyn/config/schema/pipeline/trigger/drift/evidently.py
+++ b/modyn/config/schema/pipeline/trigger/drift/evidently.py
@@ -5,39 +5,44 @@ from pydantic import Field
 from modyn.config.schema.pipeline.trigger.drift.metric import BaseMetric
 
 
-class _EvidentlyBaseDriftMetric(BaseMetric):
+class _EvidentlyPcaDriftMetric(BaseMetric):
     num_pca_component: int | None = Field(None)
 
 
-class EvidentlyModelDriftMetric(_EvidentlyBaseDriftMetric):
+class EvidentlyModelDriftMetric(_EvidentlyPcaDriftMetric):
     id: Literal["EvidentlyModelDriftMetric"] = Field("EvidentlyModelDriftMetric")
     bootstrap: bool = Field(False)
-    quantile_probability: float = 0.95
-    threshold: float = Field(0.55)
+    quantile_probability: float = Field(0.95, description="Threshold for the evidently api, unused in modyn. Usage discourged.")
+    threshold: float = Field(0.55, description="Threshold for the evidently api, unused in modyn. Usage discourged.")
 
 
-class EvidentlyRatioDriftMetric(_EvidentlyBaseDriftMetric):
+class EvidentlyRatioDriftMetric(_EvidentlyPcaDriftMetric):
     id: Literal["EvidentlyRatioDriftMetric"] = Field("EvidentlyRatioDriftMetric")
     component_stattest: str = Field(
         "wasserstein",
         description="The statistical test used to compare the components.",
     )
-    component_stattest_threshold: float = Field(0.1)
-    threshold: float = Field(0.2)
+    component_stattest_threshold: float = Field(0.1, description="Threshold for the evidently api, unused in modyn. Usage discourged.")
+    threshold: float = Field(0.2, description="Threshold for the evidently api, unused in modyn. Usage discourged.")
 
 
-class EvidentlySimpleDistanceDriftMetric(_EvidentlyBaseDriftMetric):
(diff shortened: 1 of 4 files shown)


modyn/config/schema/pipeline/trigger/drift/evidently.py
```diff
@@ -5,39 +5,52 @@
...
 
 
-class EvidentlyModelDriftMetric(_EvidentlyBaseDriftMetric):
+class EvidentlyModelDriftMetric(_EvidentlyPcaDriftMetric):
     id: Literal["EvidentlyModelDriftMetric"] = Field("EvidentlyModelDriftMetric")
     bootstrap: bool = Field(False)
-    quantile_probability: float = 0.95
-    threshold: float = Field(0.55)
+    quantile_probability: float = Field(
+        0.95, description="Threshold for the evidently api, unused in modyn. Usage discourged."
```
MaxiBoether, line 16: Copy&paste description from threshold?

modyn/config/schema/pipeline/trigger/drift/evidently.py
```diff
@@ -5,39 +5,52 @@
...
-    component_stattest_threshold: float = Field(0.1)
-    threshold: float = Field(0.2)
+    component_stattest_threshold: float = Field(
+        0.1, description="Threshold for the evidently api, unused in modyn. Usage discourged."
+    )
+    threshold: float = Field(0.2, description="Threshold for the evidently api, unused in modyn. Usage discourged.")
 
 
-class EvidentlySimpleDistanceDriftMetric(_EvidentlyBaseDriftMetric):
+class EvidentlySimpleDistanceDriftMetric(_EvidentlyPcaDriftMetric):
```
MaxiBoether, line 33: I don't get yet why everything is a Pca drift metric from an inheritance sense. I suppose if num_pca_components is none, no PCA is applied?

modyn/supervisor/internal/triggers/drift/detector/evidently.py
```diff
@@ -110,6 +118,7 @@ def _get_evidently_metrics(
             EmbeddingsDriftMetric(EVIDENTLY_COLUMN_MAPPING_NAME, _evidently_metric_factory(config)),
         )
         for metric_ref, config in metrics_config.items()
+        if config.id not in ["EvidentlyHellingerDistanceDriftMetric"]
```
MaxiBoether, line 121: can we make this a global in this file like `SPECIAL_METRICS` or something?

modyn/supervisor/internal/triggers/drift/detector/evidently.py
```diff
@@ -141,3 +150,25 @@ def _evidently_metric_factory(config: EvidentlyDriftMetric) -> EmbeddingsDriftMe
...
+    embeddings_ref: pd.DataFrame,
+    embeddings_cur: pd.DataFrame,
+) -> dict[str, MetricResult]:
+    metric_results: dict[str, MetricResult] = {}
+    for metric_ref, config in configs.items():
+        if config.id == "EvidentlyHellingerDistanceDriftMetric":
+            column_distances = [
+                # [0]: Hellinger distance, [1]: decision with dummy threshold (False)
+                _hellinger_distance(embeddings_ref[c], embeddings_cur[c], ColumnType.Numerical, 0)[0]
+                for c in embeddings_ref.columns
```
MaxiBoether, line 166: Each column is a data point, and the rows in this column vector are the embedding dimensions? (in my head it was the other way around just want to make sure we supply it the right way)

MaxiBoether (changes requested):
Thank you! Just had small comments / naming questions, after this we can merge

modyn/config/schema/pipeline/trigger/drift/evidently.py
robinholzi: no the quantile probability is also a threshold, but in the probability space

modyn/config/schema/pipeline/trigger/drift/evidently.py
robinholzi: Renamed to mixin, it's just for avoiding have to duplicate the field everywhere. And yes, none is no pca. We don't use that in our experiments though.

modyn/supervisor/internal/triggers/drift/detector/evidently.py
robinholzi: sure

MaxiBoether (approved):

PR #609, opened 2024-09-01
Fix dimensionality issue when binary classification outputs 1D instead of 2D tensor

XianzheMa:
For binary classification (e.g. DLRM), there are two problems:

1. The model output no longer has shape (batch_size, num_classes). The size is not even (batch_size, 1) as the second dimension, 1, is squeezed.
2. The model output is not passed through a softmax layer but a sigmoid layer.

This PR adapts the existing downsamplers to the binary classfication.


modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
```diff
@@ -139,7 +141,7 @@ def _select_indexes_from_scores(self, target_size: int) -> tuple[list[int], torc
         # we select those with minimal negative entropy, i.e., maximum entropy
         # Margin: We look for the smallest margin. The larger the margin, the more certain the
         # model is.
-        return np.argsort(self.scores)[:target_size], torch.ones(target_size).float()
```
XianzheMa, line 142: The previous way did not return a type compatible to method type hint.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
```diff
@@ -49,10 +49,16 @@ def inform_samples(
...
+            # BCEWithLogitsLoss requires that forward_output and target have the same shape
+            forward_output = forward_output.unsqueeze(1)
+            target = target.unsqueeze(1)
+
         last_layer_gradients = self._compute_last_layer_gradient_wrt_loss_sum(
             self.per_sample_loss_fct, forward_output, target
         )
-        scores = torch.norm(last_layer_gradients, dim=-1).cpu()
+        # pylint: disable=not-callable
+        scores = torch.linalg.vector_norm(last_layer_gradients, dim=1).cpu()
```
XianzheMa, line 61: torch.norm is already deprecated https://pytorch.org/docs/stable/generated/torch.norm.html

MaxiBoether: However, you changed the dim parameter here from -1 to 1. Is that the expected behavior?

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
```diff
@@ -71,6 +71,9 @@ def inform_samples(
     ) -> None:
         batch_size = len(sample_ids)
         assert self.matrix_content is not None
+        if forward_output.dim() == 1:
+            forward_output = forward_output.unsqueeze(1)
+            target = target.unsqueeze(1)
```
MaxiBoether, line 76: This code is basically everywhere. Can we have a utils function `_unsequeeze_if_required` or something, and in that utils function have a brief comment that this is necessary for the cases where we have binary classification with sigmoid instead of softmax?

MaxiBoether: Also, is this sufficient? I did not think this through, but all of these downsamplers were written for the softmax-case, i.e., 1 entry in a probabilty vector per class. I wonder whether this also works here, or if we need to "simulate" a softmax like you did in uncertainty implementation: 
```
# sigmoid
 preds = torch.cat((1 - preds, preds), dim=1)
```

this also applies to the other downmsaplers (craig, ...)

MaxiBoether (changes requested):
Thank you! This mostly looks good, I had a couple of clarification comments and suggestions.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
XianzheMa: Yes I validated this before.

`last_layer_gradients` is of two dimensions batch_size * gradient_size. We should calculate the norm only on the dimension 1.

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
```diff
@@ -69,28 +69,42 @@ def inform_samples(
...
-
-            # remove highest class from softmax output
-            preds[torch.ones(preds.shape[0], dtype=bool), preds_argmax] = -1.0
-
-            preds_sub_argmax = torch.argmax(preds, dim=1)  # gets new top class (=> 2nd top class)
-            second_max_preds = preds[torch.ones(preds.shape[0], dtype=bool), preds_sub_argmax]
-            scores = (max_preds - second_max_preds).cpu().numpy()
+            if feature_size == 1:
+                # for binary classification comparing how far away the element is from 0.5 after sigmoid layer
+                # is the same as comparing the absolute value of the element before sigmoid layer
```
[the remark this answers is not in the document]
XianzheMa: So for least confidence we score each sample by the class that has the highest output value. So for each sample we first need to find which class has the highest output value.

For binary classification we have two classes, so we need to *compare* their output value.

Comparing their output value *after sigmoid layer* is equivalent to comparing the output value of any one class of the two, *after sigmoid layer*, to `0.5` as their sum is one.

Furthermore, comparing the output value of any one class of the two *after sigmoid layer* to `0.5` is equivalent to comparing the output value of any one class of the two *before sigmoid layer* to `0`. The larger (and hence the positive) one has the value as the absolute value of any output value.

We assume the model output is not sigmoid transformed, because we use the `BCEWithLogitsLoss` loss as the loss function. This loss function contains a sigmoid layer. Also this is similar to the assumption that for multi-classification we assume the model output is not softmax transformed, as `CrossEntropyLoss` has the softmax layer.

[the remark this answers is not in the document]
XianzheMa: Does this answer your question? If so, I will make the comments more detailed for easy comprehension.

modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_matrix_downsampling_strategy.py
XianzheMa: I checked all downsampler implementations only uncertainty based downsamplers assume that the output is in the form of "1 entry in a probabilty vector per class". For others they are based on gradients or embedding which does not depend on the output form.

For the util function: sure I will make the change

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_gradnorm_downsampling.py
XianzheMa: dim=-1 refers to the last dimension. So they refer to the same thing (assuming it has 2 dimensions which is true)

MaxiBoether: Makes sense!

modyn/trainer_server/internal/trainer/remote_downsamplers/remote_uncertainty_downsampling_strategy.py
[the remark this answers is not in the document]
MaxiBoether: Sorry, I think I am still either missing something or am confused. `sigmoid` maps any arbitrary real to (0,1), i.e., the input is 1D.

"Comparing their output value after sigmoid layer is equivalent to comparing the output value of any one class of the two, after sigmoid layer, to 0.5 as their sum is one."

I do not understand this. I don't understand what are you comparing. The input to sigmoid is 1D, and the output is 1D, but you're transforming it to a probability. What I do agree with if the input grows/shrinks, then the sigmoid probability also grows/shrinks. So for f(x) = sigmoid(x), if x is very small, the model will be very certain (that it is class 0), if x is very large, the model will be very certain (that it is class 1), and only for x inbetween it will be uncertain (probability around 0.5). 

In the 2D case, we take the lowest probability as the score in the vector. What I would've expected here is that you transform the output like you also do for margin/entropy using actually sigmoid and torch.cat. It might be that we can avoid this and this is cheaper, but I do still not understand where torch.abs is coming from. Because both large and small x can imply uncertainty/certainty, if I am not mistaken. A very low probability also means the model is certain? But I might totally be missing something, sorry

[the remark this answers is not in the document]
MaxiBoether: Ok I only now realized that sigmoid(0) = 0.5, and hence your abs approach indeed works! Maybe you could add a comment that a model output of 0 represents total uncertainty here, and the higher the absolute value the higher/lower the probability, hence this works.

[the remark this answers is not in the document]
XianzheMa: I added detailed explanation in the comment. Let me know if we can merge this PR. Thanks!

MaxiBoether (approved):
Thank you, Xianzhe! This LGTM!

PR #613, opened 2024-09-02
feat: Add `WarmupTrigger` abstraction

robinholzi:
# Motivation

Add the warmup behavior from `DriftTrigger` also to `CostTrigger` and `PerformanceTrigger` via a `WarmupTrigger`


modyn/config/schema/pipeline/trigger/common/batched.py
```diff
@@ -10,3 +11,18 @@ class BatchedTriggerConfig(ModynBaseModel):
             "estimation, accuracy evaluation, drift detection, ...) should be performed."
         )
     )
+    warmup_intervals: int | None = Field(
+        None,
+        description=(
+            "The number of intervals before starting to use the main policy. Some "
+            "`DecisionCriteria` use this to calibrate the threshold. During the warmup, a simpler `warmup_policy` "
```
MaxiBoether, line 18: DecisionCriteria come from drifts, right? Should we change this since this docstirng is more general now?

modyn/config/schema/pipeline/trigger/cost/cost.py
```diff
@@ -113,6 +113,14 @@ def conversion_factor(self) -> float:
         """
         return self.avoidable_misclassification_latency_per_training_second
 
+    @model_validator(mode="after")
+    def warmup_policy_requirement(self) -> Self:
+        """Assert whether the warmup policy is set when a metric needs
+        calibration."""
+        if self.warmup_policy is None:
+            raise ValueError("A warmup policy is required for metrics that need calibration.")
```
MaxiBoether, line 121: change message? we don't have a metric here thats a drift term

modyn/config/schema/pipeline/trigger/performance/performance.py
```diff
@@ -84,3 +84,12 @@ def validate_decision_criteria(self) -> "PerformanceTriggerConfig":
...
                 )
         return self
+
+    @model_validator(mode="after")
+    def warmup_policy_requirement(self) -> Self:
+        """Assert whether the warmup policy is set when a metric needs
+        calibration."""
+        for criterion in self.decision_criteria.values():
+            if criterion.needs_calibration and self.warmup_policy is None:
+                raise ValueError("A warmup policy is required for metrics that need calibration.")
```
MaxiBoether, line 94: criteria?

modyn/supervisor/internal/triggers/datadrifttrigger.py
```diff
@@ -232,7 +214,7 @@ def _evaluate_batch(
         if log:
             log.evaluations.append(drift_eval_log)
 
-        if not self.warmup_completed:
+        if not self.warmup_trigger.completed:
             # during the warmup phase we always want to reset the windows as if we detected drift;
```
MaxiBoether, line 218: i just realized it (and know it wasnt touched by this pr) but maybe we should say why: otherwise the current and reference model would be very similar, right?

actually, this makes me wonder: since I use allow_overlap = True, I don't reset the model. do we expect the policies to work then at all?

MaxiBoether (changes requested):
Thanks. I could not find issues in the actual logic, only in some docstrings and an unrelated comment about resetting the windows during warmup

modyn/config/schema/pipeline/trigger/common/batched.py
robinholzi: Performance Triggers also have them. Cost triggers don't so I'll adjust that

MaxiBoether (approved):

PR #614, opened 2024-09-03
fix: Add last interval to `BetweenTwoTriggersEvalStrategy`

robinholzi:
# Motivation

As for the time after the last training there's no more training hence the time is not covered in an eval interval, we need to handle this case manually


MaxiBoether (approved):
LGTM

PR #615, opened 2024-09-03
feat: Add new arch diagram

robinholzi:


MaxiBoether (approved):
i asked claude to generate ascii art for approving something and got back an image containing ascii art.
 
image

PR #619, opened 2024-09-03
fix: Add max samples for drift

robinholzi:
# Motivation

Currently, we don't sample from drift detection windows. This parameter got lost somewhere in the midst of data drift (config) refactoring.

AmountWindowStrategies don't rely on this, but for time-based windows, we could end up with way too many samples, which makes detection very slow because of quadratic distance computation runtime.


MaxiBoether (approved):
Thanks!

PR #620, opened 2024-09-04
fix: Fix several issues 

robinholzi:
# Motivation

Mostly computation of ROC on 1 sample

### DIFF
--- a/modyn/evaluator/internal/dataset/evaluation_dataset.py
+++ b/modyn/evaluator/internal/dataset/evaluation_dataset.py
@@ -81,7 +81,7 @@ class EvaluationDataset(IterableDataset):
             self._transform = transforms.Compose(self._transform_list)
 
     @retry(
-        stop=stop_after_attempt(25),
+        stop=stop_after_attempt(10),
         wait=wait_random_exponential(multiplier=1, min=2, max=60),
         before=before_log(logger, logging.ERROR),
         after=after_log(logger, logging.ERROR),
--- a/modyn/evaluator/internal/dataset/evaluation_dataset.py
+++ b/modyn/evaluator/internal/dataset/evaluation_dataset.py
@@ -81,7 +81,7 @@ class EvaluationDataset(IterableDataset):
             self._transform = transforms.Compose(self._transform_list)
 
     @retry(
-        stop=stop_after_attempt(5),
+        stop=stop_after_attempt(25),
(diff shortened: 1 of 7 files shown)


modyn/supervisor/internal/pipeline_executor/evaluation_executor.py
```diff
@@ -419,48 +420,62 @@ def get_failure_reason(eval_aborted_reason: EvaluationAbortedReason) -> str:
 # ------------------------------------------------------------------------------------ #
 
 
-def eval_executor_single_pipeline(pipeline_dir: Path) -> None:
+def eval_executor_single_pipeline(pipeline_dir: Path) -> SupervisorLogs:
```
robinholzi, line 423: Only for more efficient manual evaluation runs

modyn/evaluator/internal/dataset/evaluation_dataset.py
```diff
@@ -81,7 +81,7 @@ def _setup_composed_transform(self) -> None:
             self._transform = transforms.Compose(self._transform_list)
 
     @retry(
-        stop=stop_after_attempt(5),
+        stop=stop_after_attempt(25),
```
MaxiBoether, line 84: i would not really increase this to 25

modyn/evaluator/internal/metrics/roc_auc.py
```diff
@@ -26,6 +26,11 @@ def _dataset_evaluated_callback(self, y_true: torch.Tensor, y_pred: torch.Tensor
         """
         assert self.evaluation_result is None
 
+        if y_pred.dim() < 1 or y_true.dim() < 1 or len(y_pred) < 2 or len(y_true) < 2:
```
MaxiBoether, line 29: y_pred / true shape should match (see next assert) so its enough to check one of them

modyn/evaluator/internal/dataset/evaluation_dataset.py
robinholzi: 10 maybe? for evaluations, it's not harmful if they take longer because of more retries. Within the business logic it could make experiments produce weird results silently.

MaxiBoether: I mean okay we can merge 10 but 25 seems to be a bit high to have in the main branch :D

MaxiBoether (approved):
we can merge this and figure out whats wrong with clang format later

PR #623, opened 2024-09-06
Speed up `get_tensors_subset` from O(n^2) to O(n)

XianzheMa:
This is a change to replace the changes in my thesis branch to return 0-based indices in the `select_points()` method in downsamplers.

I realize that returning 0-based indices is not generalizable to all downsamplers because there are downsamplers which work class by class even in the batch-then-sample mode, e.g. craig. Therefore there is no built-in concept of 0-based indices for such downsamplers (because they pick points of the same class and select from them).

Therefore I decided to tweak the implementation of `get_tensors_subset()` which is simpler.

A time complexity of O(n) is acceptable because in any selection strategy we have to look at every sample, so an extra O(n) part can be well mixed in the time complexity of a selection strategy.

### DIFF
--- a/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_get_tensor_subset.py
+++ b/modyn/tests/trainer_server/internal/trainer/remote_downsamplers/test_get_tensor_subset.py
@@ -61,7 +61,7 @@ def test_get_tensors_subset_invalid_index():
     # Test when selected_indexes contains an invalid index
     selected_indexes.append(999)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(KeyError):
         get_tensors_subset(selected_indexes, data, target, sample_ids)
 
 
--- a/modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
+++ b/modyn/trainer_server/internal/trainer/remote_downsamplers/abstract_remote_downsampling_strategy.py
@@ -20,9 +20,9 @@ def get_tensors_subset(
     and then we get the entries of data and target only for the selected
     samples
     """
-
+    sample_id2index = {sample_id: index for index, sample_id in enumerate(sample_ids)}
     # first of all we compute the position of each selected index within the batch
-    in_batch_index = [sample_ids.index(selected_index) for selected_index in selected_indexes]
+    in_batch_index = [sample_id2index[selected_index] for selected_index in selected_indexes]
 
     # then we extract the data
     if isinstance(data, torch.Tensor):


MaxiBoether:
Thank you, Xianzhe. Have you compared this implementation with your other 0-indexed based / tensor based one in some microbenchmark? It would be great to understand the implications here before merging this

XianzheMa:
> Thank you, Xianzhe. Have you compared this implementation with your other 0-indexed based / tensor based one in some microbenchmark? It would be great to understand the implications here before merging this

I can't execute a full pipeline now because I already cleared the datasets on the servers, but I measured them with the following scripts


```python
import torch



def get_tensors_subset_new(
        selected_indexes: list[int], data: torch.Tensor | dict, target: torch.Tensor, sample_ids: list
) -> tuple[torch.Tensor | dict, torch.Tensor]:
    sample_id2index = {sample_id: index for index, sample_id in enumerate(sample_ids)}
    # first of all we compute the position of each selected index within the batch
    in_batch_index = [sample_id2index[selected_index] for selected_index in selected_indexes]

    # then we extract the data
    if isinstance(data, torch.Tensor):
        sub_data = data[in_batch_index]
    else:
        sub_data = {key: tensor[in_batch_index] for key, tensor in data.items()}

    # and the targets
    sub_target = target[in_batch_index]

    return sub_data, sub_target


def get_tensors_subset_old(
        selected_indexes: list[int], data: torch.Tensor | dict, target: torch.Tensor, sample_ids: list
) -> tuple[torch.Tensor | dict, torch.Tensor]:
    sample_id2index = {sample_id: index for index, sample_id in enumerate(sample_ids)}
    # first of all we compute the position of each selected index within the batch
    in_batch_index = [sample_ids.index(selected_index) for selected_index in selected_indexes]

    # then we extract the data
    if isinstance(data, torch.Tensor):
        sub_data = data[in_batch_index]
    else:
        sub_data = {key: tensor[in_batch_index] for key, tensor in data.items()}

    # and the targets
    sub_target = target[in_batch_index]

    return sub_data, sub_target


def get_tensors_subset_0_based(
        in_batch_index: list[int], data: torch.Tensor | dict, target: torch.Tensor, sample_ids: list
) -> tuple[torch.Tensor | dict, torch.Tensor]:

    # then we extract the data
    if isinstance(data, torch.Tensor):
        sub_data = data[in_batch_index]
    else:
        sub_data = {key: tensor[in_batch_index] for key, tensor in data.items()}

    # and the targets
    sub_target = target[in_batch_index]

    return sub_data, sub_target



if __name__ == "__main__":
    import time
    cuda_device = torch.device("cuda:2")
    data = torch.randn(65536, 10, device=cuda_device)
    target = torch.randn(65536, device=cuda_device)
    # 65536 random integers
    sample_ids = torch.randint(0, 1000000, (65536,)).tolist()

    # suppose we have a selection ratio of 50%. We select 32768 samples
    # generate 32768 random indexes in range 0-65536
    in_batch_idx = torch.randint(0, 65536, (32768,)).tolist()

    selected_sample_ids = [sample_ids[idx] for idx in in_batch_idx]
    print("start experiment")
    times = 10
    warmup = 3
    for i in range(times + warmup):
        if i == warmup:
            start = time.time()
        get_tensors_subset_old(selected_sample_ids, data, target, sample_ids)
    print("Old:", time.time() - start)

    for i in range(times + warmup):
        if i == warmup:
            start = time.time()
        get_tensors_subset_new(selected_sample_ids, data, target, sample_ids)
    print("New:", time.time() - start)

    for i in range(times + warmup):
        if i == warmup:
            start = time.time()
        get_tensors_subset_0_based(in_batch_idx, data, target, sample_ids)
    print("0-based:", time.time() - start)
```

The result is
```
start experiment
Old: 66.20023608207703
New: 0.12652850151062012
0-based: 0.058889150619506836
```

Though the way proposed in this PR ("New") is still longer than 0-based indices, compared to how long a selection strategy would take (at least O(n) ), the difference between the two is ignorable.

MaxiBoether (approved):
Thank you, this microbenchmark is exactly what I was looking for and this change makes a lot of sense!

MaxiBoether:
merged this for you because i want to use it in some experiment :)

PR #626, opened 2024-09-09
refactor: Adjust performance triggering policies

robinholzi:


modyn/supervisor/internal/triggers/performance/performancetrigger_mixin.py
```diff
@@ -64,22 +64,29 @@ def _inform_new_model(self, most_recent_model_id: int, last_detection_interval:
...
-        num_samples, num_misclassifications, evaluation_scores = self._run_evaluation(
-            interval_data=last_detection_interval
+        # num_samples, num_misclassifications, evaluation_scores = self._run_evaluation(
+        #     interval_data=last_detection_interval
+        # )
+        assert self.config.mode == "hindsight", (
+            "Forecasting mode is not supported yet, it requires tracking the performance right after trigger. "
+            "However, after triggers the models has learned from the last detection interval. We would need to "
+            "maintain a holdout set for this."
         )
```
MaxiBoether, line 78: Is that (with the commented out code) what we should merge? Sorry, I am not fully sure yet about the changes here. Right now, this inform function does not do anything besides updating internal state

modyn/supervisor/internal/triggers/performance/decision_policy.py
```diff
@@ -98,22 +102,35 @@ def evaluate_decision(
         mode: TriggerEvaluationMode,
         method: ForecastingMethod,
     ) -> bool:
-        # to compute the rolling average we simply reuse the forecast method, it's not a true forecast here
-        expected = performance_tracker.forecast_optimal_performance(self.config.metric, method="rolling_average")
-        deviation = expected - evaluation_scores[self.config.metric]
-        allowed_absolute_deviation = (
-            self.config.deviation if self.config.absolute else (self.config.deviation * expected)
-        )
+        return self._wrapped.evaluate_decision(measurement=evaluation_scores[self.metric])
```
MaxiBoether, line 105: Basically all of those functions are currently just wrappers after the update. In the drift case, we use the decision policies directly. Any reason for keeping those wrappers?

modyn/supervisor/internal/triggers/utils/decision_policy.py
```diff
@@ -0,0 +1,152 @@
...
+        rolling_average = sum(self.score_observations) / len(self.score_observations)
+        deviation = (
+            measurement - rolling_average if self.triggering_direction == "higher" else rolling_average - measurement
+        )
+
+        self.score_observations.append(measurement)
+
+        if self.absolute:
+            return deviation >= self.deviation
+        return deviation >= self.deviation * rolling_average
```
MaxiBoether, line 152: I guess we should have a test for the relative case as well (codecov is saying we dont have one)

MaxiBoether (changes requested):
Thank you, I had a couple of smaller comments

modyn/supervisor/internal/triggers/performance/performancetrigger_mixin.py
robinholzi: Yep, I didn't want to remove it as it might still be useful in the future. The pre-mature implementation is meanly stripped, but I'd like to keep this interface abstraction.

modyn/supervisor/internal/triggers/performance/decision_policy.py
robinholzi: the `avoidable misclassification` trigger needs more information; hence I kept the interface. Also I don't have a lot of time for refactoring rn.

modyn/supervisor/internal/triggers/utils/decision_policy.py
robinholzi: Done

MaxiBoether (approved):

PR #627, opened 2024-09-10
hotfix: Fix type hint

MaxiBoether:
Not sure why my local pre commit hooks/even just running mypy doesn't catch this, probably some inconsistency between CI config and local config


MaxiBoether:
thanks xianzhe :)

PR #632, opened 2024-09-11
Use `torch.no_grad()` instead of inference mode in `pytorch_trainer.py`

XianzheMa:


.github/workflows/workflow.yaml
```diff
@@ -220,7 +220,7 @@ jobs:
 
       - name: Upload HTML coverage report
         if: ${{ matrix.compiler.coverage && matrix.build-type == 'Debug' }}
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
```
XianzheMa, line 223: to fix the failing CI

MaxiBoether (approved):
I made some slight adjustments. Please check them out before merging, and only merge if they make sense to you.

PR #633, opened 2024-09-11
fix!: Fix wrong directional check for low is worse metrics

robinholzi:
# Changes

- renamed `Percentile` > `Quantile` for more concise naming
- use np.quantile for dynamic quantile thresholds
- adapt and add more tests


MaxiBoether (approved):
Approving since the test LGTM. Need to think about the details of sorting but we can mergr this already

MaxiBoether (approved):
Thanks (also for switching from percentile to quantile)

PR #640, opened 2024-12-11
Updated storage and training files to enable text generation tasks. A…

sjohn4:
Made minor updates to the environment and ensured everything passes the standard Python compliance tests, maintaining compatibility with Modyn's existing functionality. Added GPT-2 as a model that can be used along with the GPT-2 tokenizer. Introduced the generative flag to the Modyn pipeline, enabling it to distinguish between generative and classification tasks. Also updated some Python test files to incorporate this flag, ensuring no errors are thrown.


modyn/config/examples/modyn_config.yaml
```diff
@@ -278,7 +278,7 @@ selector:
   local_storage_directory: "/tmp/local_storage"
   local_storage_max_samples_in_file: 1000000
   cleanup_storage_directories_after_shutdown: true
-  ignore_existing_trigger_samples: false
+  ignore_existing_trigger_samples: true
```
MaxiBoether, line 281: let's not merge this

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -409,8 +626,45 @@ def _fetch_partition_noprefetch(
                 container["weights"][idx],
             )
 
+    def _fetch_partition_noprefetch_generative(  # based on the one above
```
MaxiBoether, line 629: similar comments here. I would propose to refactor the PR such that we don't have parallel codepaths for generative

MaxiBoether (changes requested):
I briefly skimmed the PR. Thanks for the work! The biggest feedback right now would be that we should not have parallel codepaths which are mostly copy/paste of existing ones but without labels. Rather, let's have clean code where we move labels to be optional, both in the c++/protobuf/python worlds. And before I do the proper review, would be great if you could check the diff for changes that should not go onto main :)

Thanks!

sjohn4:
Redid most of this changes, this pr is not relevant anymore, I will open a new one shortly.

PR #641, opened 2024-12-28
feat: Add reusable plotting features

robinholzi:
# Motivation

Some abstractions that allow the reuse of common plotting functionality


MaxiBoether (approved):
Thanks - LGTM! I guess there is some duplication between this and our notebooks now, right?

robinholzi:
> Thanks - LGTM! I guess there is some duplication between this and our notebooks now, right?

Sure, I used the notebooks I created for sigmod as a basis for my thesis plots :) I just adjusted styling though so that plots look consistent in my latex document (e.g. pdf plots with embedded font instead of svgs as the latter can hardly be scaled in latex because of the fixed uniform font size)

PR #642, opened 2024-12-28
chore: Add robin thesis plotting notebooks

robinholzi:
# Motivation

Optional: Not needed in OSS repo, might help someone in the future recreate a plot in a similar style.


MaxiBoether:
Can you change the target branch of this PR to merge this into like `feat/RobinHolzi/thesis-plots`? Such that we have it here flying around, but not in the main branch. In this context, maybe you could also clean up any branches on the main repo you don't need anymore

PR #643, opened 2025-01-06
Text generation

sjohn4:
Redid my changes so we do not have two branching paths.


MaxiBoether (changes requested):
Thanks John! Some first comments, the most important being that we should still make labels just an optional entry instead of having 2 responses. For the next review, it would be cool if you could handle CI first, having a clean diff makes it a lot easier to review :)

MaxiBoether:
I guess we can close this PR?

PR #647, opened 2025-01-13
Text generation

sjohn4:
Took the previous feedback in mind and corrected the code.


MaxiBoether (changes requested):
Hey,

Thanks again for the changes. I did a first pass (without a detailed look at the intricacies of the OnlineDataset) and realized the GetNL call is still there. Let's make this a bit more readable by removing this as mentioned in the detailed comments, and only having the relevant diff here. If you could add some tests for the stuff you change/add, that would also be great. Thank you!

PR #649, opened 2025-02-16
Here is everything needed for training without labels nothing less an…

sjohn4:
Here is the most basic implementation for training without labels, with no extra changes. There is no mystery to this—it adds the generative flag to the pipeline and lets you train on datasets that do not have labels. These changes are pretty trivial; the online dataset does not return the labels field and instead returns None. One slightly unusual aspect is that I found the best way to make this comply with the internal logic of the storage server is to simply fill the "label" field in storage with -1 as a placeholder. Perhaps we should add a check somewhere so that you cannot use a dataset with has_labels set to false if you do not have the generative flag in your pipeline.


modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -270,13 +280,20 @@ def _get_data(
...
         assert self._uses_weights is not None
         self._sw.start("transform", resume=True)
         # mypy complains here because _transform has unknown type, which is ok
         transformed_sample = self._transform(sample)  # type: ignore
         self._sw.stop("transform")
+
+        if not self._include_labels:
+            if self._uses_weights:
+                return key, transformed_sample, weight
+            return key, transformed_sample
```
MaxiBoether, line 294: shouldnt for consistency be the label returned as None? i am not sure, but _get_data_from_storage you use None, while here the semantics change. Could you elaborate a bit here? Can we somehow compact those different cases?

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -329,13 +346,19 @@ def _prefetch_partition(self, worker_id: int, maybe_continue: bool = False) -> N
...
+                    "keys": [],
+                    "labels": [],
+                    "weights": [],
+                }
+            else:
+                self._thread_data_container[self._next_partition_to_fetch] = {
+                    "data": [],
+                    "keys": [],
+                    "weights": [],
+                }
```
MaxiBoether, line 361: I don't thik this is necessary, we can just have an empty labels container.

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -391,23 +414,51 @@ def callback_func() -> None:
...
-    ) -> Iterator[tuple[int, memoryview, int, float | None]]:
+    ) -> Iterator[tuple[int, memoryview, int, float | None]] | Iterator[tuple[int, memoryview, float | None]]:
         assert self._num_prefetched_partitions < 1
-        container: dict[str, Any] = {"data": [], "keys": [], "labels": [], "weights": []}
+        container: dict[str, Any] = {
+            "data": [],
+            "keys": [],
+            "weights": [],
+        }
+        if self._include_labels:
```
MaxiBoether, line 424: that way we should be able to avoid all of those changes.

MaxiBoether (changes requested):
did a first pass of this. i think here it is especially important to add some more unit and integrationtests as outlined. thanks for the work!

modyn/trainer_server/internal/dataset/online_dataset.py
sjohn4: Then out it goes .)

modyn/tests/storage/storage_test_utils.cpp
```diff
@@ -25,6 +26,39 @@ label_file_extension: ".lbl"
...
+)";
+  return test_config;
+}
+YAML::Node StorageTestUtils::get_dummy_file_wrapper_config_unlabeled(const std::string& byteorder) {
+  YAML::Node config;
+  config["byteorder"] = byteorder;
+  config["file_extension"] = ".txt";
+  config["label_file_extension"] = ".json";
+  config["label_size"] = 2;
+  config["record_size"] = 4;
```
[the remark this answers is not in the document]
sjohn4: This is marked very weirdly by  git diff its just a copy of the dummy configs with has_labels marked as false,

modyn/tests/storage/internal/file_wrapper/csv_file_wrapper_test.cpp
```diff
@@ -41,7 +41,33 @@ class CsvFileWrapperTest : public ::testing::Test {
 
   void TearDown() override { std::filesystem::remove_all(file_name_); }
 };
+class CsvFileWrapperTestUnlabeled : public ::testing::Test {
```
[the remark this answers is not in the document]
sjohn4: Yes I had this, I thik it threw some weir error I did not solve but I will put it back

PR #650, opened 2025-02-16
feat: Gradient clipping and gradient accumulation

sjohn4:
Here is the code for gradient clipping and accumulation.


modyn/config/schema/pipeline/training/config.py
```diff
@@ -154,7 +155,14 @@ class TrainingConfig(ModynBaseModel):
         None,
         description="Configuration for the torch.cuda.amp.GradScaler. Effective only when amp is enabled.",
     )
-
+    grad_norm: float = Field(
```
MaxiBoether, line 158: Should this really be great norm, not sth like clip? How is that called in huggingface?

modyn/config/schema/pipeline/training/config.py
```diff
@@ -154,7 +155,14 @@ class TrainingConfig(ModynBaseModel):
...
+        default=0,
+        description="Clips the gradients normed over this value, if its 0 it will not be used.",
+    )
+    gradient_accumulation_steps: int = Field(
+        default=1,
```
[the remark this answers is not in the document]
MaxiBoether: and default should probably be None? otherwise it does not make sense that the proto field is optional (holds for both)

modyn/protos/trainer_server.proto
```diff
@@ -59,6 +59,8 @@ message StartTrainingRequest {
   bool enable_accurate_gpu_measurements = 25;
   int64 record_loss_every = 26;
   bool drop_last_batch = 27;
+  optional float grad_norm = 29;
+  optional int64 gradient_accumulation_steps = 30;
```
MaxiBoether, line 63: see my other comment, either allow none or make this non optional

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -85,9 +85,9 @@ def __init__(
         self.pipeline_id = training_info.pipeline_id
         self.training_id = training_info.training_id
         self.trigger_id = training_info.trigger_id
-
+        self._grad_norm = training_info.grad_norm
         self.selector_stub = self.connect_to_selector(training_info.selector_address)
-
+        self.gradient_accumulation_steps = training_info.gradient_accumulation_steps
```
MaxiBoether, line 90: inconsistent naming: one attribute you made private (_ prefix), the other not.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -316,27 +318,26 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
...
                 with GPUMeasurement(self._measure_gpu_ops, "Backward", self._device, stopw, resume=True):
                     self._scaler.scale(loss).backward()
-
-                with GPUMeasurement(self._measure_gpu_ops, "OptimizerStep", self._device, stopw, resume=True):
-                    for _, optimizer in self._optimizers.items():
-                        self._scaler.step(optimizer)
-
-                    self._scaler.update()
+                    if self._grad_norm is not None:
+                        torch.nn.utils.clip_grad_norm_(self._model.model.parameters(), max_norm=self._grad_norm)
```
MaxiBoether, line 322: so probably we should call this parameter max norm.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -306,7 +308,7 @@ def train(self) -> None:  # pylint: disable=too-many-locals, too-many-branches
                             loss = torch.dot(self._criterion_nored(output, target), weights / weights.sum())
                         else:
                             loss = self._criterion(output, target)
-
+                        loss = loss / self.gradient_accumulation_steps
```
MaxiBoether, line 311: currently gradient accumulation steps can be None, so we should probably only do this if it is not none

modyn/trainer_server/internal/utils/training_info.py
```diff
@@ -80,5 +79,6 @@ def __init__(
 
         self.seed: int | None = request.seed if request.HasField("seed") else None
         self.tokenizer: str | None = request.tokenizer.value if request.HasField("tokenizer") else None
-
+        self.grad_norm: float | None = request.grad_norm if request.HasField("grad_norm") else None
+        self.gradient_accumulation_steps: int = request.gradient_accumulation_steps
```
MaxiBoether, line 83: inconsistent with None

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -258,6 +258,8 @@ def get_training_info(
                 label_transformer=PythonString(value=get_mock_label_transformer() if transform_label else ""),
                 grad_scaler_configuration=JsonString(value=json.dumps({})),
                 epochs_per_trigger=1,
+                grad_norm=0.1,
+                gradient_accumulation_steps=1,
```
MaxiBoether, line 262: Could you add a test that actually semantically checks the gradient accumulation logic?

modyn/config/schema/pipeline/training/config.py
sjohn4: As far as I know gradient clipping is generaly done with torch.nn.utils.clip_grad_norm_ so I think this does make sense.

modyn/config/schema/pipeline/training/config.py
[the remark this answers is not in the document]
sjohn4: Yes I changed this and added checks.

PR #656, opened 2025-04-14
Sjohn/gen targets

sjohn4:
Added a field for generative targets on the file wrappers and updated the training config to train with them.  Along with the possibility to apply different transform functions to the targets and the inputs.


MaxiBoether:
Hey John, this PR currently fails the integration tests and is also _quite_ large. Is there a chance we can split this into multiple PRs that target a different branch (not main), s.t. we first merge smaller PRs into a separate branch that then gets merged into main as soon as everything is here? In particular, can we maybe first address the storage to support non-labeled data, and then support this in the rest of the system as well?

sjohn4:
No reason to have this open anymore

PR #657, opened 2025-04-14
feat: Add some tokenizers and maximum token length flag

sjohn4:
Added the necessary code for the tokenizers, appropriate tests and the tokenizer length flag to decide sequence length in the pipeline.


environment.yml
```diff
@@ -26,6 +26,7 @@ dependencies:
       - evidently==0.4.27
       - alibi-detect==0.12.*
       - tenacity
+      - sentencepiece
```
MaxiBoether, line 29: Is this used anywhere in this PR?

modyn/models/tokenizers/distill_bert_tokenizer.py
```diff
@@ -1,24 +1,14 @@
...
-        data = torch.stack((tokens["input_ids"], tokens["attention_mask"]), dim=2)
-        data = torch.squeeze(data, dim=0)  # First shape dim is always 1, since the input is just one string
-        return data
+class DistilBertTokenizerTransform(HFTokenizerTransform):
+    def __init__(self, max_token_length: int = 300):
+        """
+        Adapted from WildTime's initialize_distilbert_transform
+        Here you can find the original implementation:
+        https://github.com/huaxiuyao/Wild-Time/blob/main/wildtime/data/utils.py
+        """
```
MaxiBoether, line 12: this comment should probably be moved to the new `HFTokenizerTransform`, since that one is based on this original impl

modyn/models/tokenizers/t5_tokenizer.py
```diff
@@ -0,0 +1,14 @@
...
+
+from .hf_tokenizer import HFTokenizerTransform
+
+
+class T5TokenizerTransform(HFTokenizerTransform):
+    def __init__(self, max_token_length: int = 128):
+        """Tokenizer transform class for the T5 model.
+
+        Args:
+            max_token_length: Maximum length for tokenization (default: 300).
```
MaxiBoether, line 11: default 300 it says here but it is set to 128

modyn/models/tokenizers/gpt2_tokenizer.py
```diff
@@ -0,0 +1,38 @@
...
+            - dim 1 contains input_ids and attention_mask.
+        """
+        if not sample:  # This is needed here since this tokenizer cannot handle an empty list
+            sample = self.tokenizer.eos_token
+        tokens = self.tokenizer(
+            sample, padding="max_length", truncation=True, max_length=self.max_token_length, return_tensors="pt"
+        )
+        data = torch.stack((tokens["input_ids"], tokens["attention_mask"]), dim=2)
+        data = torch.squeeze(data, dim=0)
+        return data
```
MaxiBoether: I don't understand. GPT2 usses huggingface, so we should use the HFTokenizerTransform. The only difference I can see is the if not sample code path. Why is that needed? The tokenizer should never get an empty string. Even in that case, right here you set sample = token, and then pass this _token_ to the tokenizer, which does not make sense. I don't understand why this extra case is needed, but if it's needed, it affects all HF tokenizers and should probably live in the parent class

MaxiBoether: last, if it truly only affects this tokenizer and is needed, you can check sample and call the superclass's __call__ function instead of re-implementing the tokenization call

modyn/protos/trainer_server.proto
```diff
@@ -59,6 +59,7 @@ message StartTrainingRequest {
   bool enable_accurate_gpu_measurements = 25;
   int64 record_loss_every = 26;
   bool drop_last_batch = 27;
+  int64 tokenizer_sequence_length =28;
```
MaxiBoether: We use this as `max_token_length` and I guess we should name it that way. This flag does not guarantee padding, does it? I.e., if it is set to 2048 and I have a short string, are we guaranteed to have padding? It should probably also be an optional int64 since not every training uses a tokenizer

modyn/trainer_server/internal/dataset/data_utils.py
```diff
@@ -86,7 +88,7 @@ def prepare_per_class_dataloader_from_online_dataset(
     drop_last: bool = True,
 ) -> torch.utils.data.DataLoader:
     # TODO(#289): Replace inefficient per class dataloader
-    dataset = PerClassOnlineDataset(
+    dataset = PerClassOnlineDataset(  # type: ignore  # pylint: disable=too-many-function-args
```
MaxiBoether: We should not use type: ignore unless necessary. If there is a typing issue we need to fix the typing issue

modyn/trainer_server/internal/dataset/online_dataset.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         shuffle: bool,
         tokenizer: str | None,
         log_path: pathlib.Path | None,
+        sequence_length: int = 128,
```
MaxiBoether: rename to max_token_length if that is truly what this is doing

MaxiBoether: also probably this is `int | None` since not every training uses a tokenizer

modyn/trainer_server/internal/utils/training_info.py
```diff
@@ -82,3 +82,4 @@ def __init__(
         self.tokenizer: str | None = request.tokenizer.value if request.HasField("tokenizer") else None
 
         self.offline_dataset_path = offline_dataset_path
+        self.tokenizer_seq_length: int = request.tokenizer_sequence_length
```
MaxiBoether: int | None? (handling similar to e.g. how self.tokenizer is set)

modyn/protos/trainer_server.proto
MaxiBoether: another question, I actually don't see this being used anywhere in the supervisor, i.e., all requests we send out right now don't set this. we should also (similar how you did it for grad clipping) add this to the pipeline config and forward it, no?

MaxiBoether (changes requested):
Thanks John. Here I had some more comments

environment.yml
sjohn4: The T5 tokenizer needs it to work, it is not used in the code but it is needed.

modyn/models/tokenizers/gpt2_tokenizer.py
sjohn4: You are right it does not make sense to put the token in the tokenizer,  I think you can tecnically recieve an empty string if you have a generative "target" that is missing, which could happen in some datasets. all other tokenizers have a token for an empty string but not gpt2, so it is the only one that needs this.

modyn/protos/trainer_server.proto
sjohn4: I think it does guarantee padding up to the tokenizer sequence length.

sjohn4: Hey yes it seems like I deleted that line somehow, I put it back :)

modyn/trainer_server/internal/dataset/online_dataset.py
sjohn4: done

modyn/trainer_server/internal/utils/training_info.py
sjohn4: done

modyn/protos/trainer_server.proto

modyn/trainer_server/internal/dataset/data_utils.py
sjohn4: Solved, not sure how though

PR #658, opened 2025-04-14
feat: Add GPT2

sjohn4:
Added the gpt2 model.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -26,6 +26,7 @@ dependencies:
       - evidently==0.4.27
       - alibi-detect==0.12.*
       - tenacity
+      - flash_attn
   - jsonschema
   - psycopg2
   - sqlalchemy>=2.0
@@ -43,11 +44,10 @@ dependencies:
   - nltk
   - pytorch::pytorch=2.2.1
   - pytorch::torchvision
-  - pytorch::cpuonly # comment out if commenting in lines below for CUDA
-#  - pytorch::pytorch-cuda=12.1
-#  - nvidia::cuda-libraries-dev=12.1.*
-#  - nvidia::cuda-nvcc=12.1.*
-#  - nvidia::cuda-nvtx=12.1.*
-#  - nvidia::cuda-cupti=12.1.*
-#  - nvidia::cuda-cudart-dev=12.1.*
-#  - nvidia::cuda-profiler-api=12.1.*
+  - pytorch::pytorch-cuda=12.1
+  - nvidia::cuda-libraries-dev=12.1.*
+  - nvidia::cuda-nvcc=12.1.*
+  - nvidia::cuda-nvtx=12.1.*
+  - nvidia::cuda-cupti=12.1.*
+  - nvidia::cuda-cudart-dev=12.1.*
+  - nvidia::cuda-profiler-api==12.1.*
--- a/modyn/models/gpt2/gpt2.py
+++ b/modyn/models/gpt2/gpt2.py
@@ -1,0 +1,84 @@
+from typing import Any
+
+import torch
+from torch import nn
+from transformers import AutoTokenizer, GPT2LMHeadModel
+
+from modyn.models.coreset_methods_support import CoresetSupportingModule
+
+
+class Gpt2:
+    def __init__(self, model_config: dict[str, Any], device: str, amp: bool) -> None:
+        self.model = Gpt2Modyn(model_config)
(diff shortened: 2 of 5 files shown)


environment.yml
```diff
@@ -26,6 +26,7 @@ dependencies:
       - evidently==0.4.27
       - alibi-detect==0.12.*
       - tenacity
+      - flash_attn
```
MaxiBoether, line 29: Stupid question, but isn't the flash attention module `flash-attn`, or does it no matter whether we use underscore or dash?

environment.yml
```diff
@@ -43,11 +44,10 @@ dependencies:
...
-#  - nvidia::cuda-nvtx=12.1.*
-#  - nvidia::cuda-cupti=12.1.*
-#  - nvidia::cuda-cudart-dev=12.1.*
-#  - nvidia::cuda-profiler-api=12.1.*
+  - pytorch::pytorch-cuda=12.1
+  - nvidia::cuda-libraries-dev=12.1.*
+  - nvidia::cuda-nvcc=12.1.*
+  - nvidia::cuda-nvtx=12.1.*
+  - nvidia::cuda-cupti=12.1.*
+  - nvidia::cuda-cudart-dev=12.1.*
```
MaxiBoether, line 52: should not touch this on main.

modyn/models/gpt2/gpt2.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class Gpt2:
+    def __init__(self, hparams: Any, device: str, amp: bool) -> None:
+        self.model = Gpt2Modyn(hparams)
+        self.model.to(device)
+
+
+class Gpt2Modyn(CoresetSupportingModule):
+    def __init__(self, hparams: Any) -> None:
```
MaxiBoether: I do not understand the hparams class. Is this a dict[str, something]? Because you use getattr, it does not seem like it, but what type is it. Can we somehow make this more explicit?

Also, can you explain how the supervisor transfers this information to the trainer server? isn't this a dict?

modyn/models/gpt2/gpt2.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if self.enable_flash_attention:
+            self.model.config.attn_implementation = "flash_attention"
+
+        # If gradient checkpointing is enabled, enable it
+        if self.enable_gradient_checkpointing:
+            self.model.gradient_checkpointing_enable()
+
+        # Load model configuration
+        self.config = GPT2Config.from_pretrained(self.model_name)
```
MaxiBoether: What is this object being used for? I am a bit confused since we already have instantiated a model

modyn/models/gpt2/gpt2.py
```diff
@@ -0,0 +1,88 @@
...
+        self.num_return_sequences = getattr(hparams, "num_return_sequences", 1)
+        self.enable_flash_attention = getattr(hparams, "enable_flash_attention", True)
+        self.enable_gradient_checkpointing = getattr(hparams, "enable_gradient_checkpointing", False)
+        self.dtype = getattr(hparams, "dtype", torch.float32)
+
+        # Load tokenizer from gpt2-large (or adjust as needed)
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name)
+        self.tokenizer.pad_token = self.tokenizer.eos_token  # Use right padding
+
+        # Load the GPT-2 model with FlashAttention if enabled
```
MaxiBoether: Remove unnecessary comments

MaxiBoether (changes requested):
Thanks John! I am mostly concerned about this hparams object

environment.yml
sjohn4: I think it either does not matter or the package is flash_attn but when brought up in the code it is flash-attn

modyn/models/gpt2/gpt2.py
sjohn4: You are right this should be a model_config[str,any] I will change it accordingly

modyn/models/gpt2/gpt2.py
sjohn4: I think this was used for the adapters but I think its obsolete I will delete it.

environment.yml
sjohn4: Will remove this, I thick this now explain why it fails the dockertest

PR #659, opened 2025-04-14
feat: Support for `peft` library

sjohn4:
Added the possibility to add extra layers to the models in the form of prefix tuning prompt tuning and lora layers, along with the possibility to add a custom adapter layer.


modyn/protos/trainer_server.proto
```diff
@@ -59,6 +59,8 @@ message StartTrainingRequest {
   bool enable_accurate_gpu_measurements = 25;
   int64 record_loss_every = 26;
   bool drop_last_batch = 27;
+  repeated string model_wrappers = 28;
+  JsonString model_wrapper_args = 29;
```
MaxiBoether: should these be optional?

modyn/config/schema/pipeline/training/config.py
```diff
@@ -164,6 +172,14 @@ def validate_gpus(cls, value: int) -> int:
             raise ValueError("Currently, only single GPU training is supported.")
         return value
 
+    @field_validator("model_wrappers")
+    @classmethod
+    def validate_model_wrappers(cls, wrappers: list[str]) -> list[str]:
+        invalid = [w for w in wrappers if w not in SUPPORTED_WRAPPERS]
+        if invalid:
+            raise ValueError(f"Unsupported model wrappers specified: {invalid}")
+        return wrappers
```
MaxiBoether: If you type the model wrappers as a list of Literal["dsf", "dsf"] etc then you don't need the set and this check. Pydantic will automatically validate the literal

modyn/config/schema/pipeline/training/config.py
```diff
@@ -119,6 +118,15 @@ class TrainingConfig(ModynBaseModel):
...
     )
+
+    model_wrappers: list[str] = Field(
+        default_factory=list,
+        description=("List of model wrappers to apply (e.g., lora, prefix_tuning, kadapter, prompt_tuning)."),
+    )
+    model_wrapper_args: dict[str, dict[str, Any]] = Field(
+        default_factory=dict,
+        description="Arguments passed to each model wrapper.",
+    )
```
MaxiBoether: should also be optional, otherwise a user will always need to supply this.

modyn/common/grpc/grpc_helpers.py
```diff
@@ -251,6 +251,8 @@ def prepare_start_training_request(
             enable_accurate_gpu_measurements=training_config.enable_accurate_gpu_measurements,
             record_loss_every=training_config.record_loss_every,
             drop_last_batch=training_config.drop_last_batch,
+            model_wrappers=training_config.model_wrappers,
+            model_wrapper_args=TrainerServerJsonString(value=json.dumps(training_config.model_wrapper_args)),
```
MaxiBoether: should probably handle the None case similar to the tokenizer ?

modyn/trainer_server/internal/utils/training_info.py
```diff
@@ -82,3 +83,7 @@ def __init__(
         self.tokenizer: str | None = request.tokenizer.value if request.HasField("tokenizer") else None
 
         self.offline_dataset_path = offline_dataset_path
+        self.model_wrappers: list[str] = list(request.model_wrappers)
+        self.model_wrapper_args: dict[str, dict[str, Any]] = (
+            json.loads(request.model_wrapper_args.value) if request.model_wrapper_args.value else {}
+        )
```
MaxiBoether: make optional,  probably requires HasField logic as for other fields

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -102,6 +103,12 @@ def __init__(
         self._scaler = torch.cuda.amp.GradScaler(enabled=training_info.amp, **training_info.grad_scaler_configuration)
         self._info("Grad scaler created.")
 
+        self._model.model = apply_adapters(
+            self._model.model, training_info.model_wrappers, training_info.model_wrapper_args
+        )
+
+        self._setup_optimizers(training_info)
```
MaxiBoether, line 109: This optimizer init is not necessary (duplicate)

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+class AdapterModel(nn.Module):
+    """A model-agnostic adapter container that applies generic adapters at
+    specified layers.
+
+    The adapter modules are applied to the hidden states from the
+    corresponding layers. The final output is the original sequence
+    output augmented by a scaled, normalized adapter fusion.
+    """
+
+    def __init__(  # pylint: disable=dangerous-default-value
```
MaxiBoether: this is a valid warning. having this list as a default will behave differently than you expect: https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+
+class ModularAdapter(nn.Module):
+    """A generic, adapter module.
+
+    It projects the hidden states down to a lower-dimensional space,
+    applies an activation,projects back up to the original dimension,
+    and adds the result via a residual connection.
+    """
+
+    def __init__(
```
MaxiBoether, line 21: Where is this and the subsequent adapter class actually used? I cannot find any reference, maybe I am missing something. Can you potentially double check this file if everything in here is needed?

modyn/models/__init__.py
```diff
@@ -6,6 +6,14 @@
...
 from .dummy.dummy import Dummy  # noqa: F401
 from .fmownet.fmownet import FmowNet  # noqa: F401
+from .modular_adapters.modular_adapters import (  # noqa: F401
+    apply_adapters,
+    apply_kadapter,
+    apply_lora,
+    apply_prefix_tuning,
+    apply_prompt_tuning,
+    count_trainable_params,
+)
```
MaxiBoether: i think these functions should not live here in the init py, people should import them from modular_adapters

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+        task_type=TaskType.CAUSAL_LM,
+        r=adapter_dim,
+        lora_alpha=adapter_alpha,
+        lora_dropout=dropout,
+        target_modules=target_modules,
+    )
+    # Apply LoRA to the inner model. This replaces some of the original layers with LoRA layers.
+    model.model = get_peft_model(model.model, lora_config)
+
+    print(f"\nTrainable parameters AFTER applying LoRA: {count_trainable_params(model)}")
```
MaxiBoether: I feel like these functions are kind of similar. Can't we have one function that does this and we just pass it the peft config (LoraConfig, or PromptTuning config, ...)?

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+        adapter_lower = adapter.lower()
+        args = adapter_args.get(adapter_lower, {})
+        if adapter_lower == "lora":
+            model = apply_lora(model, **args)
+        elif adapter_lower == "kadapter":
+            model = apply_kadapter(model, **args)
+        elif adapter_lower == "prompt_tuning":
+            model = apply_prompt_tuning(model, **args)
+        elif adapter_lower == "prefix_tuning":
+            model = apply_prefix_tuning(model, **args)
```
MaxiBoether: add exception if unknown adapter

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+
+    return model
```
MaxiBoether: the side effect is unclear here. either we update model.model (which we already do) or we instantiate something and return it (so the caller of the function can use that) but i currently dont see why we should need both

modyn/tests/models/test_modular_adapters.py
```diff
@@ -0,0 +1,263 @@
...
+            param.requires_grad = False
+
+
+# --- Simple HParams class ---
+class HParams:
+    def __init__(self, model_name_or_path="gpt2-large", device="cpu", amp=False):
+        self.model_name_or_path = model_name_or_path
+        self.device = device
+        self.amp = amp
+
```
MaxiBoether: I think this might be outdated (see gpt2 pr)

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+        for i, adapter in enumerate(self.adapters):
+            layer_idx = self.adapter_layers[i]
+            adapter_input = hidden_states[layer_idx] + fusion_state
+            fusion_state = adapter(adapter_input)
+        # Normalize fusion state and add a scaled residual to the original sequence output.
+        return sequence_output + self.scale_factor * self.layer_norm(fusion_state)
+
+
+def count_trainable_params(model: nn.Module) -> int:
+    return sum(p.numel() for p in model.parameters() if p.requires_grad)
```
MaxiBoether: move to utils?

modyn/models/modular_adapters/modular_adapters.py
```diff
@@ -0,0 +1,267 @@
...
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.adapter_layers = adapter_layers  # Layers at which to apply adapters
+        self.num_adapters = len(adapter_layers)
+        self.layer_norm = nn.LayerNorm(hidden_size)
+        # Create a list of generic adapters (one per specified layer)
+        self.adapters = nn.ModuleList(
+            [ModularAdapter(hidden_size, adapter_size=adapter_size, dropout=dropout) for _ in range(self.num_adapters)]
+        )
+        self.scale_factor = 0.1
```
MaxiBoether: where is this coming from?

MaxiBoether (changes requested):
Also a couple of comments again here, mostly cleanup of the modular adapters file :)

modyn/models/modular_adapters/modular_adapters.py
sjohn4: I guess we could do that but also, but then I am not sure if we can really apply more than one adapter at the same time, maybe someone wants to apply prompt tuning but also add Lora layers.

modyn/models/modular_adapters/modular_adapters.py
sjohn4: Hey its supposed to be an implementation off imposing kadapters into a model in a customizable way, honestly we can easilly remove its part off apply kadapter.

modyn/config/schema/pipeline/training/config.py
sjohn4: done

modyn/config/schema/pipeline/training/config.py
sjohn4: done

modyn/protos/trainer_server.proto

modyn/common/grpc/grpc_helpers.py
sjohn4: done

modyn/models/modular_adapters/modular_adapters.py
sjohn4: done

modyn/models/modular_adapters/modular_adapters.py
sjohn4: done

modyn/trainer_server/internal/trainer/pytorch_trainer.py
sjohn4: done

modyn/models/modular_adapters/modular_adapters.py
sjohn4: done

modyn/models/__init__.py
sjohn4: done

modyn/models/modular_adapters/modular_adapters.py
sjohn4: Done, restructure the file.

PR #660, opened 2025-04-14
feat: Gradient clipping and accumulation

sjohn4:
Added gradient clipping and accumulation.


modyn/config/schema/pipeline/training/config.py
```diff
@@ -154,9 +154,31 @@ class TrainingConfig(ModynBaseModel):
...
+    def validate_accumulation_steps(cls, value: int) -> int:
+        if value < 1:
+            raise ValueError("Invalid value for gradient_accumulation_steps, must be greater than 0.")
+        return value
+
+    @field_validator("max_grad_norm")
+    @classmethod
+    def validate_max_grad_norm(cls, value: float) -> float:
+        if value is not None and value <= 0:
+            raise ValueError("Invalid value for max_grad_norm, must be greater than 0.")
```
MaxiBoether: this should be checked via the min/max value of `Field`, not via extra validators

modyn/config/schema/pipeline/training/config.py
```diff
@@ -154,9 +154,31 @@ class TrainingConfig(ModynBaseModel):
         None,
         description="Configuration for the torch.cuda.amp.GradScaler. Effective only when amp is enabled.",
     )
+    max_grad_norm: float | None = Field(
+        default=None,
+        description="Clips the gradients normed over this value, if its 0 it will not be used.",
```
MaxiBoether, line 159: You mean if it's None nothing will be clipped?

modyn/protos/trainer_server.proto
```diff
@@ -59,6 +59,9 @@ message StartTrainingRequest {
   bool enable_accurate_gpu_measurements = 25;
   int64 record_loss_every = 26;
   bool drop_last_batch = 27;
+  float grad_norm = 28;
```
MaxiBoether: can we rename this to max_grad_norm?

MaxiBoether: and shouldn't it be optional float? since it's optional in the config schema?

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
```diff
@@ -432,6 +433,38 @@ def test_trainer_init_with_label_transformer(dummy_system_config: ModynConfig):
...
+    trainer_single.train()
+
+    # Instead of checking the logged "num_batches_trained" (which counts mini-batches),
+    # we compare the final model parameters to ensure they are equivalent.
+    final_params_accum = clone_params(trainer_accum._model.model)
+    final_params_single = clone_params(trainer_single._model.model)
+
+    for name in init_params.keys():
+        assert torch.allclose(final_params_accum[name], final_params_single[name], atol=1e-5), f"Mismatch in {name}"
+
```
MaxiBoether, line 466: I appreciate this test, but can we somehow check the actual accumulation and not the side effect (same parameters)? e.g., can we mock the optimizer and model to check how often this/with which data they were called? we can keep this test and add another test, but right now we don't actually test the accumulation

modyn/trainer_server/internal/trainer/pytorch_trainer.py
```diff
@@ -85,6 +85,8 @@ def __init__(
         self.pipeline_id = training_info.pipeline_id
         self.training_id = training_info.training_id
         self.trigger_id = training_info.trigger_id
+        self._max_grad_norm = training_info.max_grad_norm
+        self.gradient_accumulation_steps = training_info.gradient_accumulation_steps
```
MaxiBoether: I think if you make max grad norm private then also gradient accumulation steps should be private. Can you move those to lines to line ~ 121 (where we set a lot of internal variables)?

MaxiBoether (changes requested):
Thanks John, a couple of small comments

modyn/config/schema/pipeline/training/config.py
sjohn4: If it is None then nothing is clipped, we skip the function in the pytorch trainer, I guess we can also just skip when it is 0 since it never makes sense to clip to 0.

modyn/config/schema/pipeline/training/config.py
sjohn4: Done

modyn/protos/trainer_server.proto

modyn/tests/trainer_server/internal/trainer/test_pytorch_trainer.py
sjohn4: Made a new test, I think it does what you wanted.

modyn/trainer_server/internal/trainer/pytorch_trainer.py
sjohn4: done

PR #662, opened 2025-04-24
Sjohn/gen eval

sjohn4:
Here is the bare minimum changes needed so we can evaluate text generation_tasks plus the perplexity eval metric.


MaxiBoether:
I think we can only review/merge this as soon as the storage PR that adds support for non-labeled is merged, no? Not sure how you did this right now without having it merged

PR #663, opened 2025-04-25
Sjohn/bleu

sjohn4:
Adds Bleu evaluation and changes the core evaluation to allow for model.generate of responses instead of reading logits and picking the highest score which usually give to nonsensical sentences.  DO NOT MERGE THIS. You can take a quick look.


sjohn4:
This Pr should be done differently, this will not work well.

PR #664, opened 2025-04-25
Changes to storage, allows for no labels and generative targets.

sjohn4:
Here is everything that has to be changed in storage to be able to train without labels or generative targets.


sjohn4:
This Pr changes the code for the storage file wrappers to allow datasets to not include labels and to include either nothing  or generative targets, which can be saves as bytes like normal samples. Along with tests for these cases.

PR #665, opened 2025-04-25
pytorch_trainer and online dataset

sjohn4:
Everything needed for generative tasks appart from the dataset, we can maybe also turn this into just pytorch trainer or just online_dataset.


sjohn4:
This basically changes the online dataset and pytorch trainer to be either to handle generative targets and non labeled outputs. 
The online dataset basically now includes the possibility of receiving target samples that are also bytes like the normal samples and need either the same or different transformation functions applied. The pytorch trainer changes by having basically a branching path to handle generative training that expects longer outputs from the model. It also substitutes padding for -100 labels to be ignored in loss calculation.

